PR #1, opened 2017-01-18
validator for saved file

LegacyID1991:
Add validator which checks if user has saved their Nuke script, Maya scene or Houdini file


mottosso:
Awesome! Merging.

PR #10, opened 2017-01-19
add python.exe instead of old _python.bat solution

LegacyID1991:
old updates that wasn't in my new clone.


LegacyID1991:
add feature to list already created assets in mb.bat

PR #14, opened 2017-01-20
add feature, file saved validation. moved to outside of pluginfolder …

LegacyID1991:
…because it is a global plugin plugin runs during validationorder

### DIFF
--- a/mindbender/plugins/validate_pipeline_filesaved.py
+++ b/mindbender/plugins/validate_pipeline_filesaved.py
@@ -11,7 +11,7 @@ class ValidateCurrentSaveFile(pyblish.api.ContextPlugin):
     """
 
     label = "Saved file"
-    order = pyblish.api.CollectorOrder + 0.1
+    order = pyblish.api.ValidatorOrder - 0.1
     optional = True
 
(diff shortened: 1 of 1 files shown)


mindbender/plugins/validate_pipeline_filesaved.py
```diff
@@ -11,7 +11,7 @@ class ValidateCurrentSaveFile(pyblish.api.ContextPlugin):
     """
 
     label = "Saved file"
-    order = pyblish.api.CollectorOrder + 0.1
+    order = pyblish.api.ValidatorOrder - 0.1
```
mottosso: Excellent!

But, how come there is a `- 0.1`?

LegacyID1991: so that it runs before my other validators

LegacyID1991: it is not super important but I though it best that it is the first thing to show the user that they have been careless and not saved.

mottosso: Cool!

PR #16, opened 2017-01-20
Update " mkasset"

LegacyID1991:
Add so that mb bat doesn't give error when new asset is created and folders are empty.

### DIFF
--- a/bin/_mkproject.bat
+++ b/bin/_mkproject.bat
@@ -24,8 +24,8 @@ if "%2"=="" goto :missing_name
 if "%3"=="" goto :missing_silo
 
 set PROJECTDIR=%1%2
-set curPROJECT=%1
-set curPROJECTNAME=%2
+set CURRENTPROJECT=%1
+set CURRENTPROJECT_NAME=%2
 set PROJECTSTAGE=%3
 
 if not exist %PROJECTDIR% (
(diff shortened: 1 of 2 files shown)


bin/_mkproject.bat
```diff
@@ -24,8 +24,8 @@ if "%2"=="" goto :missing_name
 if "%3"=="" goto :missing_silo
 
 set PROJECTDIR=%1%2
-set curPROJECT=%1
-set curPROJECTNAME=%2
+set CURRENTPROJECT=%1
```
mottosso: Excellent! Would it be possible to add a comment about what these are for? For our future selves, and other developers coming along.

bin/_mkproject.bat
```diff
@@ -24,8 +24,8 @@ if "%2"=="" goto :missing_name
...
+set CURRENTPROJECT_NAME=%2
 set PROJECTSTAGE=%3
```
mottosso: This variable is also available as `MINDBENDER_SILO` below, this can safely be removed.

LegacyID1991: and it will when back2.0 or any other solution has been created

bin/_mkproject.bat
LegacyID1991: definitely. I am on it

PR #19, opened 2017-01-20
Update  mkasset

LegacyID1991:
THis is continue of #16

### DIFF
--- a/bin/_mkproject.bat
+++ b/bin/_mkproject.bat
@@ -24,8 +24,8 @@ if "%2"=="" goto :missing_name
 if "%3"=="" goto :missing_silo
 
 set PROJECTDIR=%1%2
-set curPROJECT=%1
-set curPROJECTNAME=%2
+set CURRENTPROJECT=%1
+set CURRENTPROJECT_NAME=%2
 set PROJECTSTAGE=%3
 
 if not exist %PROJECTDIR% (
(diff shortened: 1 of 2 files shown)


LegacyID1991:
Here we are again!
🎸

bin/_mkproject.bat
```diff
@@ -24,8 +24,11 @@ if "%2"=="" goto :missing_name
...
 
 set PROJECTDIR=%1%2
-set curPROJECT=%1
-set curPROJECTNAME=%2
+:: This is a variable for the function back.bat.
+set CURRENTPROJECT=%1
+set CURRENTPROJECT_NAME=%2
+:: this variable has become obsolete with the addition of MINDBENDER_SILO variable
+:: this can be removed safly if back.bat is removed or changed for something better.
 set PROJECTSTAGE=%3
```
mottosso: Don't forget to remove this.

Use `MINDBENDER_SILO` in scripts where you need this `%3` variable.

bin/_mkproject.bat
```diff
@@ -24,8 +24,11 @@ if "%2"=="" goto :missing_name
 if "%3"=="" goto :missing_silo
 
 set PROJECTDIR=%1%2
-set curPROJECT=%1
-set curPROJECTNAME=%2
+:: This is a variable for the function back.bat.
+set CURRENTPROJECT=%1
```
mottosso: Did we say to remove "current" from these?

Also there is one variable with an underscore, whereas the others do not have any underscores. I would either make each of them have it, or none of them.

PR #27, opened 2017-01-23
Cleanup

mottosso:
I've gone through the project and added documentation to things missing, fixed broken things and added tests.

The main thing I've added is a skeleton to get the Loader, Creator and Manager working in Nuke. See [`pipeline.py`](https://github.com/mottosso/mindbender-core/blob/master/mindbender/nuke/pipeline.py) in the `nuke/` folder.

Some more highlights.

1. Moved Maya plug-ins to the `/maya` Python package. Now plug-ins can be registered automatically in the `userSetup`, instead of via the `.bat`. This is important if we want to use the plug-ins in tests, that won't go through the `.bat` files.
1. Added docstrings to important functions.

One of the tests I'd like you to familiarise yourself with Arvid is the workflow test. It's a snippet of code that does what artists do, and ensures it passes validation and ends up in the right place. The idea is that, whenever a validator is added, this test is updated to reflect what is a correct model. The test will then tell you when something is changed that breaks the normal modeling workflow for artists.

I'd then expect similar tests for each of the asset families at Mindbender, including rigs and animations. These will save you time in the future, in the form of you making an update and artists coming to you afterwards, unable to work, because something broke. Hours saved, multiplied by the number of artists. Having said that, it does take some getting used to and to update. I would suggest you take the hit initially and get a feel for where it is to your advantage, and then make a decision about whether it's worth doing.

See the commit comments for what has been done specifically, some commits have extended notes about why it was done. You can click the little `...` icon to expand.


mottosso:
Hi @LegacyID1991, let me know what you think about these changes.

LegacyID1991:
Clean and neat as ever. No objections

mottosso:
Excellent. What are your thoughts on the boiler plate for Nuke? Take your time to read it through, and see if there is anything you find difficult to understand. Those notes are for you specifically.

The notes then apply to Houdini as well, along with Mudbox, Photoshop etc. All tools in Maya today should work across all apps you guys use.

mottosso:
Here is the [workflow test](https://github.com/mottosso/mindbender-core/blob/master/mindbender/maya/tests/test_workflow.py#L70), it isn't many lines of code and should be easy to read. It's how a modeler would work with the UI.

I'm picturing similar functions for how you would work in Nuke, and Alldin in Houdini etc.

LegacyID1991:
I saw the pipeline.py for nuke in [commit](https://github.com/mindbender-studio/core/pull/27/commits/2c968468215e544e6eeeedb70e689efc858d8652)
but did you mean anymore with the boiler plate for Nuke ?

The workflow test I understood, what does the quote below do ?

> 69 @with_setup(clear)

mottosso:
> but did you mean anymore with the boiler plate for Nuke ?

It was probably what you saw, as in [this](https://github.com/mottosso/mindbender-core/blob/2c968468215e544e6eeeedb70e689efc858d8652/mindbender/nuke/pipeline.py). By "boiler plate" I mean something to be filled in, in this case by you when you are looking to implement the pipeline tools for Nuke (and Houdini).

> what does the quote below do ?

`with_setup` is a called a ["decorator"](http://thecodeship.com/patterns/guide-to-python-function-decorators/), which takes an argument, in this case `clear` which is a function I've written (it's right above).

You won't really need to understand decorators in order to expand upon this test, or write more tests, all you need to understand is that `clear` is going to get run *before* this particular test is run. It's sort of an "initialisation" or "setup" to the test. Hence the name.

Here's more on decorators if you're curious.

- http://thecodeship.com/patterns/guide-to-python-function-decorators/

PR #32, opened 2017-01-24
add category for assets in mindbender Launcher.

LegacyID1991:
Add _mkassetcate.bat to be called if assets need a category like:
prop
char
env
before the asset.
update of other Terminal API and batch script to match this change.

### DIFF
--- a/bin/_mkassetscate.bat
+++ b/bin/_mkassetscate.bat
@@ -1,0 +1,47 @@
+:: Assets Category API
+::
+:: This file is called whenever a user enters an asset, such as Cat or Shot01.
+::
+:: Arguments:
+::   %1: Name of asset, e.g. Bruce
+::   %2: Absolute path to asset parent directory, e.g. m:\f01_projects\p999_Demo\assets
+::
+:: Example:
+::   $ call _mkasset Bruce %~dp0
+
+@echo off
+
+If "%ASSETSCATE%"=="" goto :missing
+
+title %PROJECTDIR% / %ASSETSCATE%
+
+cd %ASSETSCATE%
+
+:: Clear screen
+cls
+
+echo+
+echo %PROJECTNAME%
+echo You are in: %ASSETSCATE% -------
+echo+
+
+:: List available assets, without their .bat suffix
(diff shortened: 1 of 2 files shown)


LegacyID1991:
Will this creat new problems for mindbender loader and creator that we added a folder between f02_prod/asset/[assetname]/work
f02_prod/asset/[cate]/[assetname]/work
does the scheme need to be updated ?

mottosso:
> Will this creat new problems for mindbender loader and creator

Yes, it will.

Don't bother changing the directory layout, I think the problem the guys are having at the moment lies in the Loader. They want each item in there to appear grouped under a category.

For that to happen, you really only need to look at giving the Loader enough information. In this case, I think a `MINDBENDER_CATEGORY` environment variable in each asset .bat could be a good first step. This could get picked up during publishing and stored alongside each asset. (See teh Integrator plug-in for what data is currently added, and how to add more).

Then, when loading said asset, you could read this data and visually group things.

Not changing the folder structure has other benefits as well that we haven't yet gotten into, about search. If we make this environment variable, and group items in the Loader based on it, it means we could make any environment variable, any identifier or "tag" for an asset and group them thereafter.

It would enable an artist to, for example, search for "any character asset" (as is the case with categories) but also things like "any lowpoly model" or "any static prop with fewer than 3 assigned textures, from any project.". Or even just "any recently published asset".

We could add a little search bar to the Loader, that would enable the artist to search by name, or any of the available tags, along with a dropdown menu with pre-defined items, such as "Recent".

Hope this paints a convincingly enough of a picture to divert any changes to the physical directory layout on disk, and inspire some future possibility. :)

mottosso:
Another thing we talked about is enabling the artist to make a choice on category via the Creator. It could be another text field, or a dropdown menu with a few pre-defined category.

LegacyID1991:
Okey. I had a feeling that it would create issues.
I will look into a different solution then too make this possible.
The main issue was that artists wanted a step before assets to specify category because without the category the launcher list gets a bit long and intimidating.

LegacyID1991:
I will close this and return with a better solution

PR #40, opened 2017-02-09
Container Manager improvements and automatic mbID

mottosso:
This PR includes two primary updates.

1. **Container Manager Labels** - The Container Manager now displays the name of the subset, instead of the Asset. The result is that two subsets of the same asset - e.g. `modelDefault` and `lookdevDefault` - can be imported without causing confusion in the manager.
2. **Automatic IDs** - IDs are now automatically added to meshes that don't already have one (on file-save). The result is that modelCleaner isn't strictly necessary, and that "slask" meshes get an ID as well, which then passes validation.

### Automatic IDs

Upon saving the scene, all meshes that don't already have and `mbID` attribute is given one. Since a scene must be saved prior to publishing, this works transparently to the end user.


LegacyID1991:
skriver du dem där för hand eller ?
sett andra sånna meddelanden från andra på github som jag följer.
Säg att de är autogenereade ? plz :)

mottosso:
Sorry, they are all written by hand. :)

mottosso:
Oh, did you remove your comment? No need for that. :)

Should be give this PR a try? I can merge, then you can test it out locally and let me know if it works the way you'd expect.

mottosso:
Merging this. See if you can either rebase your commits on-top of this or just merge yours ontop of it.

Issue #9, opened 2017-01-19
_mkproject.bat bad help


mottosso:
### Problem

The help text for `_mkproject.bat` says the wrong thing, if the project name includes a subdirectory.

```bash
MEINDBENDER START ---------------------------


                _,--,
             .-'---./_    __
            /o \\     '-.' /
            \  //    _.-'._\
             `"\)--"`


  Welcome marcus!

  1. Type first characters of a project, e.g. "p999_"
  2. Press TAB to cycle through matching projects

 ---------------------------------------------

$ p999_Meindbender_Sync
 Uh oh..
   Specify either "assets" or "film"

 Example:
   $ p999_Meindbender_Sync\f02_prod assets
   $ p999_Meindbender_Sync\f02_prod film

$
```

This line...

```bash
   $ p999_Meindbender_Sync\f02_prod assets
```

Should be..

```bash
   $ p999_Meindbender_Sync assets
```

The reason is because the project `name` is..

```bash
call _mkproject %~dp0 %~n0\f02_prod %1
```

It includes `f02_prod` in the name.

### Implementation

To fix this, we'll need to strip characters after `\` in the name in `_mkproject.bat`.

- See [here](https://github.com/mindbender-studio/core/blob/6271520f0f447b6fc2c7b7c03430dc778345a847/bin/_mkproject.bat#L74)

mottosso:
Fixed!

LegacyID1991:
although not yet pull requested so might not be fixed ??? 😰

mottosso:
Sorry! Didn't notice.

LegacyID1991:
Implantation suggestion:
[ProjectName.bat](https://github.com/mindbender-studio/core/blob/master/template/ProjectName.bat)
```
call _mkproject %~dp0 %~n0 %1
```
could instead be
```
call _mkproject %~dp0 %~n0\f01_prod %~n0 %1
```
so it becomes
```
call _mkproject arg1 arg2 arg3 arg4
- arg1 = %~dp0 = drive + path to bat file
- arg2 = %~n0\f01_prod = "f01_prod" could be what ever, this argument is for the projectname plus extra path
- arg3 = %~n0 = Name of the project, just as before the bat name
- arg4 = %1 = silo, same as before
```

Files needed to alter:
[ProjectName.bat](https://github.com/mindbender-studio/core/blob/master/template/ProjectName.bat)
[...]

PR #41, opened 2017-02-09
small fixes and updates

LegacyID1991:
**small changes**

updated comments inside of bat files to help users who wanna customize project or assets
updated userSetup.py for maya so it listens to variables set by bat files.
updated templates of bats inside of template folder
**Added variables for project bats:**
- _MINDBENDER_SETASSETCATEGORY_ = resets the selected category of assets when user returns to root
- _MINDBENDER_ASSETCATEGORY_  = sets the available assets categories for this project
- _MINDBENDER_FPS_ =  sets the delivery fps for the project
- _MINDBENDER_RESOLUTION_WIDTH_ = sets the delivery format width for the project
- _MINDBENDER_RESOLUTION_HEIGHT_ = sets the delivery format height for the project

**added variables for assets bats:**
- _OBJECT3D_ = sets this assets category this will be querryed by the "MINDBENDER_ASSETCATEGORY" variable 
- _MINDBENDER_EDIT_IN_ = use only for shots, this sets time in
- _MINDBENDER_EDIT_OUT_ = use only for shots, this sets time in


mottosso:
This is a big one! Let's try and go through it piece by piece.

Ideally, in the future it would best to PR only commits that are somehow related. For example, when fixing #9, that would be a good PR to merge individually.

Because the fix for #9 is mixed with other commits, it means merging of the fix will be delayed because of the other commits.

LegacyID1991:
that would involve branches right ?

mottosso:
> that would involve branches right ?

Either branches, or wait to push new changes to your repo until we've merged the PR.

For branches, I would recommend you create a small test repo and experiment. They are difficult and scary, but sometimes useful. Especially when collaborating with others.

bin/_mkextrastep.bat
```diff
@@ -0,0 +1,61 @@
+:: Extra Step Terminal API
```
mottosso: What does this do?

bin/_mb.bat
```diff
@@ -29,7 +29,7 @@ if "%PYBLISH_NUKE%"=="" goto :missing
 if "%PYBLISH_LITE%"=="" goto :missing
 if "%PYBLISH_QML%"=="" goto :missing
 if "%MINDBENDER_CORE%"=="" goto :missing
-:: if "%PYBLISH_USERNAME%"=="" goto :missing
+if "%PYBLISH_USERNAME%"=="" goto :missing
```
mottosso: Why does there need to be a username?

LegacyID1991:
That one was going to be a solution of creating a category option for
assets.
It would trigger if the project bat had a variables of categories. It would
prompt you to chose a category before moving into the _mkproject and the
asset choice.

bin/_mb.bat
mottosso: And what should the value be?

1. If we set it in the local `mb.bat`, then the value would be the same for everyone who uses it, which doesn't seem right. 
2. If the value is `%USERNAME%` then it might as well be used directly.

mottosso: These are the kinds of things we should first make an issue thread about, to discuss, and then implement. Once implemented, it's important to highlight these additions in the PR description, as detailed as necessary. With examples and motivation if necessary.

LegacyID1991:
That was a test of an alternative from using the windows login id because
we still use mindbender as a main username on every single computer.

On 9 Feb 2017 12:31 p.m., "Marcus Ottosson" <notifications@github.com>
wrote:

*@mottosso* commented on this pull request.
------------------------------

In bin/_mb.bat
<https://github.com/mindbender-studio/core/pull/41#pullrequestreview-20987212>
:

> @@ -29,7 +29,7 @@ if "%PYBLISH_NUKE%"=="" goto :missing
 if "%PYBLISH_LITE%"=="" goto :missing
 if "%PYBLISH_QML%"=="" goto :missing
 if "%MINDBENDER_CORE%"=="" goto :missing
-:: if "%PYBLISH_USERNAME%"=="" goto :missing
+if "%PYBLISH_USERNAME%"=="" goto :missing

Why does there need to be a username?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/mindbender-studio/core/pull/41#pullrequestreview-20987212>,
or mute the thread

mottosso:
It's better if you comment on GitHub, as then comments would be nested under the parent.

bin/_mkextrastep.bat
mottosso: > That one was going to be a solution of creating a category option for
assets.

Was going to? Is it not currently used? Then I think we should remove it from the PR.

bin/_mb.bat
mottosso: > That was a test 

Ok, better remove temporary things from the PR.

mottosso:
If you can take a peek at the changes and remove temporary things that you don't feel should be in the main repository, then I'll take a closer look before we merge.

LegacyID1991:
PUBLISH_USERNAME was suppose to be a static global variable set by the user
before they launched my.bat

bin/_mb.bat
mottosso: > PUBLISH_USERNAME was suppose to be a static global variable set by the user
before they launched my.bat

Ok, better start an issue about it so we can talk about it and consider options, pros and cons.

LegacyID1991:
How about now ?

mottosso:
Having a look.

mottosso:
Running into problems here.

```bash
$ mb.bat
...
$ hulk assets
 Uh oh..
   Specify either "assets" or "film"

 Example:
   $ hulk assets
   $ hulk film
```

Using the [example project](https://github.com/mindbender-studio/example). If there has been a change to how projects are structured, we will need to update the example as well so I and others can reproduce the errors locally.

LegacyID1991:
Yea the edit in _mkprojects has changed how the .bat files look.
There are new templates to be used that work.

mottosso:
Don't forget to document this, especially in the PR description. Include an example of the change along with a motivation.

The motivation should ideally come from an original issue that this PR implements, that would be where we also have a discussion about the best way to approach it. This PR would be an exception, but in the future I'll be more insistent on that, for the safety of the project today and for future developers who seek to understand why we did the things we did. Otherwise we'll end up with another re-write because the next developer didn't understand things.

Does that make sense?

LegacyID1991:
Totally, i will do better in the future.
Can you elaborate on the best way to document a change like this one?
Should I just write more about it in the commits and PR or is there some other nice way to show the change ?

mottosso:
> Can you elaborate on the best way to document a change like this one? Should I just write more about it in the commits and PR or is there some other nice way to show the change ?

You could write in the original PR description above.

As someone who is receiving this change, I would have expected these things to be documented.

1. How it looked before
2. Why it wasn't sufficient/what was missing (This is usually what we cover in the issue)
3. What it looks like now
4. What a user must to in order to update.

For example..

### New Project Structure

Before, each project `.bat` file included an absolute path, a name and silo.

[**hulk.bat**](https://github.com/mindbender-studio/example/blob/master/projects/hulk.bat)

```bash
call _mkproject %~dp0 %~n0 %1
```

But because of A, B and C (this is where you document *why* a change was made), it now looks like this.

**hulk.bat**

```bash
call _mkproject %~dp0 %~n0 ??? %1
```

(Where ??? is what the change is, I still don't know what to add).

As a result, all of your project `.bat` files must be updated like this.

1. Step one in updating process
2. Step two in updating process
3. Finishing up
4. Done

mottosso:
Arvid, how do I update the project bat on my end?

LegacyID1991:
you mean how do you change hulk.bat
just grab the template ProjectName.bat and then rename it I suppose. ??
or did I misunderstand the question.

mottosso:
I'm looking to run your fork locally, but I don't have your projects. Instead, I'm using the [example project](https://github.com/mindbender-studio/example).

Which `ProjectName.bat` are you referring to? [This one](https://github.com/mindbender-studio/core/blob/master/template/ProjectName.bat). It doesn't seem updated either.

If you could post what your `ProjectName.bat` looks like, then I could use update the ones I have.

LegacyID1991:
```
:: :: Following is the commands that needs to be called in the bat
:: :: always call _mkproject
:: %~dp0 is drive plus path variable
:: %~n0 is the name of this project as the name of the .bat
:: \\f02_prod is the folder that you will be
:: %1 is the entry by the user for asset
@echo off
call _mkproject %~dp0 %~n0 \\f02_prod %1

:: :: Following is the variables that can be set in the bat
:: :: They are default set to "nothing"
:: :: leave them be if you dont have the info for them and return to them later.
set MINDBENDER_SETASSETCATEGORY=
set MINDBENDER_ASSETCATEGORY=
set MINDBENDER_FPS=
set MINDBENDER_RESOLUTION_WIDTH=
set MINDBENDER_RESOLUTION_HEIGHT=
```
like this ?

mottosso:
That's the one.

About this:

```
call _mkproject %~dp0 %~n0 \\f02_prod %1
```

What do we do for projects that don't have the `\\f02_prod` directory? For example, with the example projects.

LegacyID1991:
I guess that it doesn't really work if you dont let  __mkproject_ create the folder structure.

mottosso:
Hm, it doesn't look like[ `_mkproject.bat`](https://github.com/LegacyID1991/core/blob/8ee28fc77b315526594f2b325f188572cbb87454/bin/_mkproject.bat) is creating `f02_prod`.

And it really shouldn't, `f02_prod` is a detail of the particular project. Having it embedded into the core bat means you are locked into this structure for every project, and that when you eventually change break all previous projects.

Have a think about how we can avoid that.

LegacyID1991:
well we did avoid it by adding it to the end of like it is displayed in the second option
> %~n0 
> %~n0\f02_prod

but that was what created the bad help #9 
so I guess we go back to the old bat file but then we need to come up with a diffrent way to solve #9

mottosso:
> so I guess we go back to the old bat file but then we need to come up with a diffrent way to solve #9

I think so too.

bin/_mkproject.bat
```diff
@@ -1,29 +1,39 @@
...
+::  Arguments:
+::    %1: Path = Absolute path to projects
+:: 	  %2: Name = Name of project
+::    %3: \f02_prod
+::    %4: Silo = Parent directory name of assets
 ::
-:: Variables:
-::   %PROJECTDIR%: Absolute path to project directory
+::  Variables set by bat:
+::    %MINDBENDER_SETASSETCATEGORY%: Is the currently chosen category of assets
```
mottosso: Is the category variable still in use?

mindbender/maya/pythonpath/userSetup.py
```diff
@@ -2,6 +2,50 @@
 
 from maya import cmds
 
+import os
+
+
+def mayafpsconverter(Sfps):
```
mottosso: Something like this may be more concise.

```python
def mayafpsconverter(fps):
  return {
    None: "",
    "": "",
    "15": "game",
    "24": "film",
    "25": "pal",
    ...
  }.get(fps, fps + "fps")
```

No need to check whether the variable is a string, environment variables can only be string.

template/AssetName.bat
```diff
@@ -1,2 +1,13 @@
+:: :: Following is the commands that needs to be called in the bat
```
mottosso: Remove the double `::`

template/AssetName.bat
```diff
@@ -1,2 +1,13 @@
...
+:: %~dp0 is drive plus path variable
 @echo off
-call _mkasset %~n0 %~dp0%
\ No newline at end of file
+call _mkasset %~n0 %~dp0%
+
+:: :: Following is the variables that can be set in the bat
+:: :: They are default set to "nothing"
+:: :: leave them be if you dont have the info for them and return to them later.
+set OBJECT3D=
```
mottosso: Don't forget to document what you expect these variables to contain.

mottosso:
Getting there! Just a few more tweaks now. :)

mindbender/maya/pythonpath/userSetup.py
LegacyID1991: but will that work if the user sets something else that isn't one of mayas string arguments.
exampel 
maya accepts "12fps" as an option
but it wont accept "25fps" as an option
to get 25 fps you need to enter "pal"

bin/_mkproject.bat
LegacyID1991: it is a reset for the variable, it was suppose to be integrated with the extra step option bat that would allow the user to chose the category.

mindbender/maya/pythonpath/userSetup.py
mottosso: Yes, the second argument to `get` is the default value, if `fps` isn't available.

```python
assert {"hello": "world"}.get("notexist", "defaultvalue") == "defaultvalue"
```

mottosso: > to get 25 fps you need to enter "pal"

Oh is it the other way around? Then you can do this.

```python
def mayafpsconverter(fps):
  return {
    "": None,
    "game": "15",
    "film": "24",
    "pal": "25",
    ...
  }.get(fps, fps + "fps")
```

Put the value you expect in the variable on the left, and the value you want it to convert to on the right.

bin/_mkproject.bat
mottosso: Ok, should we not include this in this PR then?

template/AssetName.bat
LegacyID1991: will do,
wanted it to look like a headline without using double spaces.

bin/_mkproject.bat
LegacyID1991: I will do it now.

template/AssetName.bat
mottosso: Aah, ok. Good thinking, but I don't think it gets the point across.

The way it's managed in Python docstrings is to indent children of a section, and I think that could work here as well..

```bash
:: Following is the commands that needs to be called in the bat
:: always call _mkproject
::
:: Variables:
::   %~n0 is the name of this project as the name of the .bat
::   %~dp0 is drive plus path variable
```

Btw, what does `always call _mkproject` mean? Try and think of it as someone who reads it for the first time. I'm not sure it makes sense. Should he call _mkproject? When? Always? Like once a day, once per shot?

mindbender/maya/pythonpath/userSetup.py
LegacyID1991: to get 25 fps in maya then you need to enter "pal"
maya only accepts strings as input to change fps. and some of the strings are simple like:
> "12fps" => maya set settings to "12fps" and is now working with 12 fps

But for some reasons other strings are like this:
> "pal" => maya set settings to "pal" and is now working with 25 fps

So this wont work:
> "25 fps"  => maya says string is not acceptable

The example above are not actual python commands for maya.

LegacyID1991: but in the bat files I wanna enter the string like:
> set MINDBENDER_FPS=25

and that will then be returned by 
> def mayafpsconverter(fps):

as 
> "pal"

template/AssetName.bat
LegacyID1991: typo

mindbender/maya/pythonpath/userSetup.py
mottosso: Ok, this should do it.

```python
def mayafpsconverter(fps):
  return {
    "15": "game",
    "24": "film",
    "25": "pal",
    "30": "ntsc",
    "48": "show",
    "50": "palf",
    "60": "ntscf",
    None: "",
    "": "",
  }.get(fps, fps + "fps")
```

In fact, I would either make this function an official part of the API, like put it in `mindbender.lib` or not make a function out of it.

LegacyID1991: since it is so maya specific could it be in mindbender.lib or should it stay in the userSetup.py

LegacyID1991: and wont that return 
>mayafpsconverter("25")
>result: "palfps"

mottosso: Ah no you are right, it should be in `mindbender.maya.lib` in that case. But personally I would just put it [here](https://github.com/mindbender-studio/core/pull/41/files#diff-59cde0eedfb004dc984ab1e8d2b8bd9aR67) and skip the function. It's so short and specific to this task.

mindbender/maya/pythonpath/userSetup.py
```diff
@@ -5,46 +5,26 @@
...
-    if Sfps == "50":
-        condition = 1
-        return "palf"
-    if Sfps == "60":
-        condition = 1
-        return "ntscf"
+def mayafpsconverter(fps):
     ERRORSTRING = "MINDBENDER_FPS has bad value in the bat file"
-    if str(Sfps).isdigit() is False:
+    if str(fps).isdigit() is False:
```
mottosso: You don't need to check this, environment variables can only be string.

mottosso: And also, you are converting to string yourself with `str()` to this will always be `False`.

mottosso: Sorry, scratch that last part.

You still don't need to check it though.

LegacyID1991: isdigit() checks if there is a letter character in there.
it will accept "12"
but not "1a2"

mottosso: Ah, now I understand.

Could you make a note about that, for future readers (including us)? Point out that the user may provide letters intermixed with integers, and that it's bad. Otherwise, a future developer might look at this, think it isn't necessary and remove it. Like I would have done.

LegacyID1991: you want me to do it in the userSetup or in a new commit ?

LegacyID1991: done

bin/_mkproject.bat
```diff
@@ -55,9 +65,10 @@ for %%i in (*.bat) do (
     set temp=%%i
     echo   !temp:.bat=!
 )
+endlocal
 
 echo+
-echo   1. Type first letters of asset
+echo   1. Type first letters of asset category
```
mottosso: Let's remove all notions of a category, then I think we're finished with this PR.

I'll mark each for completeness.

template/ProjectName.bat
```diff
@@ -1,6 +1,22 @@
...
 @echo off
-call _mkproject %~dp0 %~n0 %1
\ No newline at end of file
+call _mkproject %~dp0 %~n0 %1
+
+:: Following is the variables that can be set in the bat
+::
+:: They are default set to "nothing"
+:: leave them be if you dont have the info for them and return to them later.
+:: MINDBENDER_SETASSETCATEGORY should always be set to nothing, this is a reset
```
mottosso: Remove this.

template/ProjectName.bat
```diff
@@ -1,6 +1,22 @@
...
+:: MINDBENDER_ASSETCATEGORY expects categorys split by a "," exampel "prop, env, char"
```
mottosso: And this.

template/ProjectName.bat
```diff
@@ -1,6 +1,22 @@
...
+:: MINDBENDER_FPS, MINDBENDER_RESOLUTION_WIDTH, MINDBENDER_RESOLUTION_HEIGHT expects a 
+:: number value without letter characters
+set MINDBENDER_SETASSETCATEGORY=
```
mottosso: And these two.

bin/_mkproject.bat
```diff
@@ -1,17 +1,26 @@
...
+::  Arguments:
+::    %1: Path = Absolute path to projects
+:: 	  %2: Name = Name of project
+::    %3: Silo = Parent directory name of assets
 ::
-:: Variables:
-::   %PROJECTDIR%: Absolute path to project directory
+::  Variables set by bat:
+::    %MINDBENDER_SETASSETCATEGORY%: Is the currently chosen category of assets
+::    %MINDBENDER_ASSETCATEGORY%: Is a list of asset category to be listed
```
mottosso: This one as well.

bin/_mkproject.bat
LegacyID1991: done

template/ProjectName.bat
LegacyID1991: done

template/ProjectName.bat
LegacyID1991: done

template/ProjectName.bat
LegacyID1991: done

Issue #43, opened 2017-02-10
Have clone special tool connect visibility


CalleHalldin:
On the current rig where we have different hair styles, noes, etc that is connected with visibility driven by an attribute on the rig,
e.g. Attribute _Woman01_ has a setdriven connection to visibility to a group containing the meshes for _Woman01, groups for other variations like _Man01_ are turned off.

The issue we noticed is that the out mesh created with the "clone special" tool does not get these visibility connection from the rigged mesh so alembics will not record these changes when exporting.

**Update** We need a a new Clone (special) tool that connects the visibility but also to keep the old one since we don't want visibility to be connected from the model reference to the rigged mesh only from the rigged mesh to the Out mesh.

CalleHalldin:
Side question that relates, can we group the cloned meshes freely inside the OUT group if we wanted to hock up the visibility between the groups manually?

mottosso:
@LegacyID1991 this can be implemented [here](https://github.com/mindbender-studio/core/blob/5d076e60f292805fef98f7d675fccaaae6480140/mindbender/maya/interactive.py#L104).

This is where currently *all* user-defined attributed are added to a new clone. What you'll need to do is to simply include `.visibility` there as well.

mottosso:
> Side question that relates, can we group the cloned meshes freely inside the OUT group if we wanted to hock up the visibility between the groups manually?

Better make this a separate issue with a more thorough description and something I can reproduce. Not sure I'm understanding what you mean.

LegacyID1991:
This issue needs to be updated
as of now I am not working on this.

mottosso:
Updated with what? I've assigned it to you to indicate that it's up your alley. You are free to implement it whenever you find the time.

LegacyID1991:
I have talked with @CalleHalldin about this feature and after that I felt that this issue needed to be updated.

mottosso:
Sorry, can you be more specific? Is there something you would like to add, should I close it or is someone else working on it?

LegacyID1991:
Leave it open, Calle will either edit this issue or I will complete it and refer to it with my solution.

Issue #42, opened 2017-02-09
LookDev transfer Error


CalleHalldin:
[Example move](https://drive.google.com/file/d/0B7RmZ_BoQHNeMmRSQm43RWozWjQ/view?usp=sharing)
Process to get to produce the bug:

- Create a model for _assetA_.
- Create a lookDev for _assetA_.
- Create model _assetB_.
- Create lookDev for _assetB_.
- Load _ModelA_ and Lookdev into a new scene
- Load _ModelB_ and Lookdev into same scene
- Load Lookdev for _ModelA_ again and it will get the _ModelB_ lookDev and it can't be changed back with the loader.

Seems like its always the First models and lookDev that you load that transfers the shader to the wrong asset. As can be seen in the video even though I create a completely new scene the bug is still there.

mottosso:
Thanks @CalleHalldin, having a look at this tomorrow morning. Make sure the steps to reproduce does reproduce the problem, that way it'll be much quicker to fix.

mottosso:
Sorry I haven't gotten into this yet.

@LegacyID1991, you might be able to dive into this. The problem must be somewhere in the loader for this family, see [here](https://github.com/mindbender-studio/core/blob/master/mindbender/maya/loaders/mindbender_look.py).

I would start printing things, and see if you can pinpoint where in the loader the problem occurs. Likely in the `maya.apply_shaders` function. If you print the [`relationships`](https://github.com/mindbender-studio/core/blob/e95bc7655bc37ebf75d09a5fb76cf42d27b9d9c8/mindbender/maya/loaders/mindbender_look.py#L52) variable, it'll give you a dictionary. You can use that dictionary to manually call `apply_shaders` and debug from there.

mottosso:
@LegacyID1991 Did you have a look at this? If not, let me know here so I know whether I should look at it.

LegacyID1991:
I am sorry to say that I have been so slammed with other work duties that I haven't gotten to this one yet.

mottosso:
That's ok, I just need to know whether I should look into it so we don't work on it at the same time. Would you like me to take over?

LegacyID1991:
go right ahead.
😄

mottosso:
Ok, I've found the problem.

The problem is that it assigns to meshes based on the name of the shader group, but if the shader group has the same name in both look dev publishes, such as `blinn1SG`, the it gets confused and assigns the wrong one.

Putting together a fix now.

[...]

PR #44, opened 2017-02-13
Cosmetics

mottosso:
Hi @LegacyID1991,

I've made a few tweaks to your latest PR, code convention and documentation. But primarily I've refactored userSetup.py, as it seems no other value than those left are possible.

Before, the default value was `fps + "fps"` which would have resulted in `17` becoming `17fps`. But Maya [can't take that value](http://help.autodesk.com/cloudhelp/2015/ENU/Maya-Tech-Docs/CommandsPython/currentUnit.html). Because of that, I also removed the check, and instead defaulted to `"pal"` if no valid value was found.

I removed the interactive dialog box warning against an invalid FPS as well, because (1) better not assume too much until you've experienced a problem with it first hand and (2) interactive errors should not happen in userSetup.py, because userSetup.py should never fail. If it fails, Maya may become unstable simply because we can't be sure about at which point in the userSetup it failed.

If you take a look at the code now, it should be both shorter and failsafe. Any errors coming in from the environment are gracefully handled and falls back to a default value.

### DIFF
--- a/template/AssetName.bat
+++ b/template/AssetName.bat
@@ -10,9 +10,6 @@ call _mkasset %~n0 %~dp0%
 :: Additional variables
 :: ------------------
 
-:: (str, optional): This variable is ??????, for example "prop, char"
-set MINDBENDER_OBJECT3D=
-
 :: (int, optional): In-time used in the edit
 set MINDBENDER_EDIT_IN=101
 
--- a/template/AssetName.bat
+++ b/template/AssetName.bat
@@ -6,13 +6,15 @@
 @echo off
 call _mkasset %~n0 %~dp0%
 
-:: Following is the variables that can be set in the bat
-::
-:: They are default set to "nothing"
-:: leave them be if you dont have the info for them and return to them later.
-:: setOBJECT expects a string like "prop, char"
-:: MINDBENDER_EDIT_IN and MINDBENDER_EDIT_OUT sets the assets time in and out for the edit. 
-:: These should not be set for assets, use this options for shots
-set OBJECT3D=
-set MINDBENDER_EDIT_IN=
-set MINDBENDER_EDIT_OUT=
+:: ------------------
+:: Additional variables
+:: ------------------
+
+:: (str, optional): This variable is ??????, for example "prop, char"
+set MINDBENDER_OBJECT3D=
+
+:: (int, optional): In-time used in the edit
+set MINDBENDER_EDIT_IN=1000
+
+:: (int, optional): Out-time used in the edit
+set MINDBENDER_EDIT_OUT=1200
--- a/template/AssetName.bat
+++ b/template/AssetName.bat
@@ -14,7 +14,7 @@ call _mkasset %~n0 %~dp0%
 set MINDBENDER_OBJECT3D=
 
 :: (int, optional): In-time used in the edit
-set MINDBENDER_EDIT_IN=1000
+set MINDBENDER_EDIT_IN=101
 
 :: (int, optional): Out-time used in the edit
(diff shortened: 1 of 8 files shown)


template/AssetName.bat
```diff
@@ -6,13 +6,15 @@
...
-set OBJECT3D=
-set MINDBENDER_EDIT_IN=
-set MINDBENDER_EDIT_OUT=
\ No newline at end of file
+:: ------------------
+:: Additional variables
+:: ------------------
+
+:: (str, optional): This variable is ??????, for example "prop, char"
+set MINDBENDER_OBJECT3D=
```
mottosso: One thing before we can merge this, what is this variable? I couldn't find it referenced anywhere in the code.

LegacyID1991: This is the variable that assigns which category that object belong to.
this was suppose to be part of the category setup where this value would be used to sort the _bat files_ / options in the assets menu.

mottosso: Ah, so it's safe to remove?

LegacyID1991: Yes

mottosso:
Did you find what was wrong with the Single Assembly validator?

LegacyID1991:
No.
I couldn't find any changes done to that plugin in the last week so I don't think it is because of the plugin.
what worries me more is that at this time, your repo and mine should be exactly the same, except for this PR that you are doing right now.

mottosso:
They are not, you can see that in your repos commit history.
- https://github.com/LegacyID1991/core/commits/master

You can checkout the master branch of this repo into your fork, then you can remove your old master and push the new one back into your repository. Or delete your fork, and re-fork.

LegacyID1991:
I did [this](https://github.com/LegacyID1991/core/commit/f5de55a1f3ff19e4a932b1da44147fce0e7d56c2)
Then I started the pull request #41 
after that we edited and changed until we merged
and I deployed both core and dev during lunch on monday.

How can they not be the same ?

mottosso:
It appears you haven't pulled from this repo, into your repo.

Here's what you can do.

```bash
$ cd your_fork
$ git remote add stable https://github.com/mindbender-studio/core.git
$ git pull stable master
```

Let me know if that works for you.

mottosso:
Included a fix for #42 here, let me know if you have any objections to merging this @LegacyID1991.

mottosso:
I updated the look of items in the container manager.

image

LegacyID1991:
I have no objections.
I like what you did with the look inside of Container Manager.
will it look like that inside of the loader as well?

mottosso:
> will it look like that inside of the loader as well?

How would you like the loader to change?

image

LegacyID1991:
not really a change,
I liked that the button was different colors and that you included info inside of the brackets.
I was thinking if there was more info that the artist could need that we could display inside of brackets.

mottosso:
Sure, just let me know what info you would like to have there and I'll add it. Should we merge this first?

LegacyID1991:
if core is stable with this pull request, then YEA lets do it

mottosso:
All done. Don't forget to update.

LegacyID1991:
>
```
$ cd your_fork
$ git remote add stable https://github.com/mindbender-studio/core.git
$ git pull stable master
```
> Let me know if that works for you.

I think it did. It did more then fast forward did.
I will start working on #43 for now and will make a PR when done, hopefully it will be correct this time

mottosso:
No problem, we'll get there.

Take a moment to understand fast-forward, it's important.

- https://ariya.io/2013/09/fast-forward-git-merge

PR #52, opened 2017-02-21
Changes to variables in project bats

LegacyID1991:
Previusly [_mkprojects.bat](https://github.com/mindbender-studio/core/blob/master/bin/_mkproject.bat) listened to 3 arguments.
```
call _mkprojects (1) (2) (3)
1. direct path before the bat file
2. bat filename + direct path after bat file
3. user input choice
```

I now introduce a new argument in the middle, making it 4 arguments
```
call _mkprojects (1) (2) (3) (4)
1. direct path before the bat file
2. direct path after the bat file
3. bat filename
4. user input choice
```

This require changes to all project bat files including:
[hulk.bat](https://github.com/mindbender-studio/example/blob/master/projects/hulk.bat)
[spiderman.bat](https://github.com/mindbender-studio/example/blob/master/projects/spiderman.bat)

from
```
call _mkproject %~dp0 %~n0 %1
```
to
```
call _mkproject %~dp0 %~n0 %~n0 %1
```


LegacyID1991:
This will also solve #9

mottosso:
> This require changes to all project bat files including:

Don't forget to include what to change these from and to.

LegacyID1991:
better ?

mottosso:
For completeness and our future selves, the motivation for this PR is (1) to solve #9 but also to (2) workaround the Windows path limit to enable a very long bat-file name, with a short project directory name.

At the moment, the project name and bat files are identical. In order to create a new project, the user must create a new bat file and gives it the name of the project. With this change, an additional step is added where the user must enter another name inside of it.

Let's do it!

PR #56, opened 2017-02-24
Fix #43

### LINKED ISSUE #43: Have clone special tool connect visibility

mottosso:
This automatically connects visibility of a specially cloned mesh.

### DIFF
--- a/mindbender/maya/interactive.py
+++ b/mindbender/maya/interactive.py
@@ -108,6 +108,10 @@ def clone_special(*args):
             value = cmds.getAttr(transform + "." + attr)
             cmds.setAttr(new_transform + "." + attr, value, type="string")
 
+        # Connect visibility
+        cmds.connectAttr(transform + ".visibility",
+                         new_transform + ".visibility")
+
 
 def clone_worldspace(*args):
     return _clone(worldspace=True)


LegacyID1991:
very good, that is what I did on my end as well.

but that is the thing about #43 

The clone special should be leaved as it was before.
and instead 
- Make another clone special command that does the visibility link.

so there will be:
capture190

mottosso:
Ok, then we have some miscommunication on our hands. @CalleHalldin just told me there was trouble implementing this fix and that there was no time to get it done.

If you have a implementation that you have been using in production already, then it would be much better if we merge that one.

LegacyID1991:
No actually my solution isn't working properly and I haven't had the time to fix it. I was gonna look into it this weekend.
So if your solution works then that is great.
But your solution is currently implemented on Clone (Special) which is not what @CalleHalldin wants.
He wants it as a seperat command.
That is the update to #43 that I asked for.

mottosso:
Ah ok. Saw the update just now.

In the future, let's not update posts. Let's always make a new post explaining what we thing different. Otherwise it might happen that the update goes unnoticed, because no notifications are sent on updates.

On your suggestion, I would recommend against it. There is no bottom to this rabbit hole and we might end up with a dozen different Clone commands. I would recommend having one that does everything, like Special, and let the user disconnect unwanted parts afterwards.

For example, with this update, Clone (Special) also connects visibility. The clones meshes are automatically selected, so the user could simply right-click in the channel box on meshes that shouldn't include visibility and break this connection.

This way, there is one command everyone can familiarise themselves with, trust and learn to work with. The alternative is having many commands we forget what they do and learn to distrust in the future. Each function would come with its own possibility of bugs as well.

Let me know your thoughts.

mottosso:
I have to run, I'll merge this. If you still don't agree, don't update your pipeline with this yet and we'll return to make changes. Give it a try.

Issue #55, opened 2017-02-23
Alembic Exporter is not exporting visibility


CalleHalldin:
The exporter does not export visibility animation on the meshes.
This is pretty urgent on the current project since we are using one rig and are just animating visibility on Meshes to turn them on and off to create 8 different characters.

mottosso:
Any updates on this?

mottosso:
Anything to do with publishing will be in the `/plugins` directory, under the name of the given plug-in you would like to update.

- [`extract_animation.py`](https://github.com/mindbender-studio/core/blob/74d0bb3a4f92847a08d069f62e1d9d44b6da129f/mindbender/maya/plugins/extract_animation.py#L41)

Here we can see that Alembic export is done by a function called `export_alembic()` which can be found [here](https://github.com/mindbender-studio/core/blob/74d0bb3a4f92847a08d069f62e1d9d44b6da129f/mindbender/maya/lib.py#L165). In this function, you'll need to add support for visibility, and then you'll be done.

Here is the documentation for Maya's Alembic Export MEL command that we use.

```bash
AbcExport -h;
// AbcExport [options]
Options:
-h / -help  Print this message.

-prs / -preRollStartFrame double
The frame to start scene evaluation at.  This is used to set the
starting frame for time dependent translations and can be used to evaluate
run-up that isn't actually translated.

-duf / -dontSkipUnwrittenFrames
When evaluating multiple translate jobs, the presence of this flag decides
whether to evaluate frames between jobs when there is a gap in their frame
ranges.

-v / -verbose
Prints the current frame that is being evaluated.

-j / -jobArg string REQUIRED
String which contains flags for writing data to a particular file.
Multiple jobArgs can be specified.

-jobArg flags:

-a / -attr string
A specific geometric attribute to write out.
This flag may occur more than once.

-df / -dataFormat string
The data format to use to write the file.  Can be either HDF or Ogawa.
The default is Ogawa.

-atp / -attrPrefix string (default ABC_)
Prefix filter for determining which geometric attributes to write out.
This flag may occur more than once.

-ef / -eulerFilter
If this flag is present, apply Euler filter while sampling rotations.

-f / -file string REQUIRED
File location to write the Alembic data.

-fr / -frameRange double double
The frame range to write.
Multiple occurrences of -frameRange are supported within a job. Each
-frameRange defines a new frame range. -step or -frs will affect the
current frame range only.

-frs / -frameRelativeSample double
[...]

PR #57, opened 2017-02-25
Almbic exporter now exports visibility

LegacyID1991:
This is related to #55 

changes has been made to:
> [_extract_animation.py_](https://github.com/mindbender-studio/core/blob/master/mindbender/maya/plugins/extract_animation.py#L50)
> [_lib.py_](https://github.com/mindbender-studio/core/blob/master/mindbender/maya/lib.py#L199)

Changes will result in that when **publishing** an animation from _PYBLISH_ the exporter now also writes the animated visiblilty to the exported alembic.


mottosso:
Hi Arvid, I've put together a little tutorial on how to avoid getting all of those commits in the commit history every time you make a PR.

[arvidguide.zip](https://github.com/mindbender-studio/core/files/801872/arvidguide.zip)

It'll enable you to work on features individually and have better control over what is in your PR. Let me know if anything is unclear.

LegacyID1991:
I watched it, no real uncertainties now but there will probably be some the first couple of times.
Do you want me to take this PR down and make a new one ❓ 
Or can we merge this one and I will make the next PR much cleaner ❔

mindbender/maya/lib.py
```diff
@@ -166,6 +166,7 @@ def export_alembic(nodes,
                    file,
                    frame_range=None,
                    uv_write=True,
+                   write_Visibility=True,
```
mottosso: Use all lowercase please, so as to conform to the surrounding naming convention - a.k.a. snake_case.

mottosso: I'm also thinking we avoid the bad naming convention used by the original `AbcExport` and prefix each of the 6 optional outputs with `write_`, including `uv_write`.

The 6 optionals being:

- uvWrite
- writeColorSets
- writeFaceSets
- writeVisibility
- writeCreases

Let's rename `uv_write` to `write_uv` and make `write_Visibility` into `write_visibility`. What do you think?

LegacyID1991: you are right, it will look much better after these changes, I am on it

LegacyID1991:
how is that?

mottosso:
Looks good to me. Have a look at getting a test in there. I can help you tomorrow if you need.

mottosso:
Perfect.

Ok, so in order to make the test, I couldn't rely on hardcoding of a path like in the [original example](https://github.com/mindbender-studio/core/issues/55#issuecomment-282494913). Although that was a very good way of getting a first draft out so do keep doing that if you need to.

Instead, in order for the test to last, we need to use the abstractions provided by the pipeline in order to (1) hunt down the published asset and (2) import it into Maya.

In order not to depend on hardcoding the path, I needed the loader to give me something back. A result of the importing, so that I could use that to query whether or not the export was successful. This was where #58 came up. Before, loading an asset returned nothing, which made it difficult to accurately find the loaded Alembic and query its visibility.

So, if all that makes sense, I'm happy with this and can merge if you are too.

LegacyID1991:
So now then this PR resolves both #58 and #55 
and now with [this](https://github.com/mindbender-studio/core/pull/57/commits/6f4a2cc3b3d0417c2cd05c13b0b1991cca2f29e7) we have a working test ?

mottosso:
Exactly!

LegacyID1991:
**UPDATE**
This PR is related to #55 and #58 

changes has been made to:

> [_extract_animation.py_](https://github.com/mindbender-studio/core/blob/master/mindbender/maya/plugins/extract_animation.py#L50)
> [_lib.py_](https://github.com/mindbender-studio/core/blob/master/mindbender/maya/lib.py#L199)
> [_pipeline.py_](https://github.com/LegacyID1991/core/blob/a2065258f2e1efd295ffa1f31c3d639a5f9c061f/mindbender/maya/pipeline.py#L250)
> [_test_workflow.py_](https://github.com/LegacyID1991/core/blob/master/mindbender/maya/tests/test_workflow.py)

Changes will result in that when publishing an animation from PYBLISH the exporter now also writes the animated visiblilty to the exported alembic.
Implements a test to be run by Travis CI to test the integrity of this new addition and future changes to this plugin

PR #60, opened 2017-02-27
New feature inspired of #12

### LINKED ISSUE #12: Fixed Task Names
### Goal

Reduce errors in spelling of tasks - e.g. `modeling` versus `modling` and enable custom functionality relative a given task - e.g. a special "modeling" menu for when modeling and a "lookdev" button for those performing lookdev.

### Implementation

One option might be to produce a `.bat` executable per task, and call upon these when starting an application.

```bash
$ modeling maya
Launching Maya...
$ lookdev katana
Launching Katana..
$ modling maya
ERROR: Unrecognised ocmmand "modling"
```

LegacyID1991:
related to #12 

Before when entered a project and an asset the user was free to type any task and have that app launch, creating a new folder and this made it possible for the user to make misstakes. exampel:
first time they entered this input:
```maya Modeling```

> maya Modeling
> creating task "Modeling"

next time the misspelled and typed
```Maya Modelling```
> Maya Modelling
> creating task "Modelling"

this fix makes so that only acceptable input in the API terminal from already created bat files as it is with projects and assets

this PR also adds a creation API called [_new.bat_](https://github.com/LegacyID1991/core/blob/fix12/bin/new.bat)
this allows the user to create projects, assets from the terminal.
example:
```

   Active Project: hulk
   You are in: film

  1000
  2000

  1. Type first letters of asset
  2. Press [TAB] to auto-complete

 --------------------------------------

$ new shot sh010
```
this will output:
```
new shot sh010 created
```
and now the artist can use:
```
sh010
```

This also works with APPS:

```new maya modeling```
adds the bat _maya modeling.bat_
and after that the artist can use either the bat or free string input to enter that task,
they cannot enter a other task though as described before, exampel:

```maya modelling```
> maya modelling
> ERROR: Unrecognised command "maya modelling"

This fixes #12 and also adds functionality to terminal API so that these limits also have a solution.

### CODE REVIEW

File: bin/_mkasset.bat [hunk 1]
```diff
@@ -16,6 +16,8 @@ set MINDBENDER_ASSETPATH=%2%MINDBENDER_ASSET%
 
 If not exist %MINDBENDER_ASSETPATH% (
 	mkdir %MINDBENDER_ASSETPATH%
+	mkdir %MINDBENDER_ASSETPATH%\work
+	mkdir %MINDBENDER_ASSETPATH%\publish
```
mottosso: There is no need to create the publish folder. It is automatically created when necessary.

bin/_mkasset.bat
```diff
@@ -24,30 +26,31 @@ title %MINDBENDER_PROJECTPATH% / %MINDBENDER_ASSET%
...
 echo   For example:
 echo+
 echo   $ maya animation
-echo   $ houdini sim
 echo   $ nuke comp
 echo+
+
+:: List available tasks
+setlocal enabledelayedexpansion
+set FOLDERQUERRY=%MINDBENDER_ASSETPATH%
```
mottosso: Typo, one R too many.

bin/maya.bat
```diff
@@ -23,9 +23,11 @@ if "%MINDBENDER_ASSET%"=="" goto :missing_set_asset
 :: If user forgets to include task with "maya"..
 if "%1"=="" goto :missing_task
 
-set MINDBENDER_WORKDIR=%CD%\work\%1\%USERNAME%\maya
+if not exist "%1" goto :missing_taskfolder
+
+set MINDBENDER_WORKDIR=%CD%\%1\%USERNAME%\maya
 if Not exist %MINDBENDER_WORKDIR% (
-	echo Creating new task "%1"..
+	echo Creating folder for task "%1"..
```
mottosso: I would keep it the way it was. The fact that tasks are really folders is an implementation detail that the user does not need to know about and one that may change later.

bin/new.bat
```diff
@@ -0,0 +1,76 @@
...
+if "%1"=="asset" goto :copyAsset
+
+if "%1"=="shot" goto :copyAsset
+
+if "%1"=="maya" goto :createMaya
+
+if "%1"=="nuke" goto :createNuke
+
+:usage
+echo Enter what you want to create
```
mottosso: Elaborate on this 

Provide detailed instructions along with examples of how to use it. Picture someone in Brazil encountering this for the first time and not having been told how to use it. He should be able to understand the instructions on his own

bin/new.bat
```diff
@@ -0,0 +1,76 @@
...
+goto :eof
+
+:missingname
+echo Enter the name of what you want to create
```
mottosso: Same goes for here, name if what? Give examples.

bin/new.bat
```diff
@@ -0,0 +1,76 @@
...
+goto :eof
+
+:copyProject
```
mottosso: Stick to one naming convention. In the other bat files you use snake_case.

bin/new.bat
```diff
@@ -0,0 +1,76 @@
...
+ren ProjectName.bat %2.bat
+echo new project %2 created
+goto :eof
+
+:copyAsset
+if not "%CD%"=="%MINDBENDER_ROOT%\%MINDBENDER_SILO%" goto :missing
+>NUL copy %MINDBENDER_CORE%\template\AssetName.bat .
+>NUL ren AssetName.bat %2.bat
+if "%1"=="shot" echo new shot %2 created
+if "%1"=="asset" echo new Asset %2 created
```
mottosso: Consistency, why is Asset title case when project and shot is not?

bin/_mkasset.bat
LegacyID1991: [fixed](https://github.com/mindbender-studio/core/pull/60/commits/b4ea425e675e26b3fae971f88cbd1ccf7340988c)

bin/maya.bat
LegacyID1991: I changed it to not confuse the user that they used creation API (_new.bat_) and then when use that which they created it says creating your task again.

bin/new.bat
LegacyID1991: Right, on it

bin/new.bat
LegacyID1991: same as above on it

bin/new.bat
LegacyID1991: you are right. will fix that.

bin/new.bat
LegacyID1991: on it

bin/maya.bat
mottosso: I wouldn't worry about that, as the person using `new` - like Calle - won't be the same as the one using the created bat files, like Derek.

But being more specific is a good idea, in this case we aren't creating a new task, we are entering a task for the first time. If Calle has created `maya modeling` but Derek hasn't modeled before, we should tell Derek that this will be the first time.

bin/_mkasset.bat
LegacyID1991: it is because we enter "work" folder.
we could ignore creating "publish".

bin/maya.bat
LegacyID1991: it now says:
> Creating new task 
But for comedy sake, I will just leave this here:

I TOLD YOU SO ❗️

bin/_mkasset.bat
LegacyID1991: the name of this [commit](https://github.com/mindbender-studio/core/pull/60/commits/b4ea425e675e26b3fae971f88cbd1ccf7340988c) should be "cleanup typo" not "cleanup type"
I am the greatest at typing.

bin/_mkasset.bat
LegacyID1991: I missread I see that now,
I did make the change though.

mottosso:
If you are happy with this, I'm happy too. Let me know and I'll merge.

LegacyID1991:
I approve.
Merge away.
Make it so..

mottosso:
image

Issue #34, opened 2017-01-31
Pyblish still thinks deleted container manager is up for publishing


LegacyID1991:
> "Managern deletar referensen i reference editorn men referensen finns fortfarnade i managern och pyblish tror fortfarnade att asset finns kvar. Antagligen pga att "Asset set" inte deletas."

**Process to find bug:**
_When working with the container manager:_
- _You load a asset into a scene_
- _Then you want to delete the asset from the scene._
- _Deleting it from the manager does the job for the reference editor._
- _The asset is still present in the manager window_
- _Open up the publishing window_
- _The asset is there as well_

LegacyID1991:
Solution could be that the asset manager doesn't delete the maya sets for the asset.

mottosso:
This this apply to assets other than Rigs?

Have a look at updating the Collector for this. Basically, if an instance is empty, maybe not collect it.

The "problem" is that an Instance is created for the rig, but isn't technically part of the rig. So when you go ahead and delete the rig, the Instance would still remain and get picked up by the Collector.

LegacyID1991:
So I am back on this.

LegacyID1991:
I will return to this one during the weekend or next week,
I am currently working on better info about what goes wrong in this issue.

PR #63, opened 2017-03-01
fix 34

LegacyID1991:
This is related to #34 

Files altered is 
[_collect_instances.py_](https://github.com/LegacyID1991/core/blob/master/mindbender/maya/plugins/collect_instances.py)

Addded a check:
- If **sets** is empty the **set** is skiped
- print skipd sets into log

### DIFF
--- a/mindbender/maya/plugins/collect_instances.py
+++ b/mindbender/maya/plugins/collect_instances.py
@@ -61,6 +61,11 @@ class CollectMindbenderInstances(pyblish.api.ContextPlugin):
                               objectsOnly=True):    # Return objectSet, rather
                                                     # than its members
 
+            # Check if objectset is empty
+            if cmds.sets(objset, q=True) is None:
+                self.log.info("Skiped following Set: \"%s\" " % objset)
+                continue
+
             if not cmds.objExists(objset + ".id"):
                 continue
 
--- a/mindbender/maya/plugins/collect_instances.py
+++ b/mindbender/maya/plugins/collect_instances.py
@@ -61,8 +61,8 @@ class CollectMindbenderInstances(pyblish.api.ContextPlugin):
                               objectsOnly=True):    # Return objectSet, rather
                                                     # than its members
 
-            # Check if objectset is empty
-            if cmds.sets(objset, query=True) is None:
+            is_empty = cmds.sets(objset, query=True) is None
+            if is_empty:
                 self.log.info("Skipped following Set: \"%s\" " % objset)
                 continue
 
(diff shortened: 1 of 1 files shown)


mottosso:
It's not necessary to write which files were altered, they are already documented in the tab at the top.

The rest is good though!

LegacyID1991:
Alright, I thought it looked good, but maybe it becomes to redundant on such a small PR.

LegacyID1991:
Wanna merge this one and close #34 ?

mindbender/maya/plugins/collect_instances.py
```diff
@@ -61,6 +61,11 @@ def process(self, context):
                               objectsOnly=True):    # Return objectSet, rather
                                                     # than its members
 
+            # Check if objectset is empty
+            if cmds.sets(objset, q=True) is None:
```
mottosso: Tiny nitpick, but could you change `q=True` to `query=True`? This way (1) arguments are consistent across the board and (2) it will be undeniable understandable by the next person who may not have used the `sets()` command before.

mindbender/maya/plugins/collect_instances.py
```diff
@@ -61,6 +61,11 @@ def process(self, context):
...
+                self.log.info("Skiped following Set: \"%s\" " % objset)
```
mottosso: Typo, `Skiped` should be `skipped`.

mindbender/maya/plugins/collect_instances.py
```diff
@@ -61,6 +61,11 @@ def process(self, context):
                               objectsOnly=True):    # Return objectSet, rather
                                                     # than its members
 
+            # Check if objectset is empty
```
mottosso: In this case, no need to explain what this does. It should be obvious from the command itself.

mottosso: Alternatively, to make it really clear, I would suggest something like this.

```python
is_empty = cmds.sets(objset, query=True) is None
if is_empty:
  continue
```

LegacyID1991:
how is that ?

mottosso:
Thanks!

mottosso:
Did this apply to anything but rigs btw?

LegacyID1991:
I am not sure that I understand.

Anything but rigs ?
with this fix the instance collector will ignore every empty **maya sets** with a mindbender family.

you mean did the issue revolve around anything other than rigs,
answer to that is that I don't know. I think it did.
The issue was more that the object that you removed within the **container manager** stayed there, giving the artist false information that the removal had not worked. when in fact it had.

mottosso:
The collector doesn't pick up containers, only instances.

The only instance created when creating a container (i.e. loading an asset) is rigs. Loading anything else and deleting it shouldn't have had this problem.

It's important we figure this out, as I've been assuming this was the cause. If it isn't, then there is something else wrong.

LegacyID1991:
Then keep #34 open.

I will do some testing tonight and see if I can delete a loaded rig from a anim scene.

mottosso:
Ok, let me know.

Issue #61, opened 2017-02-28
Launcher GUI


mottosso:
### Goal

- Simplify entering assets for artists
- Simplify creating assets for supervisors
- Enable implementation of #25 and #31
- Notify local power users (e.g. Arvid) of updates to mindbender-core
- Notify remote users of updates to mindbender-setup

### Implementation

The GUI should feature a method of choosing (1) project, (2) asset and (3) app along with the ability to create new (1) projects and (2) assets.

```bash
 ___________________________________________
|                                        _ x|
|___________________________________________|
|project1      |assets| shots|app1          |
|project2      |______|______|app2          |
|...           |asset1       |...           |
|              |asset2       |              |
|              |...          |              |
|              |             |              |
|              |             |              |
|              |             |              |
|_____________+|____________+|______________|
```

It should also facilitate editing of projects and assets, such as frame rate and start/end frame, similar to the Attribute Editor of Maya.

```bash
 _____________________________________________________________
|                                                          _ x|
|_____________________________________________________________|
|project1      |assets| shots|app1        |key    value       |
|project2      |______|______|app2        |key    value       |
|...           |asset1       |...         |..     ...         |
|              |asset2       |            |                   |
|              |...          |            |                   |
|              |             |            |                   |
|              |             |            |                   |
|              |             |            |                   |
|______________|_____________|____________|___________________|
```

mottosso:
Occurred to me that we will also want to specify task in here.

```bash
 _____________________________________________________
|                                                  _ x|
|_____________________________________________________|
|project1      |assets| shots|task1        |app1      |
|project2      |______|______|task2        |app2      |
|...           |asset1       |...          |...       |
|              |asset2       |             |          |
|              |...          |             |          |
|              |             |             |          |
|              |             |             |          |
|              |             |             |          |
|_____________+|____________+|_____________|__________|
```

mottosso:
Some inspiration, Colorbleed has got this going for similar purposes.

![image](https://cloud.githubusercontent.com/assets/2152766/23507616/d729e6c2-ff45-11e6-8399-79c9b200530b.png)

mottosso:
We'll also want to keep an eye on launched processes and their terminals. Such as via a separate page listing each started subprocess and its associated standard out and error.

tws0002:
is this explorer example going to share with us ?

mottosso:
Hi @tws0002, the example is courtesy of @BigRoy. I would inquire with him.

[...]

PR #67, opened 2017-03-02
Launcher GUI

mottosso:
First pass at Launcher, will continue to update this thread as development progresses.

For feedback on fundamental layout and function, now is the time.

untitled project

To be written..

1. Style, such as colors, icons and size of things
2. Better inspection of running processes, including killing them
3. Environment management, currently this does not add anything to the environment; the format is something we will need to discuss. Initially, I'm thinking I'll use the `.bat` files as-is for things such as start/end frame etc. But before we're through we should consider a more appropriate and extensible format, such as `ini`.

See #61 for details.

### DIFF
--- a/mindbender/__init__.py
+++ b/mindbender/__init__.py
@@ -2,16 +2,32 @@
 
 Modules in this package may modify state.
 
-Erasing the contents of each container below will completely zero out
+Erasing the contents `_state` will completely zero out
 the currently held state of mindbender-core.
 
 """
 
-_registered_data = dict()
-_registered_families = dict()
-_registered_formats = list()
-_registered_loaders = list()
-_registered_silos = set()
-_registered_loader_paths = set()
-_registered_root = {"_": ""}
-_registered_host = {"_": None}
+_state = {
+
+    # Data associated to a family, such as `startFrame`
+    # for `mindbender.animation`.
+    "data": dict(),
+
+    # Known families, such as `mindbender.rig` and `mindbender.model`
+    "families": dict(),
+
+    # Known formats, such as `.ma` and `.abc`.
+    "formats": list(),
+
+    # Known silos, such as `assets` and `film`.
+    "silos": set(),
+
+    # Known absolute paths to loaders
+    "loader_paths": set(),
+
+    # Current root
+    "root": "",
+
+    # Current host
+    "host": None,
+}
--- a/mindbender/__init__.py
+++ b/mindbender/__init__.py
@@ -9,6 +9,9 @@ the currently held state of mindbender-core.
 
 _state = {
 
+    # Applications, such as `maya` and `nuke`.
+    "apps": dict(),
+
     # Data associated to a family, such as `startFrame`
     # for `mindbender.animation`.
     "data": dict(),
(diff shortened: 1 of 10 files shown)


tws0002:
can i test this ?

mottosso:
You can run it, but it currently doesn't do much other than launch Maya. It doesn't launch it into the pipeline, like `mb.bat` is.

Currently we are evaluating appearance only.

Wait till tomorrow, and you'll be able to test it out better.

CalleHalldin:
Hey looks great! I will show it to the rest of the office tomorrow. Would it be possible to define and create "users" instead of the pipeline looking at the computer network name (or whatever it looks at)? Also to be able to create projects, assets, shots from this GUI. Not sure how far you are planning to take this. The current state looks great though!

mottosso:
> Would it be possible to define and create "users" instead of the pipeline looking at the computer network name (or whatever it looks at)?

Sure, that's possible. Let's explore the ideal solution first; that is, could you create your user via your OS? As a regular Windows user.

> Also to be able to create projects, assets, shots from this GUI. 

Sure, that's a good idea.

BigRoy:
Two minor tips:

- I'd recommend avoiding the need for the _load_ button and instead just allow double clicking the name of the app on the right hand side. (Use it like a regular "browser")
- To give instant feedback to the Artist it works very well to add a statusbar that displays brief messages from the monitor (the log) at the bottom of the UI. Then starting the application instantly results in a "Launching: x" message.

> You can run it, but it currently doesn't do much other than launch Maya. It doesn't launch it into the pipeline, like mb.bat is.

Our launcher is launching the `.bat` files to start within the pipeline. I'm assuming this would be done here as well?

mottosso:
> Our launcher is launching the .bat files to start within the pipeline. I'm assuming this would be done here as well?

I'm still working on the most straightforward solution to this.

Currently, you register names of executables with the launcher, such as `maya2016`. It then expects to find this on your `PATH`. The problem is that they also need to set the appropriate `userSetup` path for Maya, and do it differently from Nuke. So there's configuration to be made in both the launcher and the `.bat` files.

I experimented with having all of it in a single configuration, but it got complicated fast.

How are your .bat files looking?

mottosso:
To your suggestions, I like it. Will do that, thanks!

BigRoy:
> How are your .bat files looking?

In our case the launcher has a set of "root paths" that can be registered that have a suite of `.bat` files alongside `.png` icons, these are the tools that are then listed in the launcher. We have the `launcher` roots one folder down from the pipeline `bin` (pipeline .bat files): `bin/launchers/maya2016.bat`. This would then call `setup_environment.bat` from the folder above, etc. Does that make sense?

E.g. the launcher for `maya2017.bat` would look like:

```batch
@echo OFF
pushd %~dp0..\..

call set_environment.bat
call python_environment.bat
call maya2017_environment.bat
call maya2017.bat %*
```

mottosso:
> Does that make sense?

Perfect sense, I would have done it the same way.

Continuing the project in a separate repository here: 

- https://github.com/mindbender-studio/launcher/pull/1

PR #77, opened 2017-03-06
This is suppose to fix #69

### LINKED ISSUE #69: Include history and connections with lookdev assets
At the moment lookdev only publishes shadingGroups.

Octane has problem with sss and displace, it brakes each other. Solution rightnow is to apply an iDisplace, which is a mesh texture displace deformer (iDeform plugin). 

This is the workflow at the moment:
-Import desired defaultMesh.
-CloneSpecial.
-Polysmooth all cloned meshes. (Deformer nr1, needed for iDisplace to have enough poly to displace as accurate as possible)
-iDisplace (deformer nr2)

Wish: Deformers that are assigned to specific mesh in lookdev scene also gets published and assigned when loaded into render scene assembly such as the materials. 

Is this possible?

FYI: The order of Polysmooth and iDisplace are important. Also as a note is that is it possible to have one iDeform and connect it to several meshes, while the polysmooth is only per mesh assignable. It would of been nice to have one polymesh operation which has several meshes attached to it as well.

LegacyID1991:
This is the a beginning to solve #69 

added new function:
- lists valid shaders in scenen
- add valid shaders to directory

more will be added in comments below

### DIFF
--- a/mindbender/maya/lib.py
+++ b/mindbender/maya/lib.py
@@ -416,6 +416,56 @@ def apply_shaders(relationships):
             cmds.sets(meshes, forceElement=shader)
 
 
+def displacement_shader_dir(nodes):
+
+    valid_nodes = cmds.ls(nodes,
+                          long=True,
+                          recursive=True,
+                          showType=True,
+                          objectsOnly=True,
+                          type="transform")
+
+    meshes = {}
+    for mesh in valid_nodes:
+        shapes = cmds.listRelatives(valid_nodes[0],
+                                    shapes=True,
+                                    fullPath=True) or list()
+
+        if shapes:
+            shape = shapes[0]
+            if not cmds.nodeType(shape):
+                continue
+
+            try:
+                selmeshes = cmds.getAttr(mesh + ".outmesh")
+
+                if selmeshes not in meshes:
+                    meshes[selmeshes] = list()
+
+                meshes[selmeshes].append(mesh)
+
+            except ValueError:
+                continue
+
+    meshes_by_shader = dict()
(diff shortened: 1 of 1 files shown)


mindbender/maya/lib.py
```diff
@@ -416,6 +416,56 @@ def apply_shaders(relationships):
...
+
+    meshes = {}
+    for mesh in valid_nodes:
+        shapes = cmds.listRelatives(valid_nodes[0],
+                                    shapes=True,
+                                    fullPath=True) or list()
+
+        if shapes:
+            shape = shapes[0]
+            if not cmds.nodeType(shape):
```
BigRoy: This doesn't seem to actually check anything useful. The `cmds.nodeType` command will return the node type (str) of the node and when the node will always return an actual value (so be True) otherwise if the node does not exist it will raise an error. In this case it will always evaluate as True, and will never skip a node.

You might have meant to do something like this:

```python
cmds.objectType(shape, isAType="mesh")
```

Though note that `cmds.listRelatives` also allows you to filter to types, like meshes: `cmds.listRelatives(node, shapes=True, fullPath=True, type="mesh")` (In this case there's even node need to include `shapes=True` because it already filters to the mesh type anyway.

mottosso: You're right. This PR is a little out of context, me and Arvid are iterating on this and this is a first stage.

I'll be pushing a new PR later today that will deprecate this one, so I'll close this.

LegacyID1991: Thanks for you input BigRoy, I will read up on the "objectType" command.
Thanks for the feedback.

PR #78, opened 2017-03-07
First pass at #69

mottosso:
With limitations, here's a first working version of lookdev with history.

untitled project

<br>

**Publishing Workflow**

1. Load model or cache
2. Clone special, twice. First clone is `input`, second clone is `output` and what you will be working with; applying shaders, deformers, the works.
2. Put `input` in a new objectSet called `in_SEL`
1. Put clones in a group called `ROOT`, just like with models and rigs
1. Create new HistoryLookdev instance, with `in_SEL` and `ROOT`
1. Publish

<br>

**Loading Workflow**

1. Load lookdev
2. Load model or cache
2. Select model or cache reference node
3. Select lookdev reference node
3. Click Mindbender -> Auto Connect

<br>

**Known limitations**

1. **IMPORTANT** Make sure the meshes in `in_SET` have no connections to the original asset apart from `.inMesh`. For example, when you Clone Special, disconnect `.visibility` on the clone.
1. Each cache or model asset must now receive it's own unique lookdev asset, as opposed to applying a single lookdev onto many assets like before
2. File size of lookdev is now larger.
3. No more .json relationships between meshes and shaders. We didn't use it for anything but shaders, so no practical downside to this yet.

<br>

**Install**
Here is how to try this out.

```bash
$ cd mindbender-core
$ git remote add mottosso https://github.com/mottosso/mindbender-core.git
$ git pull mottosso fix69
```

@LegacyID1991 this should probably happen on your development clone, not the production version. But either works.

Let me know what you find.


LegacyID1991:
mb-dev is now working with fix69 checked out

mottosso:
Thanks Arvid

mottosso:
With the Auto Connect feature, this is essentially a first pass at Container Interactions - https://github.com/mindbender-studio/core/issues/73.

mottosso:
Considering this good enough for an initial release.

Further on, I'd like to remove the old `Look` family in Creator, and rename `History Look` to `Look`, but for now both asset types will continue to be supported.

Issue #35, opened 2017-01-31
Alembic import sets timeline in maya


LegacyID1991:
**Disclaimer**
This is not a feature request

> "Import alembic updaterar inte timelinen."

Does the asset loader change settings in maya when an asset is loaded?

Example would it change the timeline _IN_ and _OUT_

mottosso:
I should have mentioned; the time range for alembic caches comes from user-defined attributes in Maya on the Instance, e.g. `MyCharacter01_SET`.

It should have a start and end frame attribute, which comes from the current Maya timeline at the time of creating it. The export will use this attribute, as you can see in the Alembic extractor plug-in.

LegacyID1991:
I dont think we need to use that in the importer.

why dont we set the _IN_ and _OUT_ as we set _FPS_ in _.bat_ files ?

we could have it be both a startup thing for every app and do a _global validator_ for pyblish to check the env variable

wouldn't that be easier ?

mottosso:
Yes, go for it.

What I would do is have both. The environment variable represents a global value, whereas the attribute in Maya can be used to override it by the artist.

I would expect there to be a need to do things like that on a regular basis, both for making quick changes, but also for things like adding pre and post roll to individual shots and characters.

To implement that, you could set the environment variable in the bat file and read it from the extractor, but only if the Maya attribute has not been set.

At the moment the attribute is set automatically in the registration of the family in the pipeline.py of Maya. If you default this to an empty value, the artist could use this to override the environment variable.

Let me know if that makes sense!

LegacyID1991:
I think that for now everything should be set by the bat file, 
and should be validated before publish so that the user hasn't changed their settings by mistake.
It might create problems with that the user has to go to the bat file and change the value if they know the bat file to be incorrect but at least that creates a level of control that I think right now the supervisors wants.

mottosso:
I just noticed I had misunderstood the initial question; that when importing an Alembic, the Maya timeline is set.

Have a look at the animation loader for this. You can either remove it, or implement logic to only do it under special circumstances (e.g. only once, only when timeline is at a default value, only within the first 5 minutes of launching Maya).

The answer I gave above relates to exporting Alembic.

mottosso:
If this is still a problem, feel free to re-open this thread.

PR #82, opened 2017-03-08
Editorial control before publish

LegacyID1991:
**Currently:**
When launching maya for the first time on a new task, the scene gets info and settings from what is defined in the project bats.

**Addition:**
I have added a new feature, inspired by #35, two new plugins
- A context collector for pyblish, which reads the global variables set by the project bats and stores the info in _context.data_

- A validator for maya which gives the user the option to check if their scenes frame-rate and frame-range is the same as the contexts stored info.

**This is a first pass at this plugin, please check my work to see if you can find improvements.**

**Future improvments:**
- make copy of [validate_project_editorial_info.py](https://github.com/mindbender-studio/core/compare/master...LegacyID1991:additionTo35?expand=1#diff-4a65e400db74a46e2cac1446980258e4) and change variable value to app specific commands between [24](https://github.com/mindbender-studio/core/compare/master...LegacyID1991:additionTo35?expand=1#diff-4a65e400db74a46e2cac1446980258e4R24) and [35](https://github.com/mindbender-studio/core/compare/master...LegacyID1991:additionTo35?expand=1#diff-4a65e400db74a46e2cac1446980258e4R35)

- make the [collect_project_editorial_info.py](https://github.com/mindbender-studio/core/compare/master...LegacyID1991:additionTo35?expand=1#diff-dd1f32159ee605b570b8cc4798b5a92e) look at the bat files instead of looking at the global variables, as the global variables are only set on start up, then artist would need to reboot their app to get their editorial info updated.

### DIFF
--- a/mindbender/maya/plugins/validate_project_editorial_info.py
+++ b/mindbender/maya/plugins/validate_project_editorial_info.py
@@ -37,9 +37,12 @@ class ValidateMindbenderProjectEditInfo(pyblish.api.ContextPlugin):
         if scene_fps is None:
             scene_fps = "New Value"
 
-        valid_fps = context.data.get("project_fps")
-        valid_edit_in = context.data.get("project_edit_in")
-        valid_edit_out = context.data.get("project_edit_out")
+        valid_fps = context.data.get("projectFPS")
+        valid_edit_in = context.data.get("projectEditIn")
+        valid_edit_out = context.data.get("projectEditOut")
+
+        if valid_fps is None:
+            return None
 
         assert valid_fps == scene_fps, (
             "The FPS is set to %sfps and not to %sfps"
--- a/mindbender/maya/plugins/validate_project_editorial_info.py
+++ b/mindbender/maya/plugins/validate_project_editorial_info.py
@@ -34,15 +34,17 @@ class ValidateMindbenderProjectEditInfo(pyblish.api.ContextPlugin):
         if scene_fps is None:
             scene_fps = "a strange "
 
-        if "environment" not in context.data:
-            self.log.debug(" environment no set")
-            return
-
         env = context.data.get("environment", dict())
 
-        valid_fps = env.get("mindbenderFps", 0)
-        valid_edit_in = env.get("mindbenderEditIn", 0)
-        valid_edit_out = env.get("mindbenderEditOut", 0)
+        valid_fps = env.get("mindbenderFps")
+        valid_edit_in = env.get("mindbenderEditIn")
+        valid_edit_out = env.get("mindbenderEditOut")
+
+        skip_on_none = [valid_fps, valid_edit_in, valid_edit_out]
+
+        if None in skip_on_none:
+            self.log.debug(" environment no set")
+            return
 
         assert int(valid_fps) == int(scene_fps), (
             "The FPS is set to %sfps and not to %sfps"
--- a/mindbender/maya/plugins/validate_project_editorial_info.py
+++ b/mindbender/maya/plugins/validate_project_editorial_info.py
@@ -3,7 +3,8 @@ import pyblish.api
 
 
 class ValidateMindbenderProjectEditInfo(pyblish.api.ContextPlugin):
-    """
+    """ Checks your scene with editorial info
+
     All the info that gets validated
     has been set by the projects
     bat files.
@@ -16,35 +17,38 @@ class ValidateMindbenderProjectEditInfo(pyblish.api.ContextPlugin):
     label = "Validate Project Edit Info"
     optional = True
     order = pyblish.api.ValidatorOrder
-    hosts = ["maya"]
 
     def process(self, context):
         from maya import cmds
 
-        sceneIn = cmds.playbackOptions(query=True, animationStartTime=True)
-        sceneOut = cmds.playbackOptions(query=True, animationEndTime=True)
-        sceneFPS = {
-            "12fps" : 12,
-            "game"  : 15,
-            "film"  : 24,
-            "pal"   : 25,
-            "ntsc"  : 30,
-            "show"  : 48,
-            "palf"  : 50,
-            "ntscf" : 60}.get(cmds.currentUnit(query=True, time=True))
+        scene_in = cmds.playbackOptions(query=True, animationStartTime=True)
+        scene_out = cmds.playbackOptions(query=True, animationEndTime=True)
+        scene_fps = {
+                "12fps" : 12,
+                "16fps" : 16,
+                "game"  : 15,
+                "film"  : 24,
+                "pal"   : 25,
+                "ntsc"  : 30,
+                "show"  : 48,
+                "palf"  : 50,
+                "ntscf" : 60}.get(cmds.currentUnit(query=True, time=True))
 
-        validFPS = context.data.get("project_fps")
-        validEditIn = context.data.get("project_edit_in")
-        validEditOut = context.data.get("project_edit_out")
+        if scene_fps is None:
+            scene_fps = "New Value"
 
-        assert validFPS == sceneFPS, (
-            ("The FPS is set to %s" % sceneFPS) +
-            ("fps and not to %sfps" % validFPS))
+        valid_fps = context.data.get("project_fps")
+        valid_edit_in = context.data.get("project_edit_in")
+        valid_edit_out = context.data.get("project_edit_out")
 
-        assert sceneIn == validEditIn, (
-            ("Animation Start is set to %s" % sceneIn) +
-            (" and not set to \"%s\"" % validEditIn))
+        assert valid_fps == scene_fps, (
+            "The FPS is set to %sfps and not to %sfps"
+            % (scene_fps, valid_fps))
 
-        assert sceneOut == validEditOut, (
-            ("Animation End is set to %s" % sceneOut) +
-            (" and not set to \"%s\"" % validEditOut))
+        assert scene_in == valid_edit_in, (
+            "Animation Start is set to %s and not set to \"%s\""
+            % (scene_in, valid_edit_in))
+
+        assert scene_out == valid_edit_out, (
+            "Animation End is set to %s and not set to \"%s\""
+            % (scene_out, valid_edit_out))
--- a/mindbender/maya/plugins/validate_project_editorial_info.py
+++ b/mindbender/maya/plugins/validate_project_editorial_info.py
@@ -3,14 +3,11 @@ import pyblish.api
 
 
 class ValidateMindbenderProjectEditInfo(pyblish.api.ContextPlugin):
-    """ Checks your scene with editorial info
+    """Checks your scene with editorial info
 
-    All the info that gets validated
-    has been set by the projects
-    bat files.
-    If you are surtain that the info
-    is incorrect, talk to
-    project Supervisor with haste !
+    All the info that gets validated has been set by the projects bat files.
+    If you are certain that the info is incorrect, talk to your project
+    Supervisor with haste!
 
     """
 
@@ -35,23 +32,25 @@ class ValidateMindbenderProjectEditInfo(pyblish.api.ContextPlugin):
                 "ntscf" : 60}.get(cmds.currentUnit(query=True, time=True))
 
         if scene_fps is None:
-            scene_fps = "New Value"
+            scene_fps = "a Strange "
 
-        valid_fps = context.data.get("projectFPS")
-        valid_edit_in = context.data.get("projectEditIn")
-        valid_edit_out = context.data.get("projectEditOut")
+        if "environment" not in context.data:
+            continue
 
-        if valid_fps is None:
-            return None
+        env = context.data.get("environment")
 
-        assert valid_fps == scene_fps, (
+        valid_fps = env.get("mindbenderFps")
+        valid_edit_in = env.get("mindbenderEditIn")
+        valid_edit_out = env.get("mindbenderEditOut")
+
+        assert int(valid_fps) == int(scene_fps), (
             "The FPS is set to %sfps and not to %sfps"
             % (scene_fps, valid_fps))
 
-        assert scene_in == valid_edit_in, (
+        assert int(scene_in) == int(valid_edit_in), (
             "Animation Start is set to %s and not set to \"%s\""
             % (scene_in, valid_edit_in))
 
-        assert scene_out == valid_edit_out, (
+        assert int(scene_out) == int(valid_edit_out), (
             "Animation End is set to %s and not set to \"%s\""
             % (scene_out, valid_edit_out))
--- a/mindbender/maya/plugins/validate_project_editorial_info.py
+++ b/mindbender/maya/plugins/validate_project_editorial_info.py
@@ -22,7 +22,7 @@ class ValidateMindbenderProjectEditInfo(pyblish.api.ContextPlugin):
         from maya import cmds
 
         sceneIn = cmds.playbackOptions(query=True, animationStartTime=True)
-        sceneOUT = cmds.playbackOptions(query=True, animationEndTime=True)
+        sceneOut = cmds.playbackOptions(query=True, animationEndTime=True)
         sceneFPS = {
             "12fps" : 12,
             "game"  : 15,
@@ -42,7 +42,9 @@ class ValidateMindbenderProjectEditInfo(pyblish.api.ContextPlugin):
             ("fps and not to %sfps" % validFPS))
 
         assert sceneIn == validEditIn, (
-            "Animation Start is not set to \"%s\"" % validEditIn)
+            ("Animation Start is set to %s" % sceneIn) +
+            (" and not set to \"%s\"" % validEditIn))
 
-        assert sceneOUT == validEditOut, (
-            "Animation End is not set to \"%s\"" % validEditOut)
+        assert sceneOut == validEditOut, (
+            ("Animation End is set to %s" % sceneOut) +
+            ("not set to \"%s\"" % validEditOut))
--- a/mindbender/maya/plugins/validate_project_editorial_info.py
+++ b/mindbender/maya/plugins/validate_project_editorial_info.py
@@ -35,13 +35,14 @@ class ValidateMindbenderProjectEditInfo(pyblish.api.ContextPlugin):
             scene_fps = "a strange "
 
         if "environment" not in context.data:
-            continue
+            self.log.debug(" environment no set")
+            return
 
-        env = context.data.get("environment")
+        env = context.data.get("environment", dict())
 
-        valid_fps = env.get("mindbenderFps")
-        valid_edit_in = env.get("mindbenderEditIn")
-        valid_edit_out = env.get("mindbenderEditOut")
+        valid_fps = env.get("mindbenderFps", 0)
+        valid_edit_in = env.get("mindbenderEditIn", 0)
+        valid_edit_out = env.get("mindbenderEditOut", 0)
 
         assert int(valid_fps) == int(scene_fps), (
             "The FPS is set to %sfps and not to %sfps"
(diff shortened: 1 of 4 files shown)


mottosso:
Have a look at what errors Travis is giving you.

LegacyID1991:
So it seam to be giving an error because the FPS is not set at all.

Should I alter the plugin ?
or can I alter the test envrioment that travis runs ?

mottosso:
You'll have to decide whether the existence of `MINDBENDER_FPS` is a requirement for the rest of the pipeline to function.

At the moment, with these plug-ins, if the variable is missing the pipeline is unusable.

In some cases, this is what we want. When the functionality is crucial to the overall function of the pipeline. But I'd think twice about whether the frame-rate should be an optional or required feature. If it's optional, you'd need to craft your plug-in as such.

LegacyID1991:
I went with changing the test setup for now.

LegacyID1991:
**ADDITION TO PR**
Previous the setup for testing didn't get any global variables for FPS, edit in and edit out.
Files changed:
[test_workflow.py](https://github.com/LegacyID1991/core/blob/additionTo35/mindbender/maya/tests/test_workflow.py)

Prevoisuly:
```
    api.register_silo("assets")

    # Setup environment
    os.environ["MINDBENDER_ASSETPATH"] = assetdir
    os.environ["MINDBENDER_SILO"] = "assets"


def teardown():
```

and with updates to my both plugins the Travis test should work.

With update:
```
    api.register_silo("assets")

    # Setup environment
    os.environ["MINDBENDER_ASSETPATH"] = assetdir
    os.environ["MINDBENDER_SILO"] = "assets"
    os.environ["MINDBENDER_FPS"] = "25"
    os.environ["MINDBENDER_EDIT_IN"] = ""
    os.environ["MINDBENDER_EDIT_OUT"] = ""


def teardown():
```

LegacyID1991:
Finally.

I ran into a problem with my local docker, therefore the PR is full of trial and horror commits.


How is this then ?

mindbender/maya/plugins/validate_project_editorial_info.py
```diff
@@ -0,0 +1,50 @@
+# validate_project_editorial_info.py
+import pyblish.api
+
+
+class ValidateMindbenderProjectEditInfo(pyblish.api.ContextPlugin):
+    """
```
mottosso: Have a look at this docstring, and make sure it's in [Napoleon](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) format.

Code convention [here](https://github.com/mindbender-studio/core#code-convention)

mindbender/maya/plugins/validate_project_editorial_info.py
```diff
@@ -0,0 +1,50 @@
...
+    If you are surtain that the info
+    is incorrect, talk to
+    project Supervisor with haste !
+
+    """
+
+    label = "Validate Project Edit Info"
+    optional = True
+    order = pyblish.api.ValidatorOrder
+    hosts = ["maya"]
```
mottosso: Since the plug-in resides in the `maya/` directory, this won't be needed.

mindbender/maya/plugins/validate_project_editorial_info.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def process(self, context):
+        from maya import cmds
+
+        sceneIn = cmds.playbackOptions(query=True, animationStartTime=True)
+        sceneOut = cmds.playbackOptions(query=True, animationEndTime=True)
```
mottosso: Not saying this is wrong, but so you are aware - In Maya, there are two concepts of time. One is the "startTime" and "endTime", the other is is "minTime" and "maxTime".

`startTime` and `endTime` refers to the range slider, whereas `minTime` and `maxTime` refers to the currently playable area with which the animator is working.

We'll need to make sure that what we validate is the same range that is used for actually publishing the animation cache, which may be one or the other.

mindbender/maya/plugins/validate_project_editorial_info.py
```diff
@@ -0,0 +1,50 @@
...
+            "12fps" : 12,
+            "game"  : 15,
+            "film"  : 24,
+            "pal"   : 25,
+            "ntsc"  : 30,
+            "show"  : 48,
+            "palf"  : 50,
+            "ntscf" : 60}.get(cmds.currentUnit(query=True, time=True))
+
+        validFPS = context.data.get("project_fps")
```
mottosso: Python naming convention, stick to `camel_case`.

mindbender/maya/plugins/validate_project_editorial_info.py
```diff
@@ -0,0 +1,50 @@
...
+        sceneOut = cmds.playbackOptions(query=True, animationEndTime=True)
+        sceneFPS = {
+            "12fps" : 12,
+            "game"  : 15,
+            "film"  : 24,
+            "pal"   : 25,
+            "ntsc"  : 30,
+            "show"  : 48,
+            "palf"  : 50,
+            "ntscf" : 60}.get(cmds.currentUnit(query=True, time=True))
```
mottosso: Do not use `.get` unless you are prepared to handle `None`.

In this case, `None` may be returned which would produce a difficult-to-understand message for the user below.

> `The FPS is set to 'None'`

mindbender/maya/plugins/validate_project_editorial_info.py
```diff
@@ -0,0 +1,50 @@
...
+
+        validFPS = context.data.get("project_fps")
+        validEditIn = context.data.get("project_edit_in")
+        validEditOut = context.data.get("project_edit_out")
+
+        assert validFPS == sceneFPS, (
+            ("The FPS is set to %s" % sceneFPS) +
```
mottosso: No need to plus here.

```python
assert validFPS == sceneFPS, (
    "The FPS is set to %s and not set to %s" % (sceneFPS, validEditOut)
)
```

mindbender/maya/tests/test_workflow.py
```diff
@@ -48,6 +48,9 @@ def setup():
     # Setup environment
     os.environ["MINDBENDER_ASSETPATH"] = assetdir
     os.environ["MINDBENDER_SILO"] = "assets"
+    os.environ["MINDBENDER_FPS"] = "24"
```
mottosso: It's great that it passes the tests now, but we really shouldn't make these variables a requirement for the rest of the pipeline.

Think of a way to make these optional. To use and validate when they exist, but do no harm if not. They are an added bonus to the pipeline.

mindbender/plugins/collect_project_editorial_info.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class CollectMindbenderEditInfo(pyblish.api.ContextPlugin):
+    """Store projects editorial info at the time of publish"""
+
+    label = "Collect Project Edit Info"
+    order = pyblish.api.CollectorOrder
+    hosts = ["maya"]
+    families = [
+        # More will come
```
mottosso: No need for notes like these.

If you have an important but *minor* message to your future self or other developer, you may use `# TODO(yourname): Message here` or `# NOTE(yourname): Message here`. But they are to be used sparingly. For major notes, make an issue on GitHub.

mindbender/plugins/collect_project_editorial_info.py
```diff
@@ -0,0 +1,36 @@
...
+        "mindbender.animation"
+    ]
+
+    def process(self, context):
+        import os
+
+        mindbender_fps = os.getenv("MINDBENDER_FPS")
```
mottosso: Considering all environment variables are prefixed `MINDBENDER_`, maybe you can try and think of a way to collect *all* of them in one fell swoop, in some elegant way?

mindbender/plugins/collect_project_editorial_info.py
```diff
@@ -0,0 +1,36 @@
...
+        mindbender_edit_in = os.getenv("MINDBENDER_EDIT_IN")
+        mindbender_edit_out = os.getenv("MINDBENDER_EDIT_OUT")
+
+        if mindbender_fps == "" or mindbender_fps is None:
+            context.data["project_fps"] = 25
```
mottosso: Pyblish naming convention for data is `mixedCase`.

- https://api.pyblish.com/pages/AbstractEntity.data.html

mindbender/maya/plugins/validate_project_editorial_info.py
LegacyID1991: That is why I chose _aimationStartTime_.

I didn't clear any of this with the @CalleHalldin yet but if he wants then he can drop the mic on this one.

I thought it better to take animationStartTime because that is the one that we set in the [_usersetup.py_](https://github.com/mindbender-studio/core/blob/master/mindbender/maya/pythonpath/userSetup.py#L37-L43) 
So to me that was the natural choise to validate as I can see the animator using the playback for their own work and then forgetting to set playback to the edits **In** and **Out**
but the animation start and end should always be set to the shots info.

The biggest reason for this validator is too correct if the animator has by mistake changed the **AnimationStartTime** and **AnimationEndTime**

mottosso: > I thought it better to take animationStartTime because that is the one that we set in the usersetup.py

The userSetup is actually faulty at the moment. If you start Maya, you can see how it sets the endTime, but not the maxTime, which means the playable range is only 1 frame. This is a good example of where this can get confusing.

> So to me that was the natural choise to validate as I can see the animator using the playback for their own work and then forgetting to set playback to the edits In and Out
but the animation start and end should always be set to the shots info.

That's good. Just make sure to run this by animators and see if you can find out what they would naturally expect. Try not to mention how you've done it, try finding out what they would expect. So as to not affect their expectations.

mindbender/plugins/collect_project_editorial_info.py
LegacyID1991: done

mindbender/plugins/collect_project_editorial_info.py
LegacyID1991: Not sure how elegant it is, but done.

mindbender/plugins/collect_project_editorial_info.py
LegacyID1991: removed. done

mindbender/maya/plugins/validate_project_editorial_info.py
LegacyID1991: done

mindbender/maya/plugins/validate_project_editorial_info.py
LegacyID1991: done.

mindbender/maya/plugins/validate_project_editorial_info.py
LegacyID1991: done

mindbender/maya/plugins/validate_project_editorial_info.py
LegacyID1991: done.

mindbender/maya/plugins/validate_project_editorial_info.py
LegacyID1991: done.

LegacyID1991:
New Error:
> **ERROR: Modeling workflow is functional**
> **...**
> **Atomicity not held, aborting.**

and
> **ERROR: Exporting Alembic works**
> **...**
> **Atomicity not held, aborting.**

How do I fix that ?

mindbender/plugins/collect_project_editorial_info.py
```diff
@@ -0,0 +1,40 @@
+# collect_project_editorial_info.py
+import pyblish.api
+
+
+class CollectMindbenderEditInfo(pyblish.api.ContextPlugin):
```
mottosso: Let me know what you think about this.

```python
class CollectMindbenderEnvironment(pyblish.api.ContextPlugin):
    """Store projects editorial info at the time of publish"""

    label = "Collect Project Edit Info"
    order = pyblish.api.CollectorOrder
    hosts = ["maya"]
    families = [
            "mindbender.animation"
    ]

    def process(self, context):
        import os

        if "environment" not in context.data:
            context.data["environment"] = dict()

        prefix = "MINDBENDER_"
        for key, value in os.environ.items():
            if not key.startswith(prefix):
                continue

            key = key[len(prefix)]:]

            # Convert EDIT_IN to EditIn, FPS to Fps.
            key = "".join(k.title() for k in key.split("_"))

            context.data["environment"][key] = value
```

mindbender/maya/plugins/validate_project_editorial_info.py
```diff
@@ -0,0 +1,57 @@
+# validate_project_editorial_info.py
+import pyblish.api
+
+
+class ValidateMindbenderProjectEditInfo(pyblish.api.ContextPlugin):
+    """ Checks your scene with editorial info
```
mottosso: Why the short lines? The goal is 80 chars.

```python
class ValidateMindbenderProjectEditInfo(pyblish.api.ContextPlugin):
    """Checks your scene with editorial info

    All the info that gets validated has been set by the projects bat files.
    If you are certain that the info is incorrect, talk to your project
    Supervisor with haste!

    """

    def process...
```

mottosso:
> ERROR: Exporting Alembic works
Atomicity not held, aborting.
How do I fix that ?

Atomicity means [all plug-ins succeeded](https://github.com/mindbender-studio/core/blob/605477f5258a8f65b5b2a6cecec0b10a5101dc26/mindbender/plugins/integrate_asset.py#L68), one of them must not have succeeded.

I recommend you fix it by making these environment variables optional.

mindbender/plugins/collect_project_editorial_info.py
LegacyID1991: That is really cool.

did a little alteration to it, see if you like my decoration

mindbender/maya/plugins/validate_project_editorial_info.py
LegacyID1991: Because I didn't like how it looks inside of pyblish.
screengrab_2
But that might be because I am on a old laptop with low resolution screen, it might look diffrent on a high res desktop screen.

mindbender/plugins/collect_project_editorial_info.py
```diff
@@ -0,0 +1,36 @@
...
+            if not key.startswith(prefix):
+                continue
+
+            key = key[len(prefix):]
+
+            # Convert EDIT_IN to EditIn, FPS to Fps.
+            key = "".join(k.title() for k in key.split("_"))
+
+            # Decorate key with "mindbender" to match prev namning convention
+            decoration = prefix.replace("_", "")
```
mottosso: Nice, but we can achieve the same result by removing line 26.

```python
key = key[len(prefix):]
```

Let me tell you how it works.

- `len(prefix)` returns the length of the string `prefix`, which is `"MINDBENDER_"`, which is 11 characters long.
- `key[11:]` says "Give me the content of `key`, starting 11 characters in and continuing till the end, `:`. That's because strings are treated like a list, where each character is a member of this list. For example, `[1, 2, 3][0]` returns `1`, and `"123"[0]` returns `"1"`. Using the `:` operator is called "slicing". We are *slicing* a list starting from some index, such as 11, to some other index. E.g. `[1, 2, 3][0:2]` would give us `[1, 2]`. If you don't specify an end, it will assume you meant the last index. `[1, 2, 3][0:] == [1, 2, 3]`. And that's how you slice. `[1, 2, 3][1:] == [2, 3]` and `"MINDBENDER_FPS"[11:] == "FPS"`

In this case, I purposefully discarded `MINDBENDER_` because this is what we use to namespace variables in the environment. It doesn't actually add any value once read, so we don't really need it.

Let me know if this makes sense.

LegacyID1991: I thought a good ide that since we started calling them "mindbender_{variable name}" inside the bats to continue that practice, but if we dont think that it adds any value then I will just remove the decoration.
and since I didn't know any better cooler way to edit the key before it was assigned then I did the noob way of decoration.
And yea I know there is a function "style" called decorators in python, but I haven't used them yet, I have only read about them.

mottosso: Oh no, there was nothing wrong with the way you added `mindbender` to the string, that's how I'd do it as well.

But in this case, `mindbender` was first removed on line 26, and then added again on line 34. Can you spot it?

LegacyID1991: I did
> and since I didn't know any better cooler way to edit the key before it was assigned then I did the noob way of decoration.
I might not have written a good explanation the first time. But yea I saw that you could remove line 26 and get almost same result. but that is why I did my noob version of decoration, because I didn't know how to keep the context.data code etiquette / naming conv without line 26 in works.

mottosso: When in doubt, use examples.

```python
>>> key = "MINDBENDER_EDIT_IN"
>>> "".join(k.title() for k in key.split("_"))
'MindbenderEditIn'
```

Do you want the first letter `M` to be lower-case, so as to match the mixedCase naming convention?

You're right, I did not think of that. Without `MINDBENDER_` it would still not be mixedCase.

```python
>>> key = "EDIT_IN"
>>> "".join(k.title() for k in key.split("_"))
'EditIn'
```

In that case, there isn't to my knowledge any more fancy version than this.

```python
>>> key = "EDIT_IN"
>>> key = "".join(k.title() for k in key.split("_"))
>>> key[0].lower() + key[1:]
'editIn'
```

LegacyID1991:
How about that ?
we now use the fancy way to decorate the _key_ in the _dict()_ but I kept the mindbender in front by removing line 26. 
And tests run without error.

mindbender/plugins/collect_project_editorial_info.py
```diff
@@ -0,0 +1,33 @@
...
+        for key, value in os.environ.items():
+            if not key.startswith(prefix):
+                continue
+
+            # Convert MINDBENDER_EDIT_IN to MindbenderEditIn
+            # or MINDBENDER_FPS to MindbenderFps
+            key = "".join(k.title() for k in key.split("_"))
+
+            # Decorate key to match mixedCase
+            deckey = key[0].lower() + key[1:]
```
mottosso: "dekey"? Let's not get too clever. :)

mottosso:
Looks good, make that final tweak and it's mergable.

mindbender/plugins/collect_project_editorial_info.py
LegacyID1991: so: _deckey_  = _key_
or: _deckey_ = _decoratedKey_

mottosso: Just `key` is fine.

```python
key = key[0].lower() + key[1:]
```

LegacyID1991:
updated to the latest review.

Added to _userSetup.py_, 
Now maya launches with both animation start and playback min set to the info from the bat files.
same for end and max.

>    cmds.setAttr("defaultResolution.width", RESOLUTION_WIDTH)
>    cmds.setAttr("defaultResolution.height", RESOLUTION_HEIGHT)
>    cmds.currentUnit(time=FPS)
>    cmds.playbackOptions(minTime=EDIT_IN)
>    cmds.playbackOptions(maxTime=EDIT_OUT)
>    cmds.playbackOptions(animationStartTime=EDIT_IN)
>    cmds.playbackOptions(animationEndTime=EDIT_OUT)

mottosso:
> Now maya launches with both animation start and playback min set to the info from the bat files.
same for end and max.

That's fine, just keep in mind that this was already implemented in #78, [here](https://github.com/mindbender-studio/core/pull/78/files#diff-59cde0eedfb004dc984ab1e8d2b8bd9aR43).

PR #91, opened 2017-03-23
Small update to Terminal API

LegacyID1991:
update API to list options alphabeticly

### DIFF
--- a/bin/_mkasset.bat
+++ b/bin/_mkasset.bat
@@ -25,29 +25,30 @@ title %MINDBENDER_PROJECTPATH% / %MINDBENDER_ASSET%
 cls
 
 echo+
-echo   %MINDBENDER_ASSET% -----------
+echo    %MINDBENDER_ASSET% -----------
 echo+
 echo+
-echo   1. Type application and task.
-echo   2. Press [Enter] to launch application.
+echo    1. Type application and task.
+echo    2. Press [Enter] to launch application.
 echo+
-echo   For example:
+echo    For example:
 echo+
-echo   $ maya animation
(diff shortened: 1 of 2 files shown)


LegacyID1991:
_This is a update of my description:_

**Previously:**
When the Terminal API listed the available assets it did it in a order which I never really could figure out, but it wasn't alphabetically.

**With this PR:**
The Terminal API now lists the available assets alphabetically, and the option to alter this is now easier as we rely on the **_dir_** command now instead of the sorting of a **_for_** loop.
And the **_dir_** command has more options that can be used in future alteration of this code.

bin/_mkasset.bat
```diff
@@ -25,30 +25,31 @@ title %MINDBENDER_PROJECTPATH% / %MINDBENDER_ASSET%
 cls
 
 echo+
-echo   %MINDBENDER_ASSET% -----------
+echo    %MINDBENDER_ASSET% -----------
 echo+
 echo+
-echo   1. Type application and task.
-echo   2. Press [Enter] to launch application.
+echo    1. Type application and task.
```
mottosso: I think I understand why there are all these small changes on spaces/tabs, but for posterity could you make a note about why you made these?

LegacyID1991:
**Addition:**

In the Terminal API the presentation in __mkproject.bat_ and __mkasset.bat_ the indentation was sometimes 3 spaces and some time 4 spaces.
For continuity and to make it easier to read.
I made all the indentations in the presentation **4 spaces**.

mottosso:
Excellent, fully documented and looking good. :)

Issue #68, opened 2017-03-03
Organizing Work folders


CalleHalldin:
### Goal

Simplify finding the latest working file.

### Motivation

Right now on the current project you constantly have to check all "working folders" - such as `calle/maya/`, `arvid/maya/` and `emre/maya` - to find the latest file which becomes time consuming.

It is a cause of the kind of mistakes where you accidentally open up a version that you think is the latest but isn't.

At our company we share a lot of files between artists so it might be a better idea to group them altogether and treat the company as one big user.

### Implementation

Remove the `user/` directory from the `work/` directory.

```bash
# Before
Bruce/work/modeling/calle/maya/...

# After
Bruce/work/modeling/maya/...
```

Users would  then be responsible for keeping up with a naming convention for their scene files that reflect their intentions and can no longer store files that are personal to themselves in the work directory.

mottosso:
Do take a moment to format your text and keep it objective, it'll save time for everyone reading it.

For posterity, here is the original.

> Right now on the current project you constantly have to check all "Working folders" to find the latest file and there are sometimes 3 or more of them which becomes time consuming and frankly annoying.
it also feels like this way can cause mistakes where you accidentally open up a version that is the latest because you simply missed it or different users naming convention confuses you, which kinda defeats the purpose of a pipeline.
In an ideal world one artist works on one assets for the specific task until its done but that rarely happens in the land of commercials or other fast moving productions, things can be very dynamic. At least at our company.

CalleHalldin:
Alright, i will try :)
Update for this part, 
_Users would then be responsible for keeping up with a naming convention for their scene files that reflect their intentions and can no longer store files that are personal to themselves in the work directory._ 

It would be nice if we could simply have Save function within the pipeline that saves it with som standard name convention?

mottosso:
> It would be nice if we could simply have Save function within the pipeline that saves it with som standard name convention?

That wouldn't be a problem technically, what I'm trying to steer us clear from is building a dependency on the way people work. I'm of the belief that a pipeline should accommodate various styles of working, and organising your own files is a big part of that.

We could add such a button and say to simply use it as a visual guide for when we look for files, but it opens up a temptation to go further, and build tools that perhaps automatically open files and from there we open up for much greater frustration than finding peoples files. At that point, we would limit anyone from simply saving files in a way that the pipeline hasn't foreseen. "Can I save manually, will it break the pipeline?"

I've been in pipelines like this and though built with the best of intentions, it is no fun.

[...]

PR #92, opened 2017-03-27
fix for issue 68

LegacyID1991:
Related to #68 

removed userName folder from work folder structure in accord with issue 68

- This will not however move all of the present work files from old location, they still need to be manually copyd or moved to the new folder when the new tasks are created.

From now on all artists will work in the same app work dir under the tasks for each asset or shot.

### DIFF
--- a/bin/houdini.bat
+++ b/bin/houdini.bat
@@ -23,7 +23,7 @@ if "%MINDBENDER_ASSET%"=="0" goto :missing_set_asset
 :: If user forgets to include task with "houdini"..
 if "%1"=="" goto :missing_task
 
-set MINDBENDER_WORKDIR=%cd%\work\%1\%USERNAME%\houdini
+set MINDBENDER_WORKDIR=%cd%\work\%1\houdini
 if Not exist %MINDBENDER_WORKDIR% (
     echo Creating new task "%1"..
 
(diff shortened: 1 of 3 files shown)


mottosso:
Tested and works well?

LegacyID1991:
Only tested at home PC not in Facility, but yea no issues.
could do a pull to mb-dev for @CalleHalldin to try it out.

mottosso:
Looks good but yeah, better test first.

LegacyID1991:
This Branch is now active on mb-dev.bat

Go nuts @CalleHalldin

mottosso:
Should this be merged?

LegacyID1991:
Me and @CalleHalldin talked it over and we said that it works fine.
But we will wait to merge this till we have completed our current projects, otherwise all of the artists work will need to be copyd to the new folders and this will cause maya starts to missbehave like #97

mottosso:
It really shouldnt cause any issues to have changed where work folders are located, the issue of 97 seems to be that someone has moved assets around after having referenced then into a scene.

You should be able to save your files anywhere, even on the desktop and references should still work. Calle can confirm this.

The goal of every single update to this pipeline should always work with existing and future projects in parallel. If this isn't true then we have messed up somewhere.

mottosso:
Deprecating the old mb.bat

Issue #93, opened 2017-03-27
extract_animation.py exports wrong framerange


LegacyID1991:
### Problem:
Seams that the _extract_animation.py_ doesn't export the correct frame range

This bit of code doesn't seam to generate the correct frame range.
```
frame_range=(instance.data["startFrame"],
             instance.data["endFrame"])
```

### Solution

The code could be changed to this:
```
frame_range=(int(cmds.playbackOptions(query=True, animationStartTime=True)),
             int(cmds.playbackOptions(query=True, animationEndTime=True)))
```
Because that info gets checked by another validator before this code executes.

Maybe there is a better or Cooler way of doing it.
What do you think ?

mottosso:
The time is currently coming from the Instance in Maya. In the attribute editor, Extra Attributes at the bottom, you've got `Start Frame` and `End Frame`.

Those values are coming from [here](https://github.com/mindbender-studio/core/blob/master/mindbender/maya/pipeline.py#L176) which should be doing what you just proposed, but isn't.

Hasn't this been a problem before? This looks suspicious.

As a temporary workaround, whatever the artist sets those values to will be the ones exported in the Alembic.

LegacyID1991:
The data of `Start Frame` and `End Frame` is currently:
`Start Frame = 1.0`
`End Frame = 200.0`

as for has this been a problem before.
This was brought to my attention today. 
>[14:00:15] calle_meindbender: verkar som att alembicen inte tar ut det som är satt igen
[14:00:19] calle_meindbender: börjar från 0
[14:00:33] divrA: okey.
[14:00:45] divrA: jag kollar
[14:01:47] calle_meindbender: den kär från 0 till 200 typ

LegacyID1991:
I could create another validator that makes sure that the value is set correctly.

It will just be a copy of the _validate_project_editorial_info.py_ except it will work on instances instead of context.

mottosso:
For posterity, those issues are:

1. Frame range for once version of an Alembic differs from the frame range of another.
2. Publishing character from one shot in another shot will cause the wrong frame range to be used.
3. Accidental frame range differences due to what an artist may do

The value is written (and stored) with the character so as to avoid these issues. But again, it may not apply to you and the only way to know for sure is to test.

LegacyID1991:
I am not sure I understand the need to store the frame-range with the alembic.

[...]

PR #95, opened 2017-03-27
update to extract_animation.py

LegacyID1991:
### Update

The extractor now gets its frame-range from maya animation start and end.
in accordance with #93

### DIFF
--- a/mindbender/maya/plugins/extract_animation.py
+++ b/mindbender/maya/plugins/extract_animation.py
@@ -31,8 +31,8 @@ class ExtractMindbenderAnimation(pyblish.api.InstancePlugin):
             time=instance.context.data["time"],
             name=instance.data["name"])
 
-        animstart = instance.context.data["mindbenderEditIn"]
-        animend = instance.context.data["mindbenderEditOut"]
+        animstart = cmds.playbackOptions(query=True, animationStartTime=True)
+        animend = cmds.playbackOptions(query=True, animationEndTime=True)
 
         try:
             os.makedirs(dirname)
--- a/mindbender/maya/plugins/extract_animation.py
+++ b/mindbender/maya/plugins/extract_animation.py
@@ -34,6 +34,9 @@ class ExtractMindbenderAnimation(pyblish.api.InstancePlugin):
         animstart = cmds.playbackOptions(query=True, animationStartTime=True)
         animend = cmds.playbackOptions(query=True, animationStartTime=True)
 
+        if animstart == animend:
+            animend = int(animend) + 100
+
         try:
             os.makedirs(dirname)
         except OSError:
--- a/mindbender/maya/plugins/extract_animation.py
+++ b/mindbender/maya/plugins/extract_animation.py
@@ -31,6 +31,9 @@ class ExtractMindbenderAnimation(pyblish.api.InstancePlugin):
             time=instance.context.data["time"],
             name=instance.data["name"])
 
+        animstart = cmds.playbackOptions(query=True, animationStartTime=True)
+        animend = cmds.playbackOptions(query=True, animationStartTime=True)
+
         try:
             os.makedirs(dirname)
         except OSError:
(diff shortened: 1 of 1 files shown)


LegacyID1991:
okey, so that would have been to easy.

What should I do here?

`cmds.playbackOptions(query=True, animationEndTime=True)` returns `1`
making the frame-range =
Start 1
End 1

LegacyID1991:
So this is how I solved the test
```
if animstart == animend:
    animend = int(animend) + 100
```
But we probably should try and fix this with a better solution.

mindbender/maya/plugins/extract_animation.py
```diff
@@ -31,6 +31,12 @@ def process(self, instance):
             time=instance.context.data["time"],
             name=instance.data["name"])
 
+        animstart = cmds.playbackOptions(query=True, animationStartTime=True)
+        animend = cmds.playbackOptions(query=True, animationStartTime=True)
```
mottosso: Try `animationEndTime`.

LegacyID1991:
I corrected my mistake.

But I also had an idea, why no read the context data ?
So in my latest commit the `Start` and `End` is taken from the context collected info,
Is this a bad idea ? 

What do you think ?

LegacyID1991:
And lets ignore that Travis checks failed.

IF the context data read is a good idea then I will think of a way to circumvent that he doesn't have a correct environment.

mottosso:
Just to give you an idea of how it worked before.

1. When an instance is created (via the [Creator GUI](https://github.com/mindbender-studio/core/blob/3e62081449f4e8e6863a37fea0aba932c05744d0/mindbender/tools/creator/app.py#L166), two attributes [are added](https://github.com/mindbender-studio/core/blob/3e62081449f4e8e6863a37fea0aba932c05744d0/mindbender/maya/pipeline.py#L358) to the resulting Maya `objectSet`; [`startFrame` and `endFrame`](https://github.com/mindbender-studio/core/blob/3e62081449f4e8e6863a37fea0aba932c05744d0/mindbender/maya/lib.py#L224).
2. The value these attributes are given is determined by the [registered family](https://github.com/mindbender-studio/core/blob/3e62081449f4e8e6863a37fea0aba932c05744d0/mindbender/maya/pipeline.py#L176).
2. Those attributes were [read by the Collector](https://github.com/mindbender-studio/core/blob/3e62081449f4e8e6863a37fea0aba932c05744d0/mindbender/maya/plugins/collect_instances.py#L84) and stored in the Instance.
3. Finally, the attributes were [used by the Extractor](https://github.com/mindbender-studio/core/blob/3e62081449f4e8e6863a37fea0aba932c05744d0/mindbender/maya/plugins/extract_animation.py#L45).

And hopefully here you can see my mistake. The value given to those attributes was determined at launch, not during the creation of the instance as it should have been.

The reason for doing it this way is to enable you, the developer, to append more customisable attributes to instances that will eventually be editable via a GUI (the Manager). At the moment, these attributes are only accessible via the Attribute Editor in Maya, which makes them a little tricky to notice.

Another reason is for attributes to have a central location and for them to be *reliable*. You should be able to expect certain attributes on a given instance based on its family, so that you can develop plug-ins that *depend* on them.

But we can get back to that once there is a GUI to expose these a little nicer. This approach will work fine for now and solves your immediate problem, which is what is most important. I just wanted you to understand it wasn't this complex by mistake.

> And lets ignore that Travis checks failed.

Let's not. :)

This is what happens when you introduce dependencies in your code (i.e. more required environment variables), and now you'll have to deal with it.

LegacyID1991:
Cool.
Thanks for the info,

> I just wanted you to understand it wasn't this complex by mistake.
 
Ohh no I very took this for a mistake, just didn't see it being used right now, but I understand now that it was there for future implementations and features.
And when we get there it will be AWESOME.

> > And lets ignore that Travis checks failed.

>Let's not. :)

Well I didn't mean forever, ;)  just till I got the answer.

Should I go with the maya read or shall I try and implement a circumvent for Travis CI ?
Maya way is easier..

mottosso:
> Ohh no I never took this for a mistake

It could have been! There certainly *are* mistakes in the code, without a doubt. This bug is an example of that; the value should have been computed during instance creation, but isn't.

I'm certainly not immune to making mistakes, so it's good that you're vigilant.

LegacyID1991:
I went with easy fix for now.

We can always introduce the context data read or when the time is right go with the instance data.

mottosso:
Superseded by #96

Issue #102, opened 2017-04-10
Visualise source file from Loader


mottosso:
### Goal

Facilitate handover and continued work on assets originally created by someone else.

### Motivation

In an ideal world, handover would occur 1-on-1 between two artists. An explicit point in time when one artists *decides* to hand a scene over with enough time and opportunity to explain setup and scene organisation.

Some (most?) times, this isn't possible.

Instead, one artist may have loaded a published asset and decided a change is needed. In such circumstances, the artist would need to know (1) where the published asset came from, i.e. the "work file" and (2) who's responsible, in case there are enquiries or problems along the way towards an update.

### Implementation

Expose source and author metadata in Loader. Anyone should be able to easily spot from where a given published version was produced and as such be able to re-open it and publish a new version.

mottosso:
Implemented in https://github.com/mindbender-studio/core/commit/c24f34a3594f65dc606023ec7e93453db1c24f98

![image](https://user-images.githubusercontent.com/2152766/27075271-8e5cdc6a-5021-11e7-9410-3c05396267b2.png)

Issue #71, opened 2017-03-04
Choose Representation


mottosso:
### Goal

Enable use of available representations.

### Motivation

At the moment, every version comes with at least two representations. One "primary", such as the cache, and one "secondary" such as the original source file. But only one of these are usable via the Loader at the moment, and is pre-defined to get picked.

### Implementation

Add another column for representations.

Issue #70, opened 2017-03-04
Choose Version


mottosso:
### Goal

Facilitate latest, approved and deprecated versions.

### Motivation

At the moment, loading an asset automatically picks the latest version. Sometimes, versions carry specific meaning such as whether or not a version has been reviewed and approved. Other times, some versions are known to work better, perhaps because one was just published broken.

### Implementation

Append another column to the interface.

Issue #66, opened 2017-03-02
Display Additional Data


mottosso:
### Goal

Simplify making informed decisions.

**Data**

- Author
- Version
- Date
- Frames
- Duration
- Pre-roll
- Post-roll

### Inspiration

Colorbleed has got this going.

![image](https://cloud.githubusercontent.com/assets/2152766/23507551/89a8b22a-ff45-11e6-9347-0f9e8beacfc1.png)

CalleHalldin:
ColorBleeds Loader is very sexy i have to say :)

BigRoy:
A first loader port has been merged with #198

PR #116, opened 2017-05-01
Initiate work on Loader 2.0

mottosso:
Implement #66, #70, #71 and #102

**To Tree or not to Tree?**

To show more information, we could either expand on the current number of columns, or switch to a tree-view. What do you think?

image


mottosso:
Here's some reference from previous project, it's what I'll be aiming for in terms of what information is available to the user. Probably won't be including the standalone Inspector or Relationship Editor.

untitled project

Issue #130, opened 2017-05-12
Shot Building


mottosso:
### Goal

Enable publish and load of asset combinations.

### Motivation

Rendering is where all 3d resources come together and transform into 2d images. At the moment, resources are brought into the rendering application - Maya - and saved as a Maya scene file, the "master scene".

In this file, a number of decisions reside.

1. Which assets are to be part of this render, including characters and cameras?
2. Which version of each asset?
3. What should be the relationship between the assets, e.g. shaders to model, fur to animation?
4. What are the settings for the rendering engine currently used?
5. What are the *layers* and settings for each?

The trouble with this approach is that it limits the ability to assemble and define relationships between assets to the rendering artist (in this case, @Stonegrund) where any change must be made (1) by this artist and (2) within this scene.

Number 2 introduces additional complexity if the file is somewhat heavy in that (1) it takes longer to open and edit (2) the risk for file corruption is increased.

But what if anyone could contribute to a final rendered scene, without the intervention of a rendering artist? What if you the animator produce a new version of your animation, and would like to see this rendered? What about the freelance lookdev specialist who would like to see his shader in context of the actual shot, but doesn't understand or dare touch the "master scene".

In short, how do you parallelise the rendering workflow? Or indeed any workflow requiring any form of assembly, e.g. layout or animation?

One potential answer lies in Shot Building.

### Implementation

Building a shot is akin to loading an asset, only this time you load what's called a "package".

A package is a collection of assets along with their versions.

```json
{
  "someAsset": 12,
  "someOtherAsset": 1
}
```

<br>

### Publishing

Publishing of a package could be made via a dedicated tool.

![image](https://cloud.githubusercontent.com/assets/2152766/25997023/11a99bf4-3712-11e7-851c-7f22e116c870.png)

<br>

### Loading

Loading of a package could be done through Loader, via e.g. a drop-down menu at the top.

![image](https://cloud.githubusercontent.com/assets/2152766/25996720/95be7038-3710-11e7-80f7-fcfa4378cf67.png)

In this example, contents of the package is highlighted and open for modification. For example, if the animator doesn't require shaders, or wishes to disregard parts of the package for performance considerations.

Stonegrund:
[...]

Issue #165, opened 2017-06-08
Externalise Configuration


mottosso:
### Goal

Enable use of mindbender-core in varying production requirements, including paths, kinds of assets and default values.

### Motivation

At the moment, paths, content validations, how things are loaded and intended to be created are all embedded into the core repository. I've designed it such that configuration and functionality are conceptually separated but hosted together in order to simplify the development during the early stages.

The configuration is..

1. Publishing plug-ins, hosted in `/plugins`
2. Loader plug-ins, hosted in `{app}/loaders`
3. Creator plug-ins, hosted in `{app}/creators` (Related #157)

Beyond this, there is the inventory for defining and customising project and project content.

The time has now come to expand on what is made possible by the project and doing so means separating company configuration from core functionality. This will enable others, such as @BigRoy to adopt our core, whilst still maintaining their own individual workflow, and then work with us on building a stronger foundation.

### Implementation

The aforementioned configuration is to be hosted in an individual Git repository, [config](https://github.com/mindbender-studio/config). It will continue to grow independently of the underlying code and kept together via [setup](https://github.com/mindbender-studio/setup)

The goal is to enable future studios to build their own `config` repository to customise every relevant aspect of the core production pipeline.

Issue #157, opened 2017-06-08
Creator Plug-ins


mottosso:
### Goal

Facilitate multiple workflows using the same pipeline. In this case, the ability to define families and Instance attributes.

### Motivation

Currently, the families available in core are pre-defined in `mindbender.pipeline.install()`. This means that for anyone else looking to use this pipeline, but have different requirements on what types of assets there are and what they represent, they would need to maintain a patch covering the underlying git commit.

It isn't elegant nor particularly sustainable as any underlying update could throw their customisation. In a word, there wouldn't be an API for the added labour.

This issue is about defining such an API.

### Implementation

Publishing is currently customised via [external Python modules](https://github.com/mindbender-studio/core/tree/bf0e5f5036225fe6be6fa1fed480b97cee64a96f/mindbender/plugins). Each module targets a given family and can be updated and refined without interfacing with the underlying pipeline. I'd like for creator plug-ins to do the same.

Each family is defined through documentation, such as [this.](https://mindbender-studio.github.io/2.0/#contract). 

Each Creator Plug-in then..

1. Implement a given family per application, such as `mindbender.model` for Maya, and how they are later discovered during publishing.
2. Define default attributes for said family, such as `startFrame` for `mindbender.animation`
3. Define default contents, such as top-level `ROOT` group and associated object sets such as `out_SET` for `mindbender.rig`

The overall life-cycle for data in the pipeline then looks like this.

![image](https://user-images.githubusercontent.com/2152766/26918695-f63f8332-4c2a-11e7-9cb2-9b3f6562fbde.png)

Where `writing` is publishing and `reading` is loading.

Where each component is customisable via plug-ins, external to the underlying pipeline implementation. These plug-ins are then to be externalised further into an individual Git repository in which case the entirety of custom behaviour is external and swappable (potentially even at run-time).

mottosso:
This also addresses #83.

BigRoy:
Note that the problem of #181 still is something to take note of when implementing this issue. Specifically the need to "order" the custom attributes and the data. For example `start_frame` and `end_frame` and `step` I expect to be ordered after each other to have a sensible order for the artist. Currently this is being randomly ordered because they are in a `dict`.

mottosso:
Yeah, I expect that to be handled directly in the plug-in, something like this.

```python
cmds.addAttr("name_of_instance", "some_attr1", "some_value")
cmds.addAttr("name_of_instance", "some_attr2", "some_value")
[...]

PR #188, opened 2017-06-13
Loader API

mottosso:
This PR refactores the loader API in order to make it more straightforward and stable. The goal is to start building on these loaders to facilitate "group" of assets to be loaded as one.

A group constitutes a number of individual representations from the same or external assets with some relationship to each other. Such as a model, shaders and rig.

This makes the current level assets "bare" or "low level" in comparison and enabled a rich interface towards integrating various low level assets with each other, such as automatically connecting shaders to meshes, meshes and published curves to the controllers to a rig.

Here's the current "recipe" of a group.

```json
{
    "members": {
        "geometry": [
            "hero:modelDefault.ma",
            "hero:modelLow.ma",
            "hero:rigDefault.ma",
            "shot1:hero01.curves",
            "shot1:hero01.abc"
        ],
        "shader": [
            "hero:lookdevDefault.ma",
            "hero:lookdevAnimation.ma"
        ]
    },
    "presets": {
        "animation": {
            "geometry": "shot1:hero01.curves",
            "shader": "hero:lookdevAnimation.ma"
        },
        "rendering": {
            "geometry": "shot1:hero01.abc",
            "shader": "hero:lookdevDefault.ma"
        },
        "layout": {
            "geometry": "hero:rigDefault.ma",
            "shader": "hero:lookdevAnimation.ma"
        }
    }
}
```

Where the group maintains a record of all interchangeable assets within a given "category" such as "geometry". A pre-defined combination of assets are separately maintained in a "preset" or "configuration" used to determine in which way an a group is to be loaded.

For example, during shot building, it's important to distinguish a build in lighting from a build in animation. Animators want the latest rigs, lighters want the latest caches.

The `asset`, `subset` and `representation` is included in the definition, but the version number is purposefully left out. The idea is to deduce version numbers from either `latest` or `approved` tags.

- See #130 (Shot Building) for motivation.


mottosso:
### Loader API

The API has changed from..

```python
class ModelLoader(api.Loader):
    def process(self, project, asset, subset, version, representation):
       # Load file into host
```

To this..

```python
class GroupLoader(api.Loader):
    families = ["mindbender.group"]
    representations = ["group"]

    def process(self, name, namespace, context):
       # Load self.fname into host
```

Where `families` is matched against the asset being loaded, such as `mindbender.model`, and `representations` are matched against the representation, such as `.abc` or `.group` or `.gpucache`.

`name` and `namespace` were previously generated within the loader itself, but are now passed in from the API `load()` function, that way each loader can strictly generate data - such as Maya nodes - that is later post-processed together, such as being put into a container; something previously done individually in each loader.

`self.fname` is the filename on disk, based on the current root and asset passed into it. This was previously managed in each Loader, but is now unified into the `__init__` of the Loader superclass.

mottosso:
Related #165

mottosso:
Another important change is that this PR also unifies the `name` and `namespace` used by the loaders.

**This is a BACKWARDS INCOMPATIBLE change**

Before, each Loader was responsible for creating their own name and namespace, which resulted in `mindbender.model` getting a unique namespace based on the **asset**, and name based on **subset** whereas `mindbender.animation` got a common name and namespace based on **subset**.

Now names are identical across families.

```python
name = name or subset["name"]
namespace = namespace or lib.unique_namespace(asset["name"] + "_")
```

Where `name` and `namespace` is overridden in case an asset is loaded into the same namespace as another, which is the case when e.g. a rig automatically creates an animation instance for animators.

Existing scenes will operate normally, but newly loaded assets will be given this new name.

mottosso:
Next up, #157 followed by #165.

Issue #83, opened 2017-03-09
Creator Templates


mottosso:
### Goal

Reduce set-up time of new assets.

### Motivation

All assets have some level of boilerplate associated with it, such as all residing within an assembly called `ROOT`. Some assets come with more boilerplate, such as `mindbender.rig`.

We can reduce the time taken to set-up each asset, minimise potential validation errors and further guarantee the uniformity of all assets by automatically providing the user with necessary hierarchy and content per family of asset.

mottosso:
Since Creator plug-ins, this can be managed on a asset-by-asset basis in an individual configuration.

PR #190, opened 2017-06-13
Creator plugins

mottosso:
This PR implements #157, and here's the API for it.

```python
class Creator(object):
    name = None
    label = None
    family = None

    def __init__(self, name, asset, options=None, data=None):
        # Take into account data from user or graphical user interface
        # and establish final name and data.

    def process(self):
        # Create nodes and data in the host
```

**Simple example**

```python
from mindbender import maya

class CreateModel(maya.Creator):
    """Polygonal geometry for animation"""

    name = "modelDefault"
    label = "Model"
    family = "mindbender.model"
```

This minimal implementation is at feature parity with what is available today, via `register_family`, which is now deprecated.

Default data is coming from the superclass `__init__()` and can be overridden, such as in this more advanced example.

```python
from mindbender import maya
from maya import cmds

class CreateAnimation(maya.Creator):
    """Any character or prop animation"""

    name = "animationDefault"
    family = "mindbender.animation"
    label = "Animation"

    def __init__(self, *args, **kwargs):
        super(CreateAnimation, self).__init__(*args, **kwargs)

        self.data.update({
            "startFrame": lambda: cmds.playbackOptions(
                query=True, animationStartTime=True),
            "endFrame": lambda: cmds.playbackOptions(
                query=True, animationEndTime=True),
        })
```

`process()` is implemented once per host, and handles the actual creation of nodes within the host. It can be overridden for complex behaviour, laying the road for Creator Templates (#83).

### DIFF
--- a/mindbender/__init__.py
+++ b/mindbender/__init__.py
@@ -10,8 +10,8 @@ the currently held state of mindbender-core.
 _registered_data = dict()
 _registered_families = dict()
 _registered_formats = list()
-_registered_loaders = list()
+_registered_plugins = dict()
+_registered_plugin_paths = dict()
 _registered_silos = set()
-_registered_loader_paths = set()
 _registered_root = {"_": ""}
 _registered_host = {"_": None}
--- a/mindbender/api.py
+++ b/mindbender/api.py
@@ -29,26 +29,20 @@ from .pipeline import (
     discover,
 
     register_root,
-    register_data,
     register_host,
     register_format,
     register_silo,
-    register_family,
     register_plugin_path,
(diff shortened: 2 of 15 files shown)


mottosso:
This is working without issues, I'll merge this and then we @aardschok can chat about how to recreate the families you have available at the moment, and if there is anything missing!

Also welcome to talk about the interface itself, nothing is yet set in stone. E.g. should there be any instance variables? Should there be an `__init__()`, even though loaders and Pyblish plug-ins do fine without one? Should there be *any* class variables? Etc.

PR #196, opened 2017-06-14
Externalise Configuration

mottosso:
Implements #165.

A first step in this direction, the `userSetup` is reduced from performing application setup into a single command.

```python
from mindbender import api, maya
api.install(maya)
```


mottosso:
At the moment, set-up involves setting the FPS and frame range as per project and asset metadata.

This is where we can enable the end user to customise setup and teardown of an app.

**setup_maya.py**

```python
from avalon import api, maya

class SetupMaya(api.Setup):
  def process(self, project):
    maya.commands.reset_frame_range()
    maya.commands.reset_resolution()
    api.log("Launched Maya")  # For Sentry; user and time implicit
```


**teardown_maya.py**

```python
from avalon import api, maya

class TeardownMaya(api.Teardown):
  def process(self, project):
    api.log("Shutting down..")
```

mottosso:
Menu items should also get externalised, they are already Mindbender-specific and I'd expect others will have different requirements.

image

mottosso:
- DEPRECATED `api.register_silo`
- DEPRECATED `api.register_family`
- DEPRECATED `api.register_data`
- Dropped dependency for `pyblish-maya`
- Moved "File > Publish" to Pipeline menu.
- Renamed "Show Creator" to "Create...", similarly for Load and Manage
- FEATURE Hub for determining configuration globally or per project

### Hub

At first I was thinking that the configuration is a one-off, set by the company once and then that's it. But then it struck me that in order to facilitate new projects using an updated configuration, without breaking a previous project or worrying about backwards compatibility, one could set the configuration *per project*.

So now the project configuration includes the name of a Python package or module that is used as an entry-point for the configuration. The entry point implements the `install()` function for itself, and similarly for any contained software integrations, such as `maya.install()`.

The Hub then looks first for a project configuration, e.g. `project["config"]["name"]` and if none is found then it looks for a global configuration, defined by `MINDBENDER_CONFIG=polly`

The name is then imported and `install()` is run.

To the end user, this requires.

1. That the configuration is available on the PYTHONPATH at the time of launching an application
2. That the configuration implements `install()`.

What happens in install is up to the configuration. In this case, I bundled `install()` together with each plug-in so that paths would remain relative to the installation function. Down the line, I'd expect new versions of configuration to potentially re-use a lot of plug-ins from a previous version. For those situations, the configuration can simply be `polly_1_0` and `polly_1_1` and so on, each referencing plug-ins in perhaps a more global location, rather than being bundled together.

mottosso:
**Install of external configuration**

```bash
$ set PYTHONPATH=/path/to/your/config
$ set MINDBENDER_CONFIG=name_of_your_config
$ maya
```

So long at the module `name_of_your_config` implements `install()`, the rest is taken care of internally!

mottosso:
Menu items are now also separated, `Mindbender` contains pipeline-related items, whereas the configuration provides it's own menu to the side.

image

PR #198, opened 2017-06-15
New Loader

mottosso:
image


BigRoy:
This now adds a `Loader` argument to the `host.load()` method. Instead of **always** running all *loaders* registered for a representation you are now able to cherry pick one.

Preferrably I'd try to move as much as possible towards an api like:
```
host = api.registered_host()
host.load(Loader, representation)
```

Also see `mindbender.tools.cbloader.lib` for an overview that I believe would be "generic behavior".

mottosso:
I left this as-is and got to work on documentation for our new architecture and upcoming contributors. I expect to enable anyone coming into it a smooth start and introduction by the time we deploy our merge into production next week.

When you're back, you should be able to rebase these changes with no conflicts, I tried it here locally. Alternatively a plan merge should work well too.

BigRoy:
@mottosso The current state should be working with the latest state of `core/master` and `config` running Polly.

Could you test this?

Also, the _tests_ should pass now.

mottosso:
Having a look now.

BigRoy:
@mottosso The recent commits implemented this:

- fix usage of `MINDBENDER_ROOT` in one location; refactored to `MINDBENDER_PROJECTS`
- picking a specific version (medium difficulty to implement)
- running the loaders on multiple subsets at once (easy difficulty to implement)

Current state (with cbdarkstyle applied; which is not included yet):

screenshot

mottosso:
Looks great. Is the plan still to implement the sidebar? And how about a dropdown for the representations as well?

The tests are failing because of a change I made yesterday for the docs. I'll fix that asap.

mottosso:
Ok, merge your branch against master again, and it should work.

BigRoy:
> Is the plan still to implement the sidebar?

Yes. But for now it's not a priority since most of the information is available in the main window with the subsets. Some crucial information that might be missing is the _comment_ that was published with the version.

> And how about a dropdown for the representations as well?

What would you expect to differ per representation? Currently the context menu on the subsets list the loaders available for the representations in the version that is set. As such it lists all actions for the available representations.

Feel free to mock up some designs over this for to discuss some ideas.

mottosso:
> What would you expect to differ per representation? Currently the context menu on the subsets list the loaders available for the representations in the version that is set. As such it lists all actions for the available representations.

It'd be cool if choosing a repr was the same as choosing a version.

> Some crucial information that might be missing is the comment that was published with the version.

Yes, very important.

BigRoy:
> It'd be cool if choosing a repr was the same as choosing a version.

It could, but it would also automatically make it slower to pick a specific representation? Or load all selected items as a specific representation. Currently the right-mouse context menu just lists the loaders and the representation it is loading. I don't see what adding the field in the view itself would offer additionally?

If there's a use case it could be doable but then the context menu as done now might be confusing.

mottosso:
No worries, we'll work it out as we go.

mottosso:
How about we add this loader to the menu?

- Load.. (new)
- Load.. (old)

During the transition? That's let artists play with it before we commit fully.

BigRoy:
> During the transition? That's let artists play with it before we commit fully.

Will do that now, yes!

BigRoy:
> Load.. (new)
> Load.. (old)

This is now implemented.

I also fixed the loading functionality for `tools.loader` (the old one) so that it also passes along the loader. Functionality to the artist should be the same.

We're getting there, only thing now is the sidebar in the new loader. :)

mottosso:
Awesome sauce. :)

BigRoy:
Here's a preview of the current state inside of Maya:

asset_loader_sidebar

This should suffice for now?

mottosso:
About the Raw Source.

What I did was show only Source, and have a right-click Copy to Clipboard to fill in the {root}

mottosso:
But it's a detail, this is sufficient to test!

BigRoy:
> What I did was show only Source, and have a right-click Copy to Clipboard to fill in the {root}

Interesting! :)

BigRoy:
And another update! :)

actions

Here it's implemented as actions.

The _copy raw_ action copies the information that is returned directly from the database, for debugging purposes, here's an example of the copied data:

```
{u'_id': ObjectId('59424715c479bb96bd7d58e0'),
 u'data': {u'asset': u'villain',
           u'author': u'Roy',
           u'comment': u'',
           u'endFrame': 250.0,
           u'families': [u'mindbender.animation'],
           u'family': u'mindbender.animation',
           u'files': [u'villain01_SET.abc',
                      u'villain01_SET.curves',
                      u'villain01_SET.history'],
           u'id': u'pyblish.mindbender.instance',
           u'name': u'villain01_SET',
           u'source': u'{root}/debugproject/assets/villain/work/model/maya/scenes/animation_v001.ma',
           u'stagingDir': u'D:\\Projects\\debugproject\\assets\\villain\\work\\model\\maya\\stage\\villain01_SET\\20170615T103633Z',
           u'startFrame': 1.0,
           u'subset': u'villain01',
           u'time': u'20170615T103633Z'},
 u'locations': [],
 u'name': 2,
 u'parent': ObjectId('59424298c479bb96bd7d58da'),
 u'schema': u'mindbender-core:version-2.0',
 u'type': u'version'}
```

Note that this version has a lot of data because the integrator I'm using is just adding the `instance.data` at publish to the database.

mottosso:
Good idea!

BigRoy:
This should be usable and can be tested. Could you have a play with it and see if we should merge?

mottosso:
Sure, having a look now.

mottosso:
Looks good to me.

Thanks for keeping it within a single directory; it makes it a lot easier for me to see what other code it affects (in this case none!). We can easily refactor things later to address duplicate code and things.

Issue #189, opened 2017-06-13
Session() object for global state


mottosso:
### Goal

Group global variables and state together into a single object.

### Motivation

At the moment, state is spread across at least three places.

1. Environment, e.g. `MINDBENDER_PROJECT`
2. Registry, e.g. `api.register_loader_path("c:\some\path")`
3. `io.py`, e.g. `io.activate_project("this_project")`

This makes it challenging to know what the current state is, which complicates tests and maintenance. Communicating and understanding state is also complicated.

Traces of leaked global state:

- [install()](https://github.com/mindbender-studio/core/blob/bb874e8fe34ab88f002b3d334190778a539eb8cf/mindbender/pipeline.py#L40)
- [Tests](https://github.com/mindbender-studio/core/blob/bb874e8fe34ab88f002b3d334190778a539eb8cf/mindbender/tests/test_pipeline.py#L41)
- [Tools](https://github.com/mindbender-studio/core/blob/bb874e8fe34ab88f002b3d334190778a539eb8cf/mindbender/tools/creator/app.py#L154)

### Implementation

Do what Python does.

Python uses environment variables for state as well, but not at run-time. Instead, it reads it *once* upon launch, and stores it internally. `PYTHONPATH` for example is parsed and stored as a list in `sys.path`.

On initialisation of the pipeline, look at the various sources supported, such as the environment, and build a Session object/dictionary containing all state. This is then used at run-time, and can be inspected during debugging, created during testing and replicated in order to simulate more than a single state; such as when browsing assets in one project from another.

```python
session = api.Session(
  project="Hulk",
  asset="Bruce",
  task="modeling",
  loaders="/path/to/loaders",
  creators="/path/to/creators"
)

for asset in session.find({"type": "asset"}):
  print(asset["name"])
```

mottosso:
Or how about?

```python
session = api.Session(
  env={
    "project": "Hulk",
    "asset": "Bruce",
    "task": "modeling",
  },
  paths={
    "loaders": "/path/to/loaders",
    "creators": "/path/to/creators"
  }
)
```

mottosso:
This has been implemented long ago, as `avalon.Session`.

davidlatwe:
Does this related to [session.py](https://github.com/getavalon/core/blob/master/avalon/session.py) ?

mottosso:
Thanks for pointing this out, I didn't think of it; by looks of it, that file has been deprecated.

[...]

PR #205, opened 2017-06-23
Implement new Scene Inventory

BigRoy:
This PR implements a new Scene Inventory to replace the current "manager".

It currently looks like this:

scene_inventory

### Features

- Groups the same representations (asset/subset/version/representation) into a collapsible item. As such many of the same items are shown together.
- Manage versions or remove many items at once by selecting them simultaneously.
- Set a specific version for many items at once.
- Find the actual file on disk (by "exploring" it; currently windows-only)

### DIFF
--- a/avalon/maya/pipeline.py
+++ b/avalon/maya/pipeline.py
@@ -58,7 +58,8 @@ def _install_menu():
         loader,
         manager,
         publish,
-        cbloader
+        cbloader,
+        cbsceneinventory
     )
 
     from . import interactive
@@ -80,7 +81,10 @@ def _install_menu():
         cmds.menuItem("Publish...",
                       command=lambda *args: publish.show(parent=self._parent),
                       image=publish.ICON)
-        cmds.menuItem("Manage...",
+        cmds.menuItem("Manage... (new)",
+                      command=lambda *args: cbsceneinventory.show(
+                          parent=self._parent))
+        cmds.menuItem("Manage... (old)",
                       command=lambda *args: manager.show(parent=self._parent))
 
         cmds.menuItem(divider=True)
--- a/avalon/pipeline.py
+++ b/avalon/pipeline.py
@@ -528,20 +528,20 @@ def debug_host():
     def ls():
         containers = [
             {
+                "representation": "ee-ft-a-uuid1",
                 "schema": "avalon-core:container-1.0",
                 "name": "Bruce01",
-                "asset": "Bruce",
-                "subset": "rigDefault",
+                "objectName": "Bruce01_node",
+                "namespace": "_bruce01_",
                 "version": 3,
-                "silo": "assets",
             },
             {
+                "representation": "aa-bc-s-uuid2",
                 "schema": "avalon-core:container-1.0",
                 "name": "Bruce02",
-                "asset": "Bruce",
-                "subset": "modelDefault",
+                "objectName": "Bruce01_node",
+                "namespace": "_bruce02_",
                 "version": 2,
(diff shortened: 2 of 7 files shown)


mottosso:
Thanks, what about the style, like the icons, is it included?

BigRoy:
> Thanks, what about the style, like the icons, is it included?

Yes, this is the version as it runs in Maya. The icons are QtAwesome as which was already a vendor.

BigRoy:
Also, feature note. To update **all** in the scene.

1. Open this Scene Inventory
2. Ctrl + A (to select all entries)
3. Right click -> update to latest

BigRoy:
@mottosso let me know if you need anything changed prior to a merge.

mottosso:
I'll give this a try in a few mins

mottosso:
This works well.

One critique and I realise it'll be controversial so I'll try and explain it best I can.

image

I think we should remove this.

The reason is that, yes, it is convenient but it encourages users to use Windows Explorer as part of their workflow. Once we're at a stage where this won't be necessary - which we may already be at, and if not let's talk about what's missing - getting this convenience out of their hands can be tough and it will harm our ability to help them through pipeline means.

For example, deleting, renaming or moving assets can't happen from Explorer. With this function, I as an artist would expect that we've thought about what is made possible via the Explorer and facilitate me going there to make changes. It would break.

Copying assets and sending to another artist, or getting the directory path for use elsewhere are all handy features, which are also easily facilitated by pipeline tools, such as a "Copy path". In that case, we could do better than get the absolute path, but also include template information such that a path could be used by someone who isn't local and doesn't share the `{root}` drive, for example.

It also means we can collect statistics on what the user is doing, to gain a better understanding of used/unused features, and collect errors through Sentry in order to help them better.

BigRoy:
> I think we should remove this.

Sure, it's currently there for debugging purposes. I'll remove it so we can merge.

mottosso:
Ok, that was easier than expected. :) Good!

mottosso:
How about a `AVALON_DEBUGGING` for things like that, so that we can still have our tools available to us while we develop?

mottosso:
Or `AVALON_DEBUG`

BigRoy:
> Or AVALON_DEBUG

Sounds like a good idea, but since the feature isn't cross-OS compatible for now I'll remove this particular one. It will also make the code slightly simpler.

---

Also, an additional question the `app` and `task` that are passed to the context I'm currently getting them through: 
```python
    data['app'] = os.environ.get("AVALON_APP", "")
    data['task'] = os.environ.get("AVALON_TASK", "")
```

Is that the way they should be retrieved? Or should I get these differently, e.g. through the api?
The code is near the top in `avalon.tools.cbsceneinventory.app`.

mottosso:
> Is that the way they should be retrieved? Or should I get these differently, e.g. through the api?

That's the way to do it.

State is currently all in the environment. I think we should #189 (`Session()`) in mind for anything state-related. I see more and more use for it, especially for tests and the project generation I'm putting together now.

BigRoy:
Should be ready to go! :)

mottosso:
Small cosmetic tweak; could you convert the ' to "? The OCD in me twerks at the inconsistency. :)

mottosso:
Hah, twerk is not the right word.

image

..or is it?

BigRoy:
I see you twerking. Shall I replace all of the `'` to `"` to get you less twerky? Or specific ones?

mottosso:
All of them please. Death to '

PR #207, opened 2017-06-26
Fix #206, remove registry of formats, return instance from maya Creator.

### LINKED ISSUE #206: New loader loads twice when not clicking on most left column
# Problem

The new loader implemented with #198 has a bug where when clicking on a column besides the first one the loader is triggered twice for the active index.

This is because in code the "active" index is ensured to be inserted at the end, but in this case its previous entry was 'removed' incorrectly from the list. As such, it ends up being triggered twice.

# Solution

Ensure the active index is correctly removed from the selection so it doesn't run twice.

BigRoy:
- Remove registry of formats (which was deprecated and unused currently)
- Fix #206: loader incorrectly loading twice
- Return the instance from `maya.pipeline.Creator.process()` so inheriting subclasses can override _process_, call `super` and then perform additional things on the created instance node.

### DIFF
--- a/avalon/__init__.py
+++ b/avalon/__init__.py
@@ -7,7 +7,6 @@ the currently held state of avalon-core.
 
 """
 
-_registered_formats = list()
 _registered_plugins = dict()
 _registered_plugin_paths = dict()
 _registered_root = {"_": ""}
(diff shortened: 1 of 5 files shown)


mottosso:
Looks good to me, merge?

BigRoy:
Yes please! :)

Issue #107, opened 2017-04-17
Live Context Switch


mottosso:
### Goal

Reduce time taken to begin work on new asset.

### Motivation

Currently, working on an asset involves going through the launcher and booting up Maya from within a given task. Sometimes, an artist may work on many assets of very short turnaround, such as 5 assets an hour. In such cases, booting up Maya can become a bottleneck.

### Implementation

This is open for discussion, there are a few ways to go about it.

For example, there's a space in the bottom-left corner of the Maya window well suited for additional buttons.

![image](https://cloud.githubusercontent.com/assets/2152766/25084938/55d7f8e4-2357-11e7-9946-5c4b669333ed.png)

Buttons could look like the Maya icon and stacked vertically like this.

```
 __     __
|         |
| project |
|__     __|
 __     __
|         |
|  asset  |
|__     __|
 __     __
|         |
|  task   |
|__     __|
```

Each button opening a menu of available choices.

Alternatively, one might add it to the existing Mindbender menu at the top of the Maya window.

BigRoy:
Visually I think this could be great.

Though regarding the current context and switching I do have some questions:

For this to work, do we need to refactor all `os.environ['AVALON_PROJECT']`  and other look ups for Avalon state in publishing, etc. to ise Session? Or do we want the Session to ensure it remains in sync with `os.environ`?

I believe currently we sometimes do: 
```python
current_asset = os.environ['AVALON_ASSET']
```

Should that be refactored to the session? :) Or the other way around, where a change to the session also updates the environment? (In the latter case you could never operate with multiple sessions in a python environment. Is that ok?)

mottosso:
> or this to work, do we need to refactor all os.environ['AVALON_PROJECT'] and other look ups for Avalon state in publishing, etc. to ise Session?

Yes. Some things are already referencing this object, but they all should. This object should read from `os.environ` once, at startup, and then it's the main point of contact for state. Changing it would change the state, `os.environ` shouldn't affect anything.

That's the theory at least. There might be edge-cases where `os.environ` is still needed, but hopefully not!

mottosso:
I would suggest starting small, changing things like the current asset. It would affect where the Creator gets it's default asset name. That'd make a good PR, and help us evaluate the visuals. The other slots can remain locked until they get an implementation, one at a time.

BigRoy:
[...]

PR #208, opened 2017-06-26
Implement AVALON_LABEL

mottosso:
Also an initial pass of #189, Session

`AVALON_LABEL` is used by the menu item in the hosts. The intent is to enable studios to customise the look of it, to merge better with their studio. I'm calling it `Mindbender` for the Mindbender guys.

I've also made a first pass of `api.session` which doesn't make any functional changes, apart from collecting variables used in `pipeline.py`. The goal is to see whether we can reference all state from this one dictionary. At that point, we can start having a look at making so that we can instantiate a new one within the same process, so as to enable #107 and simplified testing amongst other things.

### DIFF
--- a/avalon/api.py
+++ b/avalon/api.py
@@ -25,6 +25,7 @@ from .pipeline import (
     Loader,
     Creator,
     discover,
+    session,
 
     register_root,
     register_host,
@@ -62,6 +63,7 @@ __all__ = [
     "Loader",
     "Creator",
     "discover",
+    "session",
 
     "register_host",
     "register_format",
--- a/avalon/maya/pipeline.py
+++ b/avalon/maya/pipeline.py
@@ -12,7 +12,7 @@ from ..vendor import six
 from ..vendor.Qt import QtCore, QtWidgets
 
 self = sys.modules[__name__]
-self._menu = "avaloncore"  # Unique name of menu, for uninstall purposes
+self._menu = api.session["label"] + "menu"  # Unique name of menu
 self._events = dict()  # Registered Maya callbacks
 self._parent = None  # Main Window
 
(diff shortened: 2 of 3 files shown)


mottosso:
**Usage**

```bash
$ set AVALON_LABEL=Mindbender
$ python -m launcher
```

mottosso:
**Example**

image

BigRoy:
> I've also made a first pass of api.session which doesn't make any functional changes, apart from collecting variables used in pipeline.py.

I think currently the implementation feels off. Since it's a dictionary stored at the module level you're still not allowing multiple sessions at the same time. As such in this way it wouldn't introduce any benefits, would it?

Using sessions shouldn't two tools be able to operate on different sessions within the same python interpreter session? I'd say _yes_. But if not, why would you have a session?

The implementation as described in #189 makes more sense to me, where you instantiate a session object and that's what you keep around instead of the "api" holding on to it.

mottosso:
> As such in this way it wouldn't introduce any benefits, would it?

It does not. That's what I meant with that it didn't make any changes, other than move variables around in `pipeline.py`. The idea is to keep expanding the areas that make use of these variables elsewhere, such as in plug-ins. But I don't think that can be done in one big go.

This is a small step in that direction, I'm expecting it to change into it's own object when the need arises and when we've refactored all references into using this object.

> But if not, why would you have a session?

Short answer, baby steps.

BigRoy:
Sure, in that sense it could be a decent intermediate. Should it also house `app` and `task`?

mottosso:
> Sure, in that sense it could be a decent intermediate. Should it also house app and task?

Yes, let's start with this, as this PR is about LABEL.

Continue thinking about where you use state in your code from now on, and see whether you can refactor it into `api.session`.

PR #210, opened 2017-06-26
Reduce required permissions for backwards compatibility

mottosso:


mottosso:
This avoids the increase in necessary permissions for the user.

PR #211, opened 2017-06-26
Add extra `cbloader` features: order, icons and tips!

BigRoy:
Add extra features to `avalon.tools.cbloader`:

- Loaders can now be sorted by an `.order` attribute on the Loader plug-in. This will influence how they are listed under the right-mouse context menu in the loader.
- Loaders can now have custom icons using a `.icon` string attribute, these should be font-awesome icon names.
- Additionally the icons can be colored using a `.icon_color` attribute. A string specifying a Qt color, this can also be hex formatted.
- The Loader's docstring is now also added the action's tool- and status-tip.

### Functionality

```python
class CustomLoader(api.Loader):
    """This docstring is the tooltip AND statustip!"""
    
    families = ["colorbleed.model"]
    representations = ["ma"]
    label = "Custom label"      # label shown on the action
    order = -10                 # sorting order of the actions menu for loaders
    icon = "code-fork"          # font-awesome icon name
    icon_color = "orange"       # qt color name or hex formatted string
```

### Example

Here's an example image with corresponding loaders in the code below.

actions_new

```python
class ModelLoader(LogDebugLoader):
    """Load a model"""
    families = ["colorbleed.model"]
    representations = ["ma"]
    label = "Reference model"
    order = -10
    icon = "code-fork"
    icon_color = "orange"


class GpuCacheLoader(LogDebugLoader):
    """Import a GPU cache"""
    families = ["colorbleed.model", "colorbleed.animation"]
    representations = ["abc"]
    label = "Import GPU cache"
    order = -1
    icon = "download"
    icon_color = "white"
```

### DIFF
--- a/avalon/tools/cbloader/widgets.py
+++ b/avalon/tools/cbloader/widgets.py
@@ -117,13 +117,13 @@ class SubsetWidget(QtWidgets.QWidget):
                 action.setToolTip(tip)
                 action.setStatusTip(tip)
 
-            # Support font-awesome icons using the `.icon` and `.icon_color`
+            # Support font-awesome icons using the `.icon` and `.color`
             # attributes on plug-ins.
             icon = getattr(loader, "icon", None)
             if icon is not None:
                 try:
                     key = "fa.{0}".format(icon)
-                    color = getattr(loader, "icon_color", "white")
+                    color = getattr(loader, "color", "white")
                     action.setIcon(qtawesome.icon(key, color=color))
                 except Exception as e:
                     print("Unable to set icon for loader "
--- a/avalon/tools/cbloader/widgets.py
+++ b/avalon/tools/cbloader/widgets.py
@@ -90,9 +90,14 @@ class SubsetWidget(QtWidgets.QWidget):
             self.echo("No compatible loaders available for this version.")
             return
 
+        def sorter(value):
+            """Sort the Loaders by their order and then their name"""
+            Plugin = value[1]
+            return Plugin.order, Plugin.__name__
+
         # List the available loaders
         menu = QtWidgets.QMenu(self)
-        for representation, loader in loaders:
+        for representation, loader in sorted(loaders, key=sorter):
 
             # Label
             label = getattr(loader, "label", None)
--- a/avalon/tools/cbloader/widgets.py
+++ b/avalon/tools/cbloader/widgets.py
@@ -1,5 +1,6 @@
 import datetime
 import pprint
+import inspect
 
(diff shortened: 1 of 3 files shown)


mottosso:
How about just 'color' for the icon color?

avalon/tools/cbloader/widgets.py
```diff
@@ -103,6 +110,25 @@ def on_context_menu(self, point):
 
             action = QtWidgets.QAction(label, menu)
             action.setData((representation, loader))
+
+            # Add tooltip and statustip from Loader docstring
+            tip = inspect.getdoc(loader)
```
mottosso: What about `loader.__doc__`?

BigRoy:
> How about just 'color' for the icon color?

Could work, but might be confusing as it could refer to the font/label color?

avalon/tools/cbloader/widgets.py
BigRoy: I chose for `getdoc` because it also cleand up the docstring using `inspect.cleandoc`: https://docs.python.org/2/library/inspect.html#inspect.cleandoc

mottosso:
I'm thinking it may be more than just the color of the icon; it'd be the color associated to this Loader. Could be used for more than its icon.

avalon/tools/cbloader/widgets.py
mottosso: Good call. I've had trouble with that with Pyblish, didn't know about this one.

BigRoy:
> I'm thinking it may be more than just the color of the icon; it'd be the color associated to this Loader. Could be used for more than its icon.

In that case, interesting! Not sure where this could be visually represented, but I'm liking this idea. :)

mottosso:
Cool. :) And I wouldn't worry about confusing it for the text-color. It'd be part of it's identity, rather than a direct connection to anything we've coded. Like it's label. The user chooses an appropriate label, we decide where a label is appropriate to use.

BigRoy:
Refactored `icon_color` to `color`. Example code:

```python
class ModelLoader(LogDebugLoader):
    """Load a model"""
    families = ["colorbleed.model"]
    representations = ["ma"]
    label = "Reference model"
    order = -10
    icon = "code-fork"
    color = "orange"


class GpuCacheLoader(LogDebugLoader):
    """Import a GPU cache"""
    families = ["colorbleed.model", "colorbleed.animation"]
    representations = ["abc"]
    label = "Import GPU cache"
    order = -1
    icon = "download"
    color = "white"
```

mottosso:
Orange is the new black!

mottosso:
Mergie mergie?

PR #216, opened 2017-06-27
Simplify Loader and the arguments for `host.load`

BigRoy:
- Remove "preset" argument in `host.load`
- Remove "post_process" argument in `host.load`
- Add "data" argument in `host.load`; for additional settings of a loader

Note that already implemented Loaders in configs need to now also take in the "data" as argument into `Loader.process`.

### DIFF
--- a/avalon/maya/pipeline.py
+++ b/avalon/maya/pipeline.py
@@ -274,6 +274,11 @@ def load(Loader,
         "representation": representation,
     }
 
+    # Ensure data is a dictionary when no explicit data provided
+    if data is None:
+        data = dict()
+    assert isinstance(data, dict), "Data must be a dictionary"
+
     name = name or subset["name"]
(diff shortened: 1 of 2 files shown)


BigRoy:
Maybe the `data` argument is more descriptive when named `options`?

aardschok:
I would opt for the argument to be called `options` as `options` can do something with the `data` but `data` can't do anything with the `options` ( hope this is a clear argumentation ).

BigRoy:
> I would opt for the argument to be called options as options can do something with the data but data can't do anything with the options ( hope this is a clear argumentation ).

Not really; when refactoring it to `options` there is no more `data` in that local sense. :)

BigRoy:
@mottosso what do you think?

mottosso:
Looks good, let's do it.

PR #218, opened 2017-06-28
Repair backwards compatibility with containers from before the project rename

mottosso:


mottosso:
I've been meaning to remove the notion of "pyblish" as well, since that was implemented before the pipeline was a project one its own. It can wait, but something to keep in mind. This is an "avalon.container".

BigRoy:
> I've been meaning to remove the notion of "pyblish" as well, since that was implemented before the pipeline was a project one its own. It can wait, but something to keep in mind. This is an "avalon.container".

Sounds good to me, can we push that through now.

With this backwards compatibility I'm thinking mostly, when do you ever drop it. :)

mottosso:
> I'm thinking mostly, when do you ever drop it. :)

I'd imagine it's the same for backwards compatibility in any project. In Pyblish, there is backwards compatibility from 1.0 through to 1.4.5, that's 3 years of development.

I'm doing semantic versioning in here as well, so when we do a major release, it's safe to remove compatibility so long as we document and warn about it.

PR #222, opened 2017-06-29
Add backwards compatibility to maya.ls

mottosso:
This should take care of listing assets loaded with an older version of the pipeline.

### DIFF
--- a/avalon/maya/pipeline.py
+++ b/avalon/maya/pipeline.py
@@ -229,8 +229,11 @@ def ls():
                 {"_id": io.ObjectId(data["representation"])}
             )
         except io.InvalidId:
-            api.logger.warning("Skipping %s, invalid id." % container)
-            continue
+            try:
+                document = _ls_4_0(data)
+            except io.InvalidId:
+                api.logger.warning("Skipping %s, invalid id." % container)
+                continue
 
         data = dict(
             schema="avalon-core:container-1.0",
@@ -242,6 +245,22 @@ def ls():
         )
 
         yield data
+
+
+def _ls_4_0(data):
+    """Representation ID is missing pre-5.0"""
+    document = io.locate([
+        api.session["project"],
+        data["asset"],
+        data["subset"],
+        data["version"],
+        data["representation"].strip(".")
+    ])
+
+    if not document:
+        raise io.InvalidId
+
+    return document
 
 
 def load(Loader,


mottosso:
I'm having trouble with this. 

The problem is the containers in the scene has changed their contents.

From..

```json
{
  "id": "",
  "name": "",
  "author": "",
  "loader": "",
  "families": "",
  "time": "",
  "asset": "",
  "subset": "",
  "representation": "",
  "version": 0,
  "silo": "",
  "path": "",
  "source": ""
}
```

To..

```json
{
  "id": "",
  "name": "",
  "namespace": "",
  "loader": "",
  "project": "",
  "asset": "",
  "subset": "",
  "version": 0,
  "representation": ObjectId()
}
```

The critical differences is that `representation` is the ID to the object in the database, it's used to find relevant versions to the original.

This is one of those cases where implementing backwards compatibility is possible, but introduces a lot of code..

BigRoy:
Actually, with "representation" being the ObjectId wouldn't asset and subset actually be irrelevant information since they are based on the representation.

mottosso:
> Actually, with "representation" being the ObjectId wouldn't asset and subset actually be irrelevant information since they are based on the representation.

They might. The original reason for adding so many members, was to not have to query the database so much. An optimisation.

I might still opt for this route, as currently the one thing that requires a unique database query is checking versions, and that could be checked more explicitly, e.g. a global button "check for updates" or an individual right-click menu item.

It'd enable us to make a more performant application, and database, in one swoop.

For example, and we should talk more about this once things settle, the new loader is 64x slower than the original (0.05s versus 3.5s on a complex, real world production assets) due to the excessive queries.

BigRoy:
> For example, and we should talk more about this once things settle, the new loader is 64x slower than the original (0.05s versus 3.5s on a complex, real world production assets) due to the excessive queries.

Interesting. Note that the queries it does now are upon refresh on all the subsets at once, instead of the a single "subset", then a single "representation", etc. Plus to also know whether it is the latest version we'd need to know about the other versions. Did you test both now, or are you comparing to older times you got from your loader? (If so, that one will now likely be slower because of some backwards compatible code recently introduced)

(3.5 seconds sounds really slow by the way! With the new loader which is close to our loader used without database even in scenes with over 10.000 objects (many similar ones though) I think it was faster than that, which was using cquery.)

But definitely, we should be looking forward and get good speeds.

mottosso:
Ok, here's how I'm thinking we tackle this.

Containers in Maya have a fixed set of data, very similar to any other document in the database. For documents in the database, each is imbued with a `schema` value specifying what format the data is in.

In the case of containers in Maya, this doesn't exist. Which means we can't implement functionality to detect the format before handling it.

So considering we'll likely keep changing this, and should be able to change it, I'll consider this and the past versions as 1.0 and 2.0 respectively, and embed a `schema` value into the containers created with this latest update. Since no artist is yet using this in production, it is safe to make this change.

Then, when there is a `schema` attribute on new containers, and **no** schema attribute on the previous version, we have an explicit method of detecting the schema.

For following versions, we can increment the version of the schema, and thereby enable change of containers indefinitely, including scenes with containers at mixed schema versions.

The solution is complex, but so is the problem I think. Any ideas, let me know.

mottosso:
Then the question is whether we should transform the data coming from Maya before making use of it e.g. the Manager GUI, or whether we let the Manager be the one to handle multiple versions of schemas.

For example. If a scene contains two loaded assets of one version each. The resulting dictionaries fed into the Manager could be either the data as-is, or pre-transformed into a common format such that the GUI would never need to change.

This pre-transformation would then need an additional schema - an "operator" schema, if you will - , separate from the "data" schemas.

I would say it's best to let GUIs handle this themselves. Missing or unsupported functionality could then be properly visualised in the GUI. They could potentially provide their own "migrate" functionality where possible.

mottosso:
To give you a picture of what the above was about.

```bash

# Option 1
  __________          _____ 
 |          |        |     |
 | Data v.1 |--------|     |
 |__________|        |     |
                     | GUI |
  __________         |     |
 |          |        |     |
 | Data v.2 |--------|     |
 |__________|        |_____|


# Option 2
  __________          __________      _____ 
 |          |        |          |    |     |
 | Data v.1 |--------|          |    |     |
 |__________|        |          |    |     |
                     | Data v.3 |----| GUI |
  __________         |          |    |     |
 |          |        |          |    |     |
 | Data v.2 |--------|          |    |     |
 |__________|        |__________|    |_____|
```

BigRoy:
> I would say it's best to let GUIs handle this themselves. Missing or unsupported functionality could then be properly visualised in the GUI. They could potentially provide their own "migrate" functionality where possible.

Having this visually there might also introduce some more complexity. I'm mostly afraid that even though the manager might know how to handle it the API itself doesn't.

The important bit I think is that the API returns as much consistent data as we can, as such the API should already return it... right?

mottosso:
> The important bit I think is that the API returns as much consistent data as we can, as such the API should already return it... right?

It would, in that data is handed to you in dictionary form, but their internal layout would depend on which version of the schema they adhere to.

This is unavoidable, and one of the things we can hold against an SQL approach where migration would be difficult or impossible.

Consider what would happen if we didn't allow for containers with data of two different schemas.

**container1**

```json
{
  "name": "Joe"
}
```

**container2**

```json
{
  "nickname": "Joe"
}
```

If we made this change, we would need to update our tools, like Manager, to account for it. **But**, we would have no control over what the data would look like in containers produced **before** this change.

The only option is to migrate this data. But we can't, because the data is embedded into the workfiles of artists. Migrating it would involve opening each scene and remove/add attributes from the containers themselves. Only at that point would our tools, like Manager, be able to work with them.

With this document-database oriented approach with embedded schemas, we can now facilitate data in different schemas, without requiring migration.

Neither approach is not without complexity. But it'll have to be one or the other, because change is inevitable.

BigRoy:
Still. I would say `api.ls()` would convert the data to a given format.

As such `api.ls()` would return it both as `name` if that's the newer format. But I'm not sure whether that would break other things. But otherwise it'll be hard to ever write tools in the future, because you'll need to know about the older data structures, otherwise the api would manage that?

BigRoy:
Looking at this I can see that it might keep things "closer" to the end functionality for backwards compatibility, but this would introduce the same backward compatibility functionality needs to be implemented per tool. (And I can see that number grow rapidly as time goes on?)

mottosso:
> because you'll need to know about the older data structures

Yes, we will. And we do, this is what the schemas are for. They explicitly define what is available in the data we make use of.

> but this would introduce the same backward compatibility functionality needs to be implemented per tool. (And I can see that number grow rapidly as time goes on?)

This is what I meant that either approach will involve complexity.

Let's consider the alternative, rather than have each tool support multiple schemas, we pre-process all data to conform to a single schema. That is, **Option 2** above.

Let's use the layout we use currently in the new Manager, which looks like this.

```json
{
  "representation": "5fj29ytbjdej3hj"
}
```

Where the representation is a MongoDB ID that we use from within the Manager to figure out (1) asset, (2) subset and (3) version.

**1 month later**

Let's consider a hypothetical example, where we discovered that going this path resulted in a performance penalty for shots with many containers and would like to improve performance by embedding (1) asset, (2) subset and (3) into the container itself.

```json
{
  "asset": "Bruce",
  "subset": "modelDefault",
  "version": 5,
  "representation": "ma"
```

Now there is no need to interact with the database as all of the information would reside in the local Maya scene. Great.

However, because `maya.ls()` has been designed to always return the same results, regardless of the data it is given, it would now have to convert these attributes into a MongoDB ObjectId before passing it onto the Manager, who will be expecting it.

```python
converted = {
  "representation": io.locate(api.session["project"], asset, subset, version, representation)
}
```

Now the Manager can continue working without modification, but look. Now we've made things even slower by introducing an additional query to the database for the transformation of the data meant to speed things up.

**6 months later**

We discover that MongoDB doesn't scale and decide to go for PostgreSQL. Suddently the ObjectId we use for the `representation` in the data returned by `maya.ls()` isn't relevant anymore so we must change it.

But look, we can't, because the Manager has been designed to use this ID to send queries to MongoDB. What we'll have to do, is refactor the results returned by `maya.ls()` so that it can work with the new database. And in doing so, remove support for projects using MongoDB.

The end result is that we can't transition between the two databases, but must perform a "clean switch", such as from one project to another. In reality this won't happen and would likely become a problem.

mottosso:
This is consuming more energy than it's worth.

Let's stick with this breaking change, and have artists at Mindbender run an optional migration script if needed. From now on I expect every change to the pipeline to be fully backwards compatible. This means building things in such a way that multiple versions of each schema can be maintained.

For example.

```python
# Don't
for container in maya.ls():
  print(container["some_member"]
```

```python
# Do
for container in maya.ls():
  # "schema" is a required member for all container versions
  if container["schema"] == "avalon-core:container-1.0":
    handle_1_0(container)
  elif container["schema"] == "avalon-core:container-1.0":
    handle_2_0(container)
  else:
    raise UnsupportedSchema("%s is not supported" % container["schema"])
```

As discussed above, there's complexity involved either way, so let's give this route a go and see how things turn out. It'll be a lot easier to revert a system with backwards compatibility to one without, than the other way around.

PR #226, opened 2017-06-29
Fix `cbsceneinventory` not working correctly with "filter outdated" enabled

BigRoy:
This fixes the "update to latest" not working when the 'outdated filter' was enabled.

### DIFF
--- a/avalon/tools/cbsceneinventory/proxy.py
+++ b/avalon/tools/cbsceneinventory/proxy.py
@@ -63,6 +63,13 @@ class FilterProxyModel(RecursiveSortFilterProxyModel):
         column = self.filterKeyColumn()
         index = model.index(source_row, column, source_parent)
 
+        # Filters below are per version group (top-level entries) so if
+        # it's a child index consider it non-filtered, because the header
+        # group already would be.
+        has_parent = index.parent().isValid()
+        if has_parent:
+            return True
+
         # Filter to those that have the different version numbers
         node = index.internalPointer()
         version = node.get("version", None)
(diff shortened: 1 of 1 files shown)


avalon/tools/cbsceneinventory/proxy.py
```diff
@@ -63,6 +63,13 @@ def filterAcceptsRow(self, source_row, source_parent):
         column = self.filterKeyColumn()
         index = model.index(source_row, column, source_parent)
 
+        # Filters below are per version group (top-level entries) so if
```
mottosso: Hm, I can't make sense out what this comment means. Do you think you can reformulate it?

BigRoy: https://github.com/getavalon/core/pull/226/commits/b62bc134aaefba0629c0f71ff912279ba6422a8e tries to reformulate it.

mottosso:
That makes more sense to me. What do you think?

BigRoy:
We could also do this:
```python
        ...

        # Filter the "representation" group by whether they are outdated
        is_representation_group = not index.parent().isValid()
        if is_representation_group:

            # Filter to those that have the different version numbers
            node = index.internalPointer()
            version = node.get("version", None)
            highest = node.get("highest_version", None)

            # Always allow indices that have no version data at all
            if version is None and highest is None:
                return True

            # If either a version or highest is present but not the other
            # consider the item invalid
            if version is None or highest is None:
                return False

            return version != highest

        # If an index hasn't been filtered, consider it valid
        return True
```

But I think it's good as is.

PR #230, opened 2017-07-06
Cosmetics

mottosso:
Moving build.py away from core, and tidying up a bit in io.py


mottosso:
`AVALON_EARLY_ADOPTER` enables showing/hiding of features not necessarily stable, or as in this case providing fallback functionality for untested features.

mottosso:
Let me know your thoughts on `AVALON_EARLY_ADOPTER` @BigRoy

PR #231, opened 2017-07-07
Initial draft of event mechanism

mottosso:
At the moment, Avalon adds `mbId` to specific nodes relevant to how Mindbender does their shading.

I've separated this out into Mindbender's configuration and implemented an event framework for hooking things back into the core.

```python
from avalon import api

def on_new():
  print("A new scene was created!")

api.on("new", on_new)
api.emit("new")
# A new scene was created!
```

By default, these trigger on Maya `init`, `new` and `save`

- [Example use](https://github.com/mindbender-studio/config/blob/38ab00bdf662ffdeed24c3f51e147e28c83b6354/polly/maya/__init__.py#L27)

Let me know your thoughts @aardschok.

### DIFF
--- a/avalon/__init__.py
+++ b/avalon/__init__.py
@@ -12,3 +12,4 @@ _registered_plugin_paths = dict()
 _registered_root = {"_": ""}
 _registered_host = {"_": None}
 _registered_config = {"_": None}
+_registered_event_handlers = dict()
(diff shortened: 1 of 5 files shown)


mottosso:
This is still up for discussion, I merged for the docs to compile.

Issue #54, opened 2017-02-21
Normalise references to external paths


mottosso:
### Goal

Enable a flexible project root.

Rather than references, textures and caches being stored as absolute paths, utilise Maya's path resolver to remove the need for absolute paths.

**Before**

```bash
m:\projects\projectA\assets\Bruce\rig.ma
```

**After**

```bash
$PROJECTDIR\assets\Bruce\rig.ma
```

This way, a project can reside anywhere on the artists machine, on any drive, in any subfolder.

### Implementation

On scene save, a nodes of a specific set of types are scanned for paths. Anything with the `PROJECTDIR` variable in it is automatically replaced. The types are references, files and alembic caches. Anything with support for automatic resolving.

jadecilcleton:
Hey Marcus, :)

Im facing this problem a lot, since usually I work remotely. 
I know that texture file nodes easily can store relative paths or variable replacements like you mentioned above . But for reference nodes this seems to have some issues, at least based on what I already tried. 
Do you already have some working solution you found to fix this in other situations outside mindbender pipe? 
If so, could you talk about it here?

Thank u!

mottosso:
Hey Bruno,

> But for reference nodes this seems to have some issues

What issues exactly, what have you tried?

The example project uses these at the moment, and seem to be working fine.

- See [here](https://github.com/mindbender-studio/example/blob/master/projects/hulk/assets/Bruce/work/rigging/marcus/maya/scenes/rig_v001.ma#L5)

Simplest way to try the example project for yourself is to [install it](https://mindbender-studio.github.io/#install), you can do that at home in any directory. Alternatively, [download it](https://github.com/mindbender-studio/example/archive/master.zip) and set it with you existing `mb.bat`.

jadecilcleton:
Nice, thank u for show me the exactly place you intent to replaced the reference path.
If you tested and say it works its fine, but thank u anayway to make it available for anyone to test.

This example uses an asset **modelDefault** inside a **rigDefault** file, where the project is set to the **ROOT** folder in the project ("$projectdir/assets/Bruce/publish/modelDefault/v001/modelDefault.ma"). So both files has identical **$projectdir** variable.

Now, in our current pipeline each shot has the project set inside the shot folder. For instance a **sh_010** in a **pXXX_project** has the project set to:

M:/f01_projects/pXXX_project/film/sh_010/work/animation/Mindbender/maya

[...]

PR #238, opened 2017-07-13
normalise references to external paths

mottosso:
#54 normalise references to external paths

By having your Loader inherit from `avalon.maya.Loader`, the `self.fname` now replaces the root portion of the path with `$AVALON_PROJECTS`.

**Before**

```bash
c:/projects/batman/assets/Bruce/publish/modelDefault/v003/modelDefault.ma
```

**After**

```bash
$AVALON_PROJECTS/batman/assets/Bruce/publish/modelDefault/v003/modelDefault.ma
```

Which means the project directory can reside on a different root, without affecting the internal path used by Maya. The same should/may apply to Nuke and other DCCs.

### DIFF
--- a/avalon/maya/__init__.py
+++ b/avalon/maya/__init__.py
@@ -9,6 +9,7 @@ from .pipeline import (
     uninstall,
 
     Creator,
+    Loader,
 
     ls,
     load,
@@ -39,6 +40,7 @@ __all__ = [
     "uninstall",
 
     "Creator",
+    "Loader",
 
     "ls",
     "load",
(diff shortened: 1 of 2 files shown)


mottosso:
#238 approaches this problem from the ingestion point of view, of loading references using the `$AVALON_PROJECTS` environment variable.

This means assets loaded by hand won't be getting a relative path. A perhaps acceptable sacrifice, especially considering everything going into an application *should* go through a Loader. But in practice, there will be things loaded by hand, so an additional safe-guard could be put in place which converts absolute to relative paths on, for example, scene save.

PR #247, opened 2017-07-26
Lock and before/after

mottosso:
This implements a new feature and two aliases to `api.on`

```python
from avalon import api
api.before("save", before_save)
api.after("save", after_save)
```

Which calls the given function before a given signal. Here's how before is implemented in Maya.

```python
OpenMaya.MSceneMessage.addCheckCallback(
  OpenMaya.MSceneMessage.kBeforeSaveCheck, your_callback
)
```

And an example.

```python

def before_save(return_code, _):
    """Prevent accidental overwrite of locked scene"""
    OpenMaya.MScriptUtil.setBool(return_code, not maya.is_locked())

```

Which brings me to the new feature.

**Lock**

Prevent further modifications to a scene.

```python
from avalon import maya

maya.lock()
cmds.file(save=True)
# Blocked!

maya.unlock()
cmds.file(save=True)
# Ok

# Save scene with lock in place
with maya.lock_ignored():
  maya.lock()
  cmds.file(save=True)

assert maya.is_locked() == False
```

### DIFF
--- a/avalon/api.py
+++ b/avalon/api.py
@@ -24,6 +24,8 @@ from .pipeline import (
     Session as session,
 
     on,
+    after,
+    before,
     emit,
 
     publish,
@@ -62,6 +64,8 @@ __all__ = [
     "session",
 
     "on",
+    "after",
+    "before",
     "emit",
 
     "publish",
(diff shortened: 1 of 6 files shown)


mottosso:
This works well!

PR #251, opened 2017-07-26
Resolved issue with sorting, implemented proxy between model and view

aardschok:
Due to the proxy missing from the subset widget the items added could be not be sorted correctly.

* Added proxy module
* Implemented proxy
* Minor cosmetics

PR #252, opened 2017-07-26
Added search bar to Asset Loader

aardschok:
* Implemented QLineEdit for searches.
* Text changed trigger `proxy.setFilterRegExp`
* Proxy set to `CaseInsensitive`


mottosso:
This one actually contains #251, should we close one of them?

BigRoy:
Close the other, because I believe this also has fixes for problems that the other one introduced. :)

avalon/tools/cbloader/app.py
```diff
@@ -171,7 +171,8 @@ def _versionschanged(self):
         if active:
             rows = selection.selectedRows(column=active.column())
             if active in rows:
-                node = active.data(subsets.model.NodeRole)
+                node_data = subsets.model.NodeRole
+                node = active.data(node_data)
```
BigRoy: Personally I preferred the old style here: `node = active.data(subsets.model.NodeRole)`

avalon/tools/cbloader/delegates.py
```diff
@@ -106,7 +106,9 @@ def displayText(self, value, locale):
 
 class VersionDelegate(QtWidgets.QStyledItemDelegate):
     """A delegate that display version integer formatted as version string.
-    
+
+    Note:
+        It is required to set source_model attribute.
```
BigRoy: This added docstring note isn't actually true so can be removed (that was only from some debugging in development)

avalon/tools/cbloader/widgets.py
```diff
@@ -10,6 +10,7 @@
 from .model import SubsetsModel
 from .delegates import PrettyTimeDelegate, VersionDelegate
 from . import lib
+from .proxy import FilterProxyModel
```
BigRoy: This import isn't used. This also means the added `proxy.py` file is redundant in `cbloader`.

BigRoy:
@aardschok could you check the above comments made here and after changing test whether the functionality is all there? :) Then we should be safe to merge this.

aardschok:
@BigRoy Everything works with said feedback implemented

mottosso:
Would it be possible to make a small gif showcasing this new functionality? Ideally I'd like for all visual changes to come with a gif or still image.

aardschok:
## Results:
asset_loader_sorted_by_version
asset_loader_filter_by_look
asset_loader_sorted_by_family

mottosso:
Great! That'll make our future selves and any followers happy. :)

I try and do that to encourage comments without actually having to (1) clone, (2) checkout your PR branch, (3) set it up locally and (4) find a way to run it. A simple picture takes all that away.

BigRoy:
Great work @aardschok!

PR #254, opened 2017-07-27
Manager - Remove container if exists when removing item

### LINKED ISSUE #253: Manager fails to remove reference and linked container
issue #118 not working

Issue remains where container is left in the scene ad not properly removed when using `avalon.maya.pipeline.remove(container)`.

To reproduce the issue:
1. Reference in a few subsets through the Asset Loader
2. Remove a referenced subset through the Manager
3. Refresh the Manager to see the "removed" subset remains with the container name as objectName

image

aardschok:
Fix #253 

Added extra check ( `cmds.objExists(node)` ) to counter error where the container is already removed.
When removing a referenced Alembic the removal includes the container when removing referenced Maya Ascii it might linger after the file has been removed from references.

### DIFF
--- a/avalon/maya/pipeline.py
+++ b/avalon/maya/pipeline.py
@@ -522,8 +522,11 @@ def remove(container):
     fname = cmds.referenceQuery(reference_node, filename=True)
     cmds.file(fname, removeReference=True)
 
-    if cmds.objExists(node):
+    try:
         cmds.delete(node)
+    except ValueError:
+        # Already implicitly deleted by Maya upon removing reference
+        pass
 
     try:
         # If container is not automatically cleaned up by May (issue #118)
--- a/avalon/maya/pipeline.py
+++ b/avalon/maya/pipeline.py
@@ -525,6 +525,7 @@ def remove(container):
     try:
         # If container is not automatically cleaned up by May (issue #118)
         cmds.namespace(removeNamespace=namespace, deleteNamespaceContent=True)
+        cmds.delete(node)
     except RuntimeError:
         pass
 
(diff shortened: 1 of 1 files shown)


avalon/maya/pipeline.py
```diff
@@ -522,6 +522,9 @@ def remove(container):
     fname = cmds.referenceQuery(reference_node, filename=True)
     cmds.file(fname, removeReference=True)
 
+    if cmds.objExists(node):
```
mottosso: No need to test, the node is guaranteed to exist.

aardschok: Referenced Alembic animation, tried `cmds.detele(node)`.
Result when removing through Manager:

```python
# Traceback (most recent call last):
#   File "C:\Users\User\Documents\development\avalon\core\avalon\tools\cbsceneinventory\app.py", line 96, in <lambda>
#     lambda: self.show_remove_warning_dialog(items))
#   File "C:\Users\User\Documents\development\avalon\core\avalon\tools\cbsceneinventory\app.py", line 250, in show_remove_warning_dialog
#     host.remove(item)
#   File "C:\Users\User\Documents\development\avalon\core\avalon\maya\pipeline.py", line 526, in remove
#     
# ValueError: No object matches name: shot1_01__test_001__CON
```
Existence of node cannot be guaranteed.

mottosso: Ah, it looks like Maya is being smart and automatically deletes it for us under some circumstances.

Ok, in that case the only critique would be to delete first, ask later, in preparation for the day Maya becomes multi-threaded and could create data inbetween two subsequent commands, like a filesystem can.

```python
try:
  cmds.remove(..)
except TheError:
  pass
```

PR #258, opened 2017-07-28
Fix #257 - removed self._collection from entirely from  module

### LINKED ISSUE #257: Lingering deprecated variable self._collection in IO module
## Issue
When trying to publish the `io.insert_many` function seems to use `self._collection` which has been removed from the module all together. 

## Solution
Replace `self._collection` with `self._database[Session["AVALON_PROJECT"]]`

aardschok:
Resolved #257

### DIFF
--- a/avalon/io.py
+++ b/avalon/io.py
@@ -161,7 +161,6 @@ def uninstall():
 
     self._mongo_client = None
     self._database = None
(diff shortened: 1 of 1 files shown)


mottosso:
Nice work, awaiting Travis.

Issue #72, opened 2017-03-04
Selective Sync


mottosso:
### Goal

Browse and download assets from any location.

- Depends on #31 and #25

### Implementation

The loader displays local and remote assets, offering to "Make local". This would involve coupling SyncThing with the loader, along with somehow enabling one user to fetch available assets from another location.

The connection could be bi-directional, in that one freelancer may want to browse assets at Mindbender in Gothenburg, whereas another studio may be producing assets that Mindbender may be interested in. In this way, assets can be stored and shared from and to any location.

Records of available assets will initially be stored at Mindbender Gothenburg, later to potentially be delegated to a remote man-in-the-middle, such as Firebase.

mottosso:
With #31 and #25 in place, this can now move forward.

At the moment, the artist can browse and start work on a remote project via the Launcher. The hierarchy will be created in an identical fashion to others working at other locations on the same project.

Currently, attempting to load an asset not currently available locally, it simply fails.

![image](https://cloud.githubusercontent.com/assets/2152766/25986669/c528cd6e-36e8-11e7-9dca-2d9ad0221ab2.png)

What should happen, is a dialog or button with the option to download this and dependent assets.

### Implementation

There are two challenges involved here.

1. How does a **server** broadcast files?
2. How does a **client** receive them?

<br>

### Server

![image](https://cloud.githubusercontent.com/assets/2152766/25991032/ca09f0c2-36f9-11e7-865d-0a7357a73e25.png)

We need some way of a producer of content, like the Mindbender studio, to enable others to download files from it. The same goes for when another studio or freelancer produces content that the Mindbender studio needs to use.

Spontaneously, I'd imagine a simple file-server like Flask. Each party can advertise themselves via their public IP.

```json
{
  "type": "version",
  "locations": ["24.51.234.1:21011"],
}
```

Where the first location is where the file was first produced. Anyone downloading the file, assuming he is also capable of broadcasting files, then appends to this list.

1. Mindbender Gothenburg produces `/hulk/Bruce/modelDefault/1/ma`
2. Mindbender Brazil downloads this file
3. The file is now available from two locations.

```json
{
  "type": "version",
  "locations": ["24.51.234.1:21011", "194.51.2.1:21011"],
}
```

For this to happen, there needs to be an actively running server where anyone is producing content. The spontaneous solution to this is to imbue the Launcher with this ability. So long as the Launcher is running, others are able to download files from this location.

This would work well for one-off freelancing gigs getting set-up quickly.

[...]

PR #266, opened 2017-08-01
Make available offline

mottosso:
It's [back](https://github.com/getavalon/core/pull/134)!

**Featuring**

- Customisable root, one artists working on `M:\` and another at `/my/server` works wonders
- Secure, full support for TLS encrypted transfers with multi-user authentication
- Bi-directional synchronisation, download on-demand and automatic upload on publish

**Download**

untitled project

**Upload**

untitled project

### DIFF
--- a/avalon/api.py
+++ b/avalon/api.py
@@ -30,8 +30,6 @@ from .pipeline import (
 
     publish,
 
-    loaders_by_representation,
-
     register_root,
     register_host,
     register_plugin_path,
@@ -69,8 +67,6 @@ __all__ = [
     "emit",
 
     "publish",
-
-    "loaders_by_representation",
 
     "register_host",
     "register_plugin_path",
--- a/avalon/inventory.py
+++ b/avalon/inventory.py
@@ -18,7 +18,6 @@ Usage:
 import os
 import sys
 import copy
-import json
 
 from avalon import schema, io
 from avalon.vendor import toml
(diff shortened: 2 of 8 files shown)


mottosso:
Regarding limitations, here's the status.

- [ ] No cancelling of downloads
- [X] No batch download of, say, an entire asset
- [X] Closing the Loader interrupts active downloads

Multi-selection of say the latest version of every subset of a shot works, there is a counter `(2)` in the download status notification at the bottom, and closing the GUI preserves the download state, unless a force-close (by holding SHIFT) is made in which case download is cancelled and temporary files are removed.

Cancelling of downloads is unsupported but possible by force-closing the GUI.

mottosso:
As for critical questions answered, A+!

- [X] Is #72 possible?
- [X] Does it make sense for the user?
- [X] Is the UI/UX intuitive?
- [X] What are the security implications to this approach?
- [X] What is the performance? Is it feasible?
- [X] Is the performance impact acceptable on the client? How about the server?

mottosso:
Bam, this is good to go.

About using the old Loader; that is temporary. I had the implementation for Offline Files already done through git history and due to time constraints merely re-instantiated that. The goal is definitely to get this into the new loader; I've made it as decoupled as possible here, with core parts delegated to `io.download` and other parts easily extractable, such as the `_download_manager` with module-level references.

PR #271, opened 2017-08-21
Python 3.6 support for clique

aardschok:
* replaced `basestring` with `six.string_types` for python 3.6 support


mottosso:
Win! Awaiting travis.

PR #281, opened 2017-09-28
Compat

aardschok:
Fixes Maya2018 issue with `MASH/scripts/googleapiclient` importing `http` instead of `pyblish0base/vendor/six.py`

### DIFF
--- a/avalon/maya/compat.py
+++ b/avalon/maya/compat.py
@@ -1,28 +1,29 @@
-"""Compatibility issues six.py / Maya2018
+"""Compatibility
 
-The Maya 2018 version tries to import the `http` module from
-Maya2018\plug-ins\MASH\scripts\googleapiclient\http.py in stead of the module
-from six.py. This import conflict causes a crash Avalon's publisher.
-This is due to Autodesk adding paths to the PYTHONPATH environment variable
-which contain modules instead of only packages.
-
-This will be used until Autodesk has fixed the issue
+This module is
 """
+import os
 
 
-def check_compatibility():
+def remove_googleapiclient():
     """Check if the compatibility must be maintained
 
-    The compatibility will only be maintained when the Maya version is 2018
+    The Maya 2018 version tries to import the `http` module from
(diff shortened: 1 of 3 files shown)


avalon/maya/compat.py
```diff
@@ -0,0 +1,29 @@
+"""Compatibility
+
+This module is
```
BigRoy: _This module is_ what? :D

Maybe you just stopped typing there for a second. Hehe.

avalon/maya/pipeline.py
```diff
@@ -35,6 +35,11 @@ def install(config):
     _register_callbacks()
     _set_project()
 
+    # Check if maya version is compatible else fix it, Maya2018 only
+    # Should be run regardless of batch mode
+    if hasattr(cmds, "about") and cmds.about(version=True) == "2018":
+        compat.install()
```
BigRoy: I wouldn't expect this to call `compat.install()`, but I'd expect it to call `compat.remove_googleapiclient` so the code here explains what it does. 

The `install()` in `compat` module for me seems redundant. Unless that `compat.install()` does the check internally in that file. Whether that's preferred maybe @mottosso might have some better arguments about that.

mottosso: Yeah, we chatted about that in private (for some reason?). I'd prefer `compat.install()` as I expect there to be more compatibility fixes long term and I'd rather not pollute the `pipeline.install` logic and readability with it.

BigRoy: Ah ok, so then the compatibility checks are moved there as well. (version checks, etc.) And `compat.install()` is just triggered here. That makes sense to me.

aardschok: I 've moved the version check to the `compat` module. Made more sense to me and less contamination of the `pipeline.install`.

mottosso:
Looks fantastic to me. Merge at will, and let me know if you need write-permissions!

Issue #278, opened 2017-09-20
Instance Creator - Locking the family name


aardschok:
_Note: This issue corresponds or partial corresponds with issues #194 / #246 / #269_

# Discussion
In our studio one of the artists rewrote the entire subset name to `"sign"` in two different tasks, modeling and lookdev. After publishing the artist could not find her published files anymore due to the lookdev files being stored in the same folder under `v002`. What happened was what @mkolar actually described.

```
Example result in publish folder:
prop_sign/
--sign/
----/v001/sign.ma (model)
----/v002/sign.ma (lookdev)
```

# Our solution
To counter this from happening we want to lock the `family` name in the `subset`. Everything being made from that point on will always lead with the family name, e.g.: "lookWet".

We also want to remove `name` attribute from the Creator class and focusing solely on the `family` attribute. This is possible by using the `string.rsplit` which is already in use in the current Instance Creator.

# Conflicts
* Backwards compatibility
    In the case of the current logic (1) vs new logic (2):
    1. `label = (plugin.name or plugin.family.lower().rsplit(".", 1)[-1] + "Default")`
    2. `label = "Default"` , this will result in `{family}Default`
    We might not be able to create instances which need to have very specific names such as `RenderGlobals`.

* publishing variants
    When locking the family we will no be able to implement the logic proposed in #269 to publish subsets/variants nicely even though this also result in non-backwards compatible paths or being forced migrate those already published.

Thoughts are welcome!

mottosso:
> In our studio one of the artists rewrote the entire subset name to "sign" in two different tasks,

This is not a design flaw, but misuse.

Here, "sign" isn't the name of a subset but the name of an asset. A subset is part of an asset, like a `model` or a `rig`. Here, a `sign` is part of a `sign`, which doesn't make any sense, hence the problem.

I'm not opposed changing things, but we shouldn't do it to accommodate for misuse. We should instead prevent misuse from happening, and then talk about design changes. For example by not enabling the user to type the name of a subset, but to choose it. If the choice is going to be the family, then why not remove the textfield altogether? Let it be just the family.

In this case, what does `sign` as a subset actually refer to? What was the *intent*?

The scenario is:

1. The user picks "lookdev" from the list of available families
2. The subset name field is automatically filled in, e.g. `lookdevDefault`
3. The user then changes the name to `sign`

[...]

PR #283, opened 2017-10-03
Fix mayapy launch erroring on main window during on_init(), implement launching cbloader in active context, only containerize if any nodes are actually loaded by a loader, and minor other things

BigRoy:
This is implements a couple of changes, I'll briefly describe them all:

1. In the instance creator the family name is now **always** prefixed before the subset name.

_Related to:_ #278 

instance_creator

The subset field does not require typing the family name in front anymore, but this is enforced in the creator. The bottom field will display the final name, and you can type your specific subset's name, e.g. "default".

> Note: The subset name's first letter is always uppercased, even if you type "default" in the field for a model family it will become `modelDefault`. (This is also visible in the bottom field)

Additionally it also supports a little dropdown (button to the right of the subset field) to select an already existing subset name within that family within that asset, so for `character` it would for example show "Default" if `modelDefault` was already published before.

2. The `avalon.tools.cbloader` can now launch within the current context.

Whenever working in the asset `character` it will now upon launch show that specific asset. For example when you're working in the lighting for shot 010 and you open the launcher it will open up inside that specific shot, so should most likely display the published subsets for that specific scene.

3. A fix is implemented for maya batch or mayapy failing on `on_init`.

In the initialization the code tried to find the maya main window, but in batch mode it didn't have that specific element and would error out. That specific case is now captured, implemented here: https://github.com/Colorbleed/core/blob/c7dbc9564c68d9b78cbe176851c7c8e8b7746f94/avalon/maya/pipeline.py#L667

4. `discover()` in `pipeline.py` will now ignore python files starting with underscore.

To quickly disable a loader or other plug-in you can prefix the file with an underscore and it will be ignored.

> Note: We also pushed the `os.path.isfile()` check until after the extension check just because that would reduce the amount of disk access (since the extension check wouldn't need disk access yet, whereas `os.path.isfile` would.) That was purely an optimization choice aside of it just making more sense to me too.

5. After loading in host Maya the `containerize` is only triggered if the loader stored any nodes within itself, so an empty container would never be created.

> Note: as far as I know this should be backwards compatible in the way that I believe no loaders were ever intended to create empty containers in the scene. If instead they were, this might break backwards compatibility.

### DIFF
--- a/avalon/api.py
+++ b/avalon/api.py
@@ -30,8 +30,6 @@ from .pipeline import (
 
     publish,
 
-    loaders_by_representation,
-
     register_root,
     register_host,
     register_plugin_path,
@@ -69,8 +67,6 @@ __all__ = [
     "emit",
 
     "publish",
-
-    "loaders_by_representation",
 
     "register_host",
     "register_plugin_path",
--- a/avalon/inventory.py
+++ b/avalon/inventory.py
@@ -18,7 +18,6 @@ Usage:
 import os
 import sys
 import copy
-import json
 
 from avalon import schema, io
 from avalon.vendor import toml
@@ -308,49 +307,6 @@ def _save_config_1_0(project_name, data):
     schema.validate(document)
 
     io.save(document)
-
-
-def _parse_bat(root):
-    """Return environment variables from .bat files
-
-    Handles:
-        Integers:
-            $ set AVALON_VARIABLE=1000
-        Strings:
-            $ set AVALON_VARIABLE="String"
-        Plain
-            $ set AVALON_VARIABLE=plain
-
-
-    """
-
-    name = os.path.basename(root)
-    dirname = os.path.dirname(root)
-    data = dict()
-
-    try:
-        with open(os.path.join(dirname, name + ".bat")) as f:
-            for line in f:
-                if not line.startswith("set AVALON_"):
-                    continue
-
-                key, value = line.rstrip().split("=")
-                key = key[len("set AVALON_"):].lower()
-
-                # Automatically convert to proper datatype
-                try:
-                    value = json.loads(value)
-                    value = int(value)
-                except ValueError:
-                    # Value wasn't an integer
-                    pass
-
-                data[key] = value
-
-    except IOError:
-        pass
-
-    return data
 
 
 def _report(added, updated):
--- a/avalon/io.py
+++ b/avalon/io.py
@@ -3,10 +3,15 @@
 import os
 import sys
 import time
+import errno
+import shutil
 import logging
+import tempfile
 import functools
+import contextlib
 
 from . import schema, Session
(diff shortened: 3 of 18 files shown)


mottosso:
Looks good!

image

This needs work. It isn't clear what the user is asked to type in. Is `default` automatically titlecased to `modelDefault`? Is `default` the subset itself? If so, what is it?

How about a horizontal layout?

```
 _______   _________   ___
|       | |         | |   |
| model | | Default | | v |
|_______| |_________| |___|
```

Where the `model` textbox is disabled, indicating that it will bear the name of the selected family, and `Default` is the editable final suffix to that name, defaulting to `Default` or whatever makes sense for the family.

BigRoy:
I noticed just now that the `load` method in host has a lot of different return types, it currently returns a `list` on `OSError`, `None` when no nodes were loaded and a `str` when containerized with `containerise`. 

This might be somewhat confusing, but was already a problem before (where it would return either an empty list or a string).

---

> This needs work. It isn't clear what the user is asked to type in. Is default automatically titlecased to modelDefault? Is default the subset itself? If so, what is it?

Yes, this is true. The bottom field is "non-editable" there. We could style it differently so it's more clear. We decided to keep it enabled and set it to read only so that if you wanted to copy paste the text there you could.

The text at the bottom is the full subset name as it will end up after publishing.

> How about a horizontal layout?

That could work too, but it should still be clear the left field is **not** editable. I believe the problem then is still the same, it's requires a clearer styling for that.

This would also require when typing in the right field that it automatically changes the first letter to an uppercase as the user types. For me both sounds fine, and I think the current situation, though with the bottom field clarified, could work at least as well if not better. Feel free to bring up some more arguments. :)

mottosso:
> For me both sounds fine, and I think the current situation, though with the bottom field clarified, could work at least as well if not better

Ah, ok. Yes in that case I think simply disabling the bottom one (which would cause it to draw with a greyish low-contrast color) should suffice.

Nice work!

mottosso:
Or even make it just a label, without the surrounding box. To really drive the point home that it is non-editable and just for show.

BigRoy:
I decided for now to just disable it, since it looked clearer then just swapping it with a QLabel. Now it looks like this:

afbeelding

Does this suffice for the time being?

mottosso:
> Does this suffice for the time being?

For sure. Thanks for doing that.

mottosso:
Code looks good too!

mottosso:
Merge at will!

BigRoy:
Actually, should we bump minor version?

mottosso:
Versions aren't really being used at the moment so don't worry about it.
What is being used is the commit id,  for embedding the submodule in setup
distributions, and we'll get those regardless.

BigRoy:
Cool, merging this now.

Issue #195, opened 2017-06-14
Deprecate host.load() and host.create()


mottosso:
### Goal

Delegate all host-specific responsibility into plug-ins.

### Motivation

At the moment, plug-ins determine company- and host-specific behaviour of the pipeline, *in addition to the host interface*.

```python
from mindbender import api, maya

# Run the host-specific installer
api.install(maya)

# It registers these members to be used by the include tools, the Loader and Creator.
maya.create("myInstance", family="rig")
maya.load("myAsset")
```

The `create()` function then determines, based on the family, which of the Creator plug-ins to run. Similarly, `load()` determines which Loader plug-ins to use based on the families of the published asset.

What these two functions do is almost entirely generic, with the exception of `load()` also performing a post-process of encapsulating all loaded nodes into a `container`, an object used to manage loaded assets in the host (in the case of Maya, an `objectSet`).

### Implementation

Delegate the one outlier from a completely generic system in `load()` to the superclass of the Loader plug-in.

We can then start looking at making menus and scripts part of the plug-in architecture as well, for a completely loosely coupled system with entirely customisable behavior and functionality.

BigRoy:
I'm working on this right now.

As I see it the `create`, `load`, `remove` and `update` methods become the responsibility of the Loader plug-ins. A loader would look like:

```python
# pseudocode (the arguments of the methods are just placeholders here)
class Loader(api.Loader):
    def load(self, representation):
        pass
    def update(self, version=-1):
        pass
    def remove(self, container):
        pass
```

As such a `host` itself would not implement specific loading logic but the "static" logic would move into the core api of Avalon, as such the api would expose:

```python
# Create an instance for publishing
api.create(name, asset, family, options, data)

# Load
api.load(Loader, representation, namespace, name, data)

# Remove a container
api.remove(container)

# Change version of a container
api.update(container, version=2)
```

---

Here's a quick copy and paste of how it could be refactored into the core `pipeline.py`:

```python
def create(name, asset, family, options=None, data=None):
    """Create a new instance

    Associate nodes with a subset and family. These nodes are later
    validated, according to their `family`, and integrated into the
    shared environment, relative their `subset`.

    Data relative each family, along with default data, are imprinted
    into the resulting objectSet. This data is later used by extractors
    and finally asset browsers to help identify the origin of the asset.

    Arguments:
        name (str): Name of subset
        asset (str): Name of asset
        family (str): Name of family
        options (dict, optional): Additional options from GUI
[...]

PR #291, opened 2017-10-20
#195

BigRoy:
Here's a draft api revamp for #195.

Basicallly the idea is that not the `host` decides how something is loaded and managed but the `Loader` implements all the logic. As such these loaders will now manage the `load`, `update`,  `remove` functionality.

At this point it's not 100% backwards compatible, here's the catch:

- You'll need to update all your Loaders to use the new `Loader.load` method.
    - For Maya there's `avalon.maya.pipeline.ReferenceLoader` - when you inherit your Loader from that specific loader you can keep your old `process` logic and it should work as expected.
- Any calls to `host.load`, `host.update` or `host.remove` should be respectively refactored to `api.load`, `api.update`, `api.remove`.
    - Also check for this in your current Loaders! If you trigger a `host.load` from your plug-in make sure to also refactor that to use the new `api.load()` method.

---

Other notes:

- The logic for "inheritance" check in `discover` has changed so that it allows checking multiple inheritance instead of only name of direct parent class.
- This could have some other commits that haven't been pulled into getavalon/core yet, but they should be minor.
- Note that the `update` method on the Loader plug-in doesn't have the same signature as the `api.update` method. (Because the `api.update` method already gets the representation for you based on version number alone so that logic doesn't have to be pushed into the Loader and Avalon's core handles the main logic there.)
- *Renaming the Loader can have confusing results* - currently the loader that's picked to remove or update a _container_ is based on its `loader` data, which is purely the Loader's class name. As such, when a Loader class is renamed (or removed!) it cannot be managed anymore. @mottosso Not sure how to deal with that otherwise?

---

> Note: *I've tested some tools but of course take a run with this yourself to see whether it's stable!*

### DIFF
--- a/avalon/maya/pipeline.py
+++ b/avalon/maya/pipeline.py
@@ -408,9 +408,6 @@ class ReferenceLoader(Loader):
              data=None):
 
         asset = context['asset']['name']
-        subset = context['subset']['name']
-
-        name = name or subset
         namespace = namespace or lib.unique_namespace(
             asset + "_",
             prefix="_" if asset[0].isdigit() else "",
--- a/avalon/maya/pipeline.py
+++ b/avalon/maya/pipeline.py
@@ -331,6 +331,30 @@ def containerise(name,
     return container
 
 
+def parse_container(container, validate=True):
+    """Return the container node's full container data.
+    
+    Args:
+        container (str): A container node name. 
+
+    Returns:
+        dict: The container schema data for this container node.
+        
+    """
+    data = lib.read(container)
+
+    # Backwards compatibility pre-schemas for containers
+    data["schema"] = data.get("schema", "avalon-core:container-1.0")
+
+    # Append transient data
+    data["objectName"] = container
+
+    if validate:
+        schema.validate(data)
+
+    return data
+
+
 def ls():
     """List containers from active Maya scene
 
@@ -346,16 +370,7 @@ def ls():
         containers += lib.lsattr("id", identifier)
 
     for container in sorted(containers):
-        data = lib.read(container)
-
-        # Backwards compatibility pre-schemas for containers
-        data["schema"] = data.get("schema", "avalon-core:container-1.0")
-
-        # Append transient data
-        data["objectName"] = container
-
-        schema.validate(data)
-
+        data = parse_container(container)
         yield data
 
 
--- a/avalon/maya/pipeline.py
+++ b/avalon/maya/pipeline.py
@@ -272,6 +272,262 @@ def lock_ignored():
         self._ignore_lock = False
 
 
+def load(Loader,
+         representation,
+         name=None,
+         namespace=None,
+         data=None):
+    """Load asset via database
+    Arguments:
+        Loader (api.Loader): The loader to process in host Maya.
+        representation (dict, io.ObjectId or str): Address to representation
+        name (str, optional): Use pre-defined name
+        namespace (str, optional): Use pre-defined namespace
+        data (dict, optional): Additional settings dictionary
+    """
+
+    assert representation is not None, "This is a bug"
+
+    if isinstance(representation, (six.string_types, io.ObjectId)):
+        representation = io.find_one({"_id": io.ObjectId(str(representation))})
+
+    version, subset, asset, project = io.parenthood(representation)
+
+    assert all([representation, version, subset, asset, project]), (
+        "This is a bug"
+    )
+
+    context = {
+        "project": project,
+        "asset": asset,
+        "subset": subset,
+        "version": version,
+        "representation": representation,
+    }
+
+    # Ensure data is a dictionary when no explicit data provided
+    if data is None:
+        data = dict()
+    assert isinstance(data, dict), "Data must be a dictionary"
+
+    name = name or subset["name"]
+    namespace = namespace or lib.unique_namespace(
+        asset["name"] + "_",
+        prefix="_" if asset["name"][0].isdigit() else "",
+        suffix="_",
+    )
+
+    # TODO(roy): add compatibility check, see `tools.cbloader.lib`
+
+    Loader.log.info(
+        "Running '%s' on '%s'" % (Loader.__name__, asset["name"])
+    )
+
+    try:
+        loader = Loader(context)
+
+        with lib.maintained_selection():
+            loader.process(name, namespace, context, data)
+
+    except OSError as e:
+        logger.info("WARNING: %s" % e)
+        return list()
+
+    # Only containerize if any nodes were loaded by the Loader
+    nodes = loader[:]
+    if not nodes:
+        return
+
+    return containerise(
+        name=name,
+        namespace=namespace,
+        nodes=loader[:],
+        context=context,
+        loader=Loader.__name__)
+
+
+class Creator(api.Creator):
+    def process(self):
+        nodes = list()
+
+        if (self.options or {}).get("useSelection"):
+            nodes = cmds.ls(selection=True)
+
+        instance = cmds.sets(nodes, name=self.name)
+        lib.imprint(instance, self.data)
+
+        return instance
+
+
+class Loader(api.Loader):
+    def __init__(self, context):
+        super(Loader, self).__init__(context)
+        self.fname = self.fname.replace(
+            api.registered_root(), "$AVALON_PROJECTS"
+        )
+
+
+def create(name, asset, family, options=None, data=None):
+    """Create a new instance
+    Associate nodes with a subset and family. These nodes are later
+    validated, according to their `family`, and integrated into the
+    shared environment, relative their `subset`.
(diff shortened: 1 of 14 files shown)


BigRoy:
Here's how we refactored our config to update with these changes: https://github.com/BigRoy/colorbleed-config/commit/fcd0498f7cad0cca744f624f2d3262c90b40f0c3

And here's how with these new API we're able to load gpuCaches and manage them through the manager: https://github.com/BigRoy/colorbleed-config/commit/88ff47311efaa38b7b99033e5b1fae033524c421

mottosso:
> The logic for "inheritance" check in discover has changed so that it allows checking multiple inheritance instead of only name of direct parent class.

A little warning here, the original reason for doing it this way was because when Avalon modules are reloaded the Loader classes lose their "identity" and can't be compared with the current class anymore, causing a `isinstance` to fail with a hard-to-debug error message.

Not everyone does this, and it's solely for debugging and development and won't affect the end user. But the alternative is either (1) having to reload Maya when changing modules or (2) finding a way of breaking the cyclic dependency currently causing this issue.

> Renaming the Loader can have confusing results

Yes, that's the nature of identifying objects by name. IMO the convenience trumps the complication in this case and I'd say the solution is to simply never change the name of loaders. We could embed an "id" or "label" though, rather than the class name so that devs can refactor at-will, without breaking things.

> At this point it's not 100% backwards compatible, here's the catch:

Sorry, this is a must. Even if it's temporarily 100% compatible, it's important for the transition.

mottosso:
Could you share an example of a loader using this new API?

mottosso:
> Could you share an example of a loader using this new API?

Sorry, I just saw you shared one in your subsequent post!

BigRoy:
> Not everyone does this, and it's solely for debugging and development and won't affect the end user. But the alternative is either (1) having to reload Maya when changing modules or (2) finding a way of breaking the cyclic dependency currently causing this issue.

Actually, I believe the code I implemented is still doing a check based on names but allows any of the parent classes to be named like that. :) So have a look and maybe test it... but it's working here.

> Sorry, this is a must. Even if it's temporarily 100% compatible, it's important for the transition.

In that case I believe we need to have a metaclass kind of backwards compatibility like in Pyblish where it would "magically" rebuild. 

The compatibility is currently only for the notes at the top! And, to clarify, it's compatible with older assets and scenes, you just need to update your local code too use these new classes/methods.

BigRoy:
**Note:** I'm still fixing some tiny quirks and typos in the code, but it should be relatively stable - though don't merge until I've the extra fixes in of course. The current state should allow you to play around.

BigRoy:
With this new API [here's another Loader preview, a draft for a "setdress" kind of package that can hold any other type of asset linked into its hierarchy](https://github.com/BigRoy/colorbleed-config/blob/dev/colorbleed/plugins/maya/load/load_setdress.py)... almost like a shot builder.

BigRoy:
I've cherry-picked the relevant commits from my `dev` branch and tested it here locally with my colorbleed config.

> Note: I'm still fixing some tiny quirks and typos in the code, but it should be relatively stable - though don't merge until I've the extra fixes in of course. The current state should allow you to play around.

These are now resolved.

> The compatibility is currently only for the notes at the top! And, to clarify, it's compatible with older assets and scenes, you just need to update your local code too use these new classes/methods.

This is still true. As such read the refactoring notes at the top.

- If you have custom code outside of avalon calling `host.load` you'll need to refactor to `api.load`, the same for `host.update` and `host.remove`. 
- If you have loaders with `process` implemented for Maya you can inherit from `avalon.maya.pipeline.ReferenceLoad` to preserve old behavior. To update to the new api you can inherit from `api.Loader` and implement your `load` logic completely.

---

@mottosso could you test this on your end?

mottosso:
Having a look now

mottosso:
Ok, it isn't backwards compatible. It has to be.

I've pushed a suggested method of retaining backwards compatibility. It's incomplete, but should highlight the essentials. Have a look and let me know what you think.

BigRoy:
> Ok, it isn't backwards compatible. It has to be.

Very unfortunate, the code could have been so much cleaner. Just to be sure. In what way does it need to be backwards compatible? If you'd update the config your scenes should be backwards compatible.

In a way it's only a code breaking compatibility, with an update there you can continue to work with older scenes. Refactoring my config took me 5 minutes. ;)

mottosso:
It needs to not interrupt the artist in any way, the old Loader plug-ins must work as-is, such that they can transition piece by piece.

Pyblish is a good example of this. You'll still find some plug-ins use the very old `set_data` interface. When encountered, you swap it for `.data`.

We need to make this a habit; upgrades should never break anything. It'll create a few of upgrading and I want upgrades to be trustworthy.

BigRoy:
> It needs to not interrupt the artist in any way, the old Loader plug-ins must work as-is, such that they can transition piece by piece.

Sure. The change doesn't interrupt the artists, just the tech will need to be updated along. I've had this from time to time too with Pyblish though but I understand where you're coming from. This will just be much harder to maintain.

I'll have a look at getting this backwards compatible.

mottosso:
> This will just be much harder to maintain.

It will be worth it. The alternative is being afraid of upgrades, and having to update large portions of code, in multiple repositories, all at once. *That* is hard.

Once we're in the compatibility mindset, we'll find ways of making it elegant. It takes some getting used to. For one, those functions I chucked back into maya.pipeline can go into a `maya.compat` and implicitly imported into maya.pipeline. See how pyblish does this its compat module. The result is static, legacy functions, residing away from relevant code so it doesn't bother anyone. Virtually zero cost to readability and maintenance.

mottosso:
We're in a good position here, as the loaders use different names for its inner methods. It'd be much harder to do this have we used the old `process` name!

BigRoy:
> It'd be much harder to do this have we used the old process name!

Sure, but if someone know ever implements a `process` method than it would automatically assume it's an older loader which it might not be. To me it's just an error in the new logic, but again, I can see where it's good to maintain older logic too - looking at it from my perspective it tends to just create more code that can break.

For example, in this case it'd mean the backwards compatibility is only maintained for Maya functionality. What if there was already code for the same in other hosts. It's not actually maintaining that too.

mottosso:
> For example, in this case it'd mean the backwards compatibility is only maintained for Maya functionality. 

The legacy load was only ever implemented for Maya, so it's quite safe to assume. These kind of things will have to be made on a case-by-case basis.

> Sure, but if someone know ever implements a process method than it would automatically assume it's an older loader which it might not be. 

Yes, a better approach would be to embed something in  your new Loader classes that distinguishes itself from the old ones. Maybe add a `version = "2.0"`? Up to you, I just wanted to show an option.

BigRoy:
@mottosso How can I now push a commit here without your commit in there? I seem to be unable to commit to my branch. Do I need to revert yours first?

mottosso:
Pull/merge mine first

BigRoy:
There we go, that should be it.

BigRoy:
I'm going into a meeting now; then I'll give this another test. Could you check this on your end for now @mottosso?

mottosso:
Will do

BigRoy:
I think this latest commit was another crucial thing I missed. :) Now it should be there.

mottosso:
Looking good Roy, I've made a few cosmetic tweaks (please install a linter), but other than that I think it's great.

avalon/pipeline.py
```diff
@@ -673,6 +675,7 @@ def create(name, asset, family, options=None, data=None):
         plugins.append(plugin)
 
     assert plugins, "No Creator plug-ins were run, this is a bug"
+    print("Here: %s" % instance)
```
BigRoy: @mottosso Did you intend to commit this?

avalon/maya/pipeline.py
```diff
@@ -13,8 +13,10 @@
 from ..vendor.Qt import QtCore, QtWidgets
 
 # Backwards compatibility
-from .compat import load, update, remove
-from ..pipeline import create
+load = compat.load
```
BigRoy: Love it. Thanks!

avalon/pipeline.py
mottosso: :S

mottosso: I'll remove it now.

avalon/maya/pipeline.py
mottosso: I would go further actually.

Have a look a [this](https://github.com/pyblish/pyblish-base/blob/master/pyblish/compat.py#L109).

It's got two advantages. 

1. It doesn't pollute readability.
2. *Not* importing this module means cleanly cutting backwards compatibility.

Because it's backwards compatibility, there is rarely if ever a need to modify the members in this module, so it's safe to make more difficult to read, grok and edit.

BigRoy:
Looks good. I'm merging this.

PR #299, opened 2017-11-09
Added `active` attribute to Creator.data

### LINKED ISSUE #298: Add "active" toggle on instance
Add `active` to instance data to allow artists to set the state before the collector picks up the instances.
This partially solves the issue of keeping you choice of what is to publish even when refreshing after running the validators for example.

## Argumentation for the name `active`
https://github.com/pyblish/pyblish-base/issues/307

aardschok:
Resolved #298

## Additional note
Each instance collector plugin in the configuration will need to be updated to translate the key to Pyblish logic. 

Example translation (also keeps things backwards compatible ;) )
```python
if "active" in data:
    data["publish"] = data["active"]
```

### DIFF
--- a/avalon/pipeline.py
+++ b/avalon/pipeline.py
@@ -213,7 +213,8 @@ class Creator(object):
             "id": "pyblish.avalon.instance",
             "family": self.family,
             "asset": asset,
-            "subset": name
+            "subset": name,
+            "active": True
         }, **(data or {}))
 
     def process(self):


mottosso:
Looks good to me! Do we need a tickbox in the Creator GUI for it as well?

aardschok:
I don't think a checkbox add that more to the tool. I assume people make an instance when they want to publish it. If the need arises they can turn it off on the specific instances. We will see how it turns out, if there is need for it we can always add it later.

BigRoy:
> I don't think a checkbox add that more to the tool. I assume people make an instance when they want to publish it. If the need arises they can turn it off on the specific instances. We will see how it turns out, if there is need for it we can always add it later.

I'd agree with this. No need to add it in the UI for now. 👍

mottosso:
Merge at will

PR #300, opened 2017-11-11
Use Session instead of `os.environ`

BigRoy:
I've started work on refactoring all tools and code to rely more on the Session and avoid falling back to `os.environ`.

This way we can start introducing "live context switching" by changing the active session.

- Refactored `os.environ` calls to `api.Session`
- Removed `AVALON_WORKDIR` in favor of `api.get_work_directory` so it's easier to remain in sync with the current session. This could be added back in but seems like a (possibly redundant?) optimization

I want to discuss what else will be needed to make it manageable to easily switch a context.


avalon/pipeline.py
```diff
@@ -720,6 +720,29 @@ def get_representation_context(representation):
...
+
+    The work directory is the parsed work template of the project configuration
+    against the current Session dict.
+
+    Returns:
+        str: Path to work directory
+
+    """
+    project = io.find_one({"type": "project"})
+    template = project['config']['template']['work']
```
mottosso: Try not to intermix function with calls to the database. This would be better suited to having the template passed as an argument.

avalon/io.py
```diff
@@ -132,9 +132,6 @@ def _from_environment():
             # Name of current app
             ("AVALON_APP", None),
 
-            # Path to working directory
-            ("AVALON_WORKDIR", None),
```
mottosso: What's the motivation here?

avalon/pipeline.py
```diff
@@ -720,6 +720,29 @@ def get_representation_context(representation):
     return context
 
 
+def get_work_directory():
```
mottosso: Must say I'm not a fan of these getter-style functions..

avalon/io.py
BigRoy: The reason is that this is computed based on the other keys, the workdir is based on the current context and isn't a variable on its own. The problem with storing it as one is that it needs to be ensured to remain in-sync with context changes. Whenever the context changes you'll need to update `AVALON_WORKDIR`.

I'm having similar issues with `AVALON_SILO` - it's dependent on the current `asset`.
In my mind they are both optimizations to reduce possible queries.

We could create an `update` method that would refresh all these "computed" values.

avalon/pipeline.py
BigRoy: Would be happy to discuss alternatives.

avalon/pipeline.py
BigRoy: Sure, so this would be more like a `format_template(template)` function?

avalon/io.py
mottosso: > We could create an update method that would refresh all these "computed" values.

Yeah. The context changing is under our control, so we can just update relevant values when that happens. For example, when one changes task or asset.

avalon/pipeline.py
mottosso: To me, the only reason to have a getter is to compute something when returning it. In our case, I think that should rarely if ever be the case, because implicit computations are difficult to debug and easy to misuse. It's too easy to use amongst calls that don't take long to compute, and then suddenly we've got a really slow code base with little idea of where time is being spent. A better alternative is making heavy calls explicit, to make it more difficult to use them, so that they stand out.

In this particular case, my suggestion would be the above; to update the computed version in Session directly.

avalon/io.py
BigRoy: Maybe I need to think about this some more. I'm expecting there will be quite some tools that could end up triggering context changes and I'd like to try and avoid ensuring everywhere that every session attribute remains in sync, because say this would not be "automatically refreshed somewhere" I'd have to at least do something like:

```python
# change my current session
api.Session['AVALON_ASSET'] = "villain"

# refresh silo
asset = io.find_one({"name": "villain", "type": "asset"})
api.Session['AVALON_SILO'] = asset['silo']

# refresh workdir
project = io.find_one({"type": "project"})
template = project['config']['template']['work']
api.format_template(template)
```

And this logic could be in multiple locations?
Maybe something like `api.refresh_session(session)` would make more sense?

Any ideas?

mottosso: How about a `update_asset(name)`?

mottosso: In general I don't think we can expect to change the values of Session directly. Could it be read-only?

BigRoy: > In general I don't think we can expect to change the values of Session directly. Could it be read-only?

It could be I'd say, yes. 

> How about a update_asset(name)?

This might make it slow if we're doing `update_asset` then `update_task` then `update_app`?
How about an `update_session` method that would allow to update multiple things at once? 

```python
#pseudocode
def update_session(session, project=None, asset=None, task=None, app=None):
    """Update session in-place to new context"""
    
    if asset:
        assert asset_exists, "Asset must exist"
        session['AVALON_ASSET'] = asset
        
    if any((project, asset)):
        # update silo

    if any((project, asset, task, app)):
        # update workdir

api.update_session(api.Session, asset="villain", task="modeling")
```

mottosso: How about `update_current_context(context)`? Session contains a lot more than asset and task, and that way it's only applicable to the current one (which at the moment is the only one there is).

Also signals should be `mixedCase`, e.g. `currentContextUpdated`

BigRoy: The problem with `context` to me is, what should it contain? Is it `asset` or `AVALON_ASSET`? And where can I find which attributes I _can_ update?

> and that way it's only applicable to the current one

Yes, but this might make the api less backwards compatible in the future when we *do* support multiple sessions to be passed around. I might be thinking to far ahead though. ;)

> Session contains a lot more than asset and task,

Correct - but it also means that if we update any of them that the updates are actually propagated I believe? Say you'd update the `AVALON_DB` attribute? What needs to happen? Will it do `io.uninstall()` and start reinstalling? Or would the user manually do that? which could be fine and more explicit? What else are you expecting to update?

mottosso: > Yes, but this might make the api less backwards compatible in the future when we do support multiple sessions to be passed around. I might be thinking to far ahead though. ;)

There will always be a current Session; it's the one Maya defaults to when booting up the Loader etc. It's forwards-compatible.

> Say you'd update the AVALON_DB attribute? What needs to happen? 

You wouldn't do it by modifying Session. You would use another `update_*` function. Session is a data container-only, read-only.

> The problem with context to me is, what should it contain? Is it asset or AVALON_ASSET? And where can I find which attributes I can update?

That's a good point..

How about..

```python
context = api.Context()

# Default values come from Session
context["asset"] = "OtherAsset"

api.update_current_context(context)
```

Not saying it's "the one". Still thinking of what's the best option.

BigRoy: @mottosso How about something like this (still pseudocode):

```python
def update_current_session(context):
    """Update active session in-place to new context

    Returns:
        dict: The key and values that have been changed in the session.

    """

    # Find the changed keys
    changed = {key: value for key, value in context.items() if 
               value != Session[key]}

    # Update project (or database) directly so any updates afterwards are
    # correct
    io_reinstall = False
    for key in ["AVALON_PROJECT",
                "AVALON_DB",
                "AVALON_TIMEOUT",
                "AVALON_MONGO"]:
        if key in changed:
            # These keys will require a `io.install()` call which uses the
            # current `os.environ` state. As such we'll need to update the
            # environment
            os.environ[key] = changed[key]
            io_reinstall = True

    # Install new project database
    if io_reinstall:
        io.uninstall()
        io.install()

    # Update silo when asset changes
    if "AVALON_ASSET" in changed:
        # Update silo for the asset
        asset_document = io.find_one({"name": changed["AVALON_ASSET"],
                                      "type": "asset"},
                                     projection={"silo": True})
        assert asset_document, "Asset does not exist"

        # Store our change to AVALON_SILO
        changed["AVALON_SILO"] = asset_document['silo']

    # Update work directory
    if any(key in changed for key in ["AVALON_PROJECT",
                                      "AVALON_ASSET",
                                      "AVALON_TASK",
                                      "AVALON_APP",
                                      "AVALON_USER"]):
        project = io.find_one({"type": "project"},
                              projection={"config.template.work": True})
        template = project['config']['template']['work']
        changed['AVALON_WORKDIR'] = format_template(template, Session)

    # Trigger updates on the Session
    Session.update(changed)

    return changed
```

mottosso: Wuw, that's much too complex! Also we shouldn't implicitly install/uninstall things. Those are heavy operations that do a lot of stuff and should only be done once. Think of it like installing software onto disk.

Also `context` doesn't include `AVALON_`, that's a prefix to distinguish variables in the environment. Internally, it's just `asset` etc.

I don't think we should go for `update_session`, I expect parts of it to be updated independently, especially tasks, then assets and lastly projects. I don't expect databases to change very often at all.

Maybe your expectations are different? Tell me what you've got in mind.

BigRoy: Expecting to:
- At any time in an application to switch task or asset; this is one of the most requested features in our studio.
- Switch user maybe?

I would avoid to:
- Change project (whilst in application) - changing project would just require you to open up the application for that project which makes more sense to me.
- Change database or anything that would require a reinstall or changes global settings, like sentry paths, database names, etc.

As such I'm only expecting to change things like asset, task, app, user to change. (And with that to update `AVALON_SILO` and `AVALON_WORKDIR` to change too.

mottosso: Also, some of the members of Session can't change independently, now that I think of it.

Changing `project` would invalidate `asset`, along with `silo` as you mentioned. So we should make sure updates happen with all dependencies at once. Passing `context` would guarantee that. `update_session` would too, but is too heavy handed.

And unless it wasn't clear, `context` in this.. context (pun!) is this fella:

- https://github.com/getavalon/core/blob/5c0ba8d613db056a549cc4709ab7a8393627491e/avalon/maya/compat.py#L75

Also found here:

- https://github.com/getavalon/core/blob/9d8015b2798c9469e0316a2bfce78bd0d44bace0/avalon/schema/representation-2.0.json#L57

We could potentially break this out into an individual schema.

> At any time in an application to switch task or asset; this is one of the most requested features in our studio. Switch user maybe?

Yeah, that's my impression too. Let's focus on these three for now.

BigRoy: Sure, we could do this:

```python
def update_current_session(asset=None,
                           task=None,
                           app=None,
                           user=None):
    """Update active session in-place to new context"""

    mapping = {
        "AVALON_ASSET": asset,
        "AVALON_TASK": task,
        "AVALON_APP": app,
        "AVALON_USER": user
    }
    changed = {key: value for key, value in mapping.items() if
               Session[key] != value}
    if not changed:
        return

    # Update silo when asset changed
    if "AVALON_ASSET" in changed:
        # Update silo for the asset
        asset_document = io.find_one({"name": asset,
                                      "type": "asset"},
                                     projection={"silo": True})
        assert asset_document, "Asset must exist"
        changed["AVALON_SILO"] = asset_document['silo']

    # Update work directory
    project = io.find_one({"type": "project"},
                          projection={"config.template.work": True})
    template = project['config']['template']['work']
    changed['AVALON_WORKDIR'] = format_template(template, Session)

    Session.update(changed)

    return changed
```

Or do the same whilst having a single argument like `context`?

This one could currently also be used like:

```update_current_session(**context)``` 
(if the context only has the allowed keys that can be changed)

Comment (mottosso) [reply avalon/io.py]: The "delta" implementation is cool, but I think we should stick with something simple for a 1.0.

Let's assume that making a change to any member of `context` updates the Session in the same way, and performs the same database operations.

`user` isn't part of `context`, let's make that separate. It's safe to assume you wouldn't want to change Asset and User at the same time.

Here's my suggestion.

```python
context = api.Context()
context["task"] = "animation"
api.update_current_context(context)

def update_current_context(context):
  workdir = compute_workdir(...)

  assert all_ok

  Session["AVALON_ASSET"] = context["asset"]
  Session["AVALON_TASK"] = context["task"]
  ...
```

avalon/maya/pipeline.py
```diff
@@ -479,3 +492,8 @@ def _before_scene_save(return_code, client_data):
     OpenMaya.MScriptUtil.setBool(return_code, True)
 
     api.emit("before_save", [return_code, client_data])
+
+
+def _on_current_context_updated(*args):
+    logger.info("Running _on_current_context_updated()")
+    _set_project()
```
mottosso: In order to protect against switching to a task with a non-existent project, we should check to see whether the project exists before switching to it.

Down the line, we can introduce a dialog to the user, asking him to create the project if it doesn't exist. For this to happen, we need to separate the folder creation aspect of the Launcher into core. One step at a time.

avalon/pipeline.py
```diff
@@ -720,6 +720,84 @@ def get_representation_context(representation):
...
+        dict: The changed key, values in the current Session.
+
+    """
+
+    mapping = {
+        "AVALON_ASSET": context.get("asset", None),
+        "AVALON_TASK": context.get("task", None),
+        "AVALON_APP": context.get("app", None),
+    }
+    changed = {key: value for key, value in mapping.items() if
```
mottosso: I would skip this.

Always update everything in the context. Regardless of whether no change was found. Down the line, when we're confident this works, we can look at making optimisations like these.

avalon/pipeline.py
```diff
@@ -720,6 +720,84 @@ def get_representation_context(representation):
...
+    template = project['config']['template']['work']
+    _session = Session.copy()
+    _session.update(changed)
+    changed['AVALON_WORKDIR'] = format_template(template, _session)
+
+    # Update the full session in one go to avoid half updates
+    Session.update(changed)
+
+    # Emit session change
+    emit("currentContextUpdated", changed.copy())
```
mottosso: I would expect to get the Context from this signal, not the environment variables from a partial Session object. What do you think?

BigRoy: This will pass on only whatever is changed. If at that point in time you'd like to get the active session state you can get it from `api.Session`.

BigRoy: Ah wait, you meant it would pass on the `asset`, `task` and `app`  keys. Hmm. That could be interesting. How about passing both context *and* the changed environment?

mottosso: That's a good idea.

But first, there's one thing that's miscommunicated I think. `context` is a dictionary with a fixed membership, like specified in its schema. It should be passed whole to the function, and returned whole. No partials.

```python
def Context():
	return {
	    "project": api.Session["AVALON_PROJECT"],
	    "asset": api.Session["AVALON_ASSET"],
	    "silo": api.Session["AVALON_SILO"],
	    "subset": api.Session["AVALON_SUBSET"],
	    "version": api.Session["AVALON_VERSION"],
	    "representation": api.Session["AVALON_REPRESENTATION"]"
	}

context = Context()
context["project"] = "hulk"
context["asset"] = "Bruce"
api.update_current_context(context)
```

However now that I type that out, I realise not all members are in Session, and not all members can or should be changed this way, like Version and Representation..

That's a flaw in this design; we can't call it Context if it isn't going to be a Context else we'd confuse it with the use of Context elsewhere.. I'm not sure how to deal with this one.

My problem with passing `project=` and `asset=` directly is that it's not discoverable enough. Which members should be part of this function? Subset too? Silo? There's no way to tell without reading the docs or by trail and error. The thing about Context is that it's known and used in more than one place. It's worth learning about and doesn't come with surprises.

Any ideas?

mottosso: I'm inclined to go with..

```python
api.update_asset("Bruce")
api.update_task("modeling")
```

Or, though cringy..

```python
api.set_asset("Bruce")
```

BigRoy: Ah, no. I see the confusion. The context with subset, version and representation is only relevant for a published file - exactly I'd go as far and let's say a *context* is always about a published file.

What we're updating here is the current Session, the current work area or workspace. It defines what "work environment" I'm currently in. That is **not** the context of a published file. This "work area" also has `app` and `task` which the context does not have - possibly even have `user` (I'd say).

I see to much issues with `update_asset` and `update_task` honestly. It might simplify some logic but it would ultimately also become slower.

> My problem with passing project= and asset= directly is that it's not discoverable enough. Which members should be part of this function? 

Hmm, interesting. I *love!* functions that expose the arguments with clear names, an IDE will let me know what I can pass along and often there's the docstring that complements the arguments.

I'd personally opt for either of:

1. Choose a single argument name that describes the dictionary in a clear concise way, e.g. `workspace` or `workarea` or `environment`
2. Give the function the exact arguments that are allowed, like `asset`, `task`, `app`. 

And additionally refactor the function name to something clearer, like `update_session_environment` or `update_session_workarea`?

mottosso: > I see to much issues with update_asset and update_task honestly. It might simplify some logic but it would ultimately also become slower and not necessarily safer because the whole work area is a "full set" working together.

Let's dive into that. I've got an idea of context switching will work from my days at Framestore and it's possible our ideas differ.

**Scenario 1**

1. User launches Maya, the current Session is fully qualified
2. User decides to switch from one asset to another
    - Clicks on the Current Asset icon, and is provided with a list of available assets
    - Picks one and Bam, asset updated.

**Scenario 2**

1. ...
2. User decides to switch project
    - Clicks on the Current Project icon, sees a list
    - Clicks on a relevant project
    - Current Asset button updates alongside it, to reflect that no asset has been set

In both of these cases, only one is ever updated at a time, and it's clear which of them does what.

With regards to performance.. I wouldn't worry about it. This could take seconds without being a problem.

> I love! functions that expose the arguments with clear names, an IDE will let me know what I can pass along and often there's the docstring that complements the arguments.

Agreed, but this isn't about arguments in general, but the fact that the function name doesn't hint towards what the arguments may be.

For example, consider this..

```python
api.update_task(...
```

What is the argument? Task!

```python
api.update_session(...
```

Now what is the argument? Task? Asset? Both? Either? A Session dictionary?

Your IDE may provide interactive hints, but a reader of the code and those without hints are left in the blind. Also I wouldn't lean on hints being available. They are an aid, not a necessity.

BigRoy: Both your scenarios sound alright, albeit that *scenario 2* is not supported in Avalon because it would require the `io` to uninstall and reinstall to another project. As such the current updating does not allow to switch project.

Anyhow, I'm seeing this scenario most often:

Scenario 3
1. User decides to pick up another asset's task
  - Clicks on the asset
  - Clicks on the task
  - Sets the work area

They have now switched to asset X and task Y.

BigRoy: I'm most concerned with "triggering updates" on the changes, say I'd like to initialize the new environment I have entered, but I don't now when I'm done switching to the new full context. Say I'm in task `modeling` for asset `hero` and want to go to asset `villain`'s task `lighting`.

If I do:
```python
api.set_asset("villain")
```

Will that now also update my `AVALON_WORKDIR` to `villain/modeling`? Would it even already start creating folders - say I'd want to automate it that far? How would I delay that until the work area switch was fully completed? (Say I'd want to avoid the `villain/modeling` to ever be initialized?

mottosso: Good point, how about we reset the task as well?

```
- project
   - asset
      - task
```

Then you could listen on just `taskChanged`.

BigRoy: > Good point, how about we reset the task as well?

Yes, **but**... then your Scenario 1. My environment would not be initialized because I'm only listening to `taskChanged`.

mottosso: I'm not sure I see  the problem. Your workarea would have changed only when the task has changed, so changing the project wouldn't trigger any new folders to be created? Only when you change the task does that happen?

```python
api.set_asset("Bruce")
# No folders are created
api.set_task("animation")
# Bak, task folders created
```

What's the alternative?

BigRoy: Hmm, actually. No? Unless when you do `api.set_asset("Bruce")` you are currently in limbo (no workspace) until you actually set a task.

Say I'm `Alice/modeling` and want to go to `Bruce/animation`.

```python
api.set_asset("Bruce")
# I'm now in `Bruce/modeling`; potentially I could start working! Actually, I just *entered* this workspace!
api.set_task("animation")
# I'm now in `Bruce/animation`
```

The alternatives are to "tell" when to enter the environment, where either:

1. You are in limbo until you trigger `api.enter_workarea()`
2. You have a single point of entry that capsulates your move to another work area, e.g. a single function.

mottosso: Ok, let's give your version a go. +1

BigRoy: Thanks for the good discussion though, it will be important to remain critical to aspects like these. I'll update the PR and see where we get.

mottosso: No worries. I think that limbo is fine, but hard to say without actually using it.

BigRoy:
Here we go, this starts making sense @mottosso!

mottosso:
Could you type up a summary of changes, along with an example of the task switching thingy?

BigRoy:
### Changes

This implements `update_current_task` and exposes it to the `api`. 

With it you can change your current work environment (live in a Session) to a new `asset`, `task` or `app`. Let's assume we're working in asset `Bruce` and are in the `modeling` task within app `maya`. 

Here's some steps we can take:

```python
api.update_current_task(task="lighting")
# You'll now be in Bruce/modeling/maya

api.update_current_task(asset="Claire")
# You'll now be in Claire/modeling/maya

# Each call to update_current_task will trigger a `taskChanged` signal
# If you instantly want to switch to another asset on another task
# you should do
api.update_current_task(task="rigging", asset="Jack")
# You'll now be in Jack/rigging/maya
```

To make this work, this has changed additionally:

- References to `os.environ` are replaced with the `api.Session` so it's always using the live session state. As such `os.environ` at some point (when updating the current task) could become out of sync with the `Session`. The `os.environ` describes the "launching environment" but the Session is the current live version of your current work environment, including the asset, task, app, etc.

- The `update_current_task` emits a `taskChanged` signal that Avalon's default maya host will connect to and call "_set_project" to try and remain in sync with the current task - also when being updated, see: https://github.com/getavalon/core/pull/300/files#diff-a21b0383dbeefa22023f1e9517b9fb68R497

avalon/pipeline.py
```diff
@@ -720,6 +720,85 @@ def get_representation_context(representation):
...
+        assert asset_document, "Asset must exist"
+        changed["AVALON_SILO"] = asset_document['silo']
+
+    # Compute work directory (with the temporary changed session so far)
+    project = io.find_one({"type": "project"},
+                          projection={"config.template.work": True})
+    template = project['config']['template']['work']
+    _session = Session.copy()
+    _session.update(changed)
+    changed['AVALON_WORKDIR'] = format_work_template(template, _session)
```
mottosso: Use double-ticks

avalon/pipeline.py
```diff
@@ -720,6 +720,85 @@ def get_representation_context(representation):
...
+    # Update the full session in one go to avoid half updates
+    Session.update(changed)
+
+    # Emit session change
+    emit("taskChanged", changed.copy())
+
+    return changed
+
+
+def format_work_template(template, session=None):
```
mottosso: Make this private.

mottosso:
Looking good. Some minor nitpicks, then merge at will!

tokejepsen:
@BigRoy Think there are some mistakes in the example:
´´´python
api.update_current_task(task="lighting")
# You'll now be in Bruce/lighting/maya

api.update_current_task(asset="Claire")
# You'll now be in Claire/lighting/maya

# Each call to update_current_task will trigger a `taskChanged` signal
# If you instantly want to switch to another asset on another task
# you should do
api.update_current_task(task="rigging", asset="Jack")
# You'll now be in Jack/rigging/maya
```

How does it work when switching to a different app?

tokejepsen:
> This is for switching within an app.

Ahh. Got confused by ```With it you can change your current work environment (live in a Session) to a new asset, task or app. ```

BigRoy:
Actually, it would support switching to a different app context - I'm not sure yet though what that would mean in the context of running it in a host application. You can do `update_current_task(app="nuke")` for example - since the session would allow it and there's no other logic involved.

Note that `app` doesn't necessarily reflect the `host`, as all `mayabatch`, `mayapy` and `maya` are "app" `maya` - because that's what defines the application directory.

avalon/pipeline.py
BigRoy: @mottosso across the file there are many lines that use single ticks - also on older lines. Do you want me to swap all?

avalon/pipeline.py
BigRoy: 👍

avalon/pipeline.py
BigRoy: @mottosso are you able to take a look at this? :) I'd love to get this ready for merging.

mottosso: Just these for now is fine, but try and stay consistent.

BigRoy:
That should clean up things. 👍

PR #302, opened 2017-11-28
enable host config uninstall

davidlatwe:
Was testing config swap by `Reload Pipeline` inside Maya, found that `avalon.pipeline.uninstall()` did trigger the uninstall of pipeline config, but host config's uninstall did not run.

### DIFF
--- a/avalon/maya/pipeline.py
+++ b/avalon/maya/pipeline.py
@@ -53,11 +53,8 @@ def install(config):
     pyblish.register_host("mayapy")
     pyblish.register_host("maya")
 
-    try:
-        config = importlib.import_module(config.__name__ + ".maya")
-    except ImportError:
-        pass
-    else:
+    config = find_host_config(config)
+    if hasattr(config, "install"):
         config.install()
 
 
@@ -82,17 +79,23 @@ def _set_project():
     cmds.workspace(workdir, openWorkspace=True)
 
 
+def find_host_config(config):
+    try:
+        config = importlib.import_module(config.__name__ + ".maya")
+    except ImportError:
+        config = None
+
+    return config
+
+
 def uninstall(config):
     """Uninstall Maya-specific functionality of avalon-core.
 
     This function is called automatically on calling `api.uninstall()`.
 
     """
-    try:
-        config = importlib.import_module(config.__name__ + ".maya")
-    except ImportError:
-        pass
-    else:
+    config = find_host_config(config)
+    if hasattr(config, "uninstall"):
         config.uninstall()
 
     _uninstall_menu()
--- a/avalon/maya/pipeline.py
+++ b/avalon/maya/pipeline.py
@@ -23,6 +23,7 @@ self._menu = "avalonmaya"  # Unique name of menu
 self._events = dict()  # Registered Maya callbacks
 self._parent = None  # Main Window
 self._ignore_lock = False
+self._host_config = None
 
 AVALON_CONTAINERS = ":AVALON_CONTAINERS"
 IS_HEADLESS = not hasattr(cmds, "about") or cmds.about(batch=True)
@@ -59,6 +60,7 @@ def install(config):
         pass
     else:
         config.install()
+        self._host_config = config
 
 
 def _set_project():
@@ -88,6 +90,11 @@ def uninstall():
     pyblish.deregister_host("mayabatch")
     pyblish.deregister_host("mayapy")
     pyblish.deregister_host("maya")
+
+    try:
+        self._host_config.uninstall()
+    except AttributeError:
+        pass
 
 
 def _install_menu():
--- a/avalon/maya/pipeline.py
+++ b/avalon/maya/pipeline.py
@@ -23,7 +23,6 @@ self._menu = "avalonmaya"  # Unique name of menu
 self._events = dict()  # Registered Maya callbacks
 self._parent = None  # Main Window
 self._ignore_lock = False
-self._host_config = None
 
 AVALON_CONTAINERS = ":AVALON_CONTAINERS"
 IS_HEADLESS = not hasattr(cmds, "about") or cmds.about(batch=True)
@@ -60,7 +59,6 @@ def install(config):
         pass
     else:
         config.install()
-        self._host_config = config
 
 
 def _set_project():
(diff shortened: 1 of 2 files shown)


mottosso:
Looks good to me, but don't you think the config should be uninstall first? It's installed [*after*](https://github.com/getavalon/core/blob/79991d3b61877385c1cbf3704b2c24dcc615404a/avalon/pipeline.py#L76) the core, so to avoid garbage collection from breaking the config before it is uninstalled, it should probably follow the same order but backwards.

davidlatwe:
Thanks, Marcus.
Ummm sure, but isn't that backwards already ?
From `core.pipeline` [uninstall](https://github.com/getavalon/core/blob/79991d3b61877385c1cbf3704b2c24dcc615404a/avalon/pipeline.py#L96-L106), there calls `registered_host.uninstall`, which will run `config.maya.uninstall` in my PR, and back to run `config.uninstall` next.

mottosso:
You're right, that's a bug. It should be the other way around.

**Install**

1. install core
2. install config

**Uninstall**

1. Uninstall config
2. Uninstall core

davidlatwe:
Or do you mean that the uninstall order of `registered_host` and `registered_config` in `core.pipeline.uninstall` should be reversed first ?

mottosso:
Or wait, no that is right. It's uninstalling the host first, config second. The host is installed last, so it should be uninstalled first.

The point I'm trying to make is, things should be uninstalled in the reverse order that they were installed.

davidlatwe:
Ohoh, I did think of that, then the install/uninstall of sub-config (host config) should be handled by the config itself.
If so, the config needs to know which host is running, maybe this is related to the `api.discover(api.Application)` you mentioned in gitter ?

davidlatwe:
Hey Marcus, I am a bit confused about the current install/uninstall order suggestion you said, maybe see if I understand this right ?

---

### currently
##### Install
* call `avalon.api.install(maya)`
    * do - **global**
    * found `global config`
    * call `avalon.maya.install(config)`
        * do - **host**
        * found `maya config`
        * call `maya config.install()`
            * do - **host-config**
    * call `global config.install()`
        * do - **global-config**

##### Uninstall
* call `avalon.api.uninstall()`
    * call `avalon.maya.uninstall()`
        * undo - **host**
        * call `maya config.uninstall()` (this PR)
            * undo - **host-config**
    * call `global config.uninstall()`
        * undo - **global-config**
    * undo - **global**

### idealy
##### Install
* call `avalon.api.install(maya)`
    * do - **global**
    * found `global config`
    * call `global config.install()` ---- (change 1)
        * do - **global-config**
    * call `avalon.maya.install(config)`
        * do - **host**
        * found `maya config`
        * call `maya config.install()`
            * do - **host-config**

##### Uninstall
* call `avalon.api.uninstall()`
    * call `avalon.maya.uninstall()`
        * call `maya config.uninstall()` (this PR)
            * undo - **host-config**
        * undo - **host** ------------------ (change 2)
    * call `global config.uninstall()`
        * undo - **global-config**
    * undo - **global**

---

If so, to be **idealy**, the change needs to be made are :

1. move this [line](https://github.com/getavalon/core/blob/79991d3b61877385c1cbf3704b2c24dcc615404a/avalon/pipeline.py#L76) `config.install()` up before `host.install(config)`

2. move `host config.uninstall()` (this PR) up top before these [line](https://github.com/getavalon/core/blob/79991d3b61877385c1cbf3704b2c24dcc615404a/avalon/maya/pipeline.py#L86-L90)

Is that right ?

mottosso:
Nevermind what I said; I've overcomplicated the situation.

To tackle the topic of this PR, isn't the config is uninstalled [here](https://github.com/getavalon/core/blob/79991d3b61877385c1cbf3704b2c24dcc615404a/avalon/pipeline.py#L104)?

You say that it doesn't uninstall when you do reload.. I'd really rather we didn't focus too much on that particular feature. It's very hacky and unlikely to work in most cases.

When you develop, it's best to reload the specific module you are modifying, and restart Maya for anything more major.

When you call this, does it behave as you'd expect?

```python
from avalon import api
api.uninstall()
```

davidlatwe:
> When you develop, it's best to reload the specific module you are modifying, and restart Maya for anything more major.

Oh, yes, of course. I just use that reload button as a convenient to do `api.uninstall()` :)

> To tackle the topic of this PR, isn't the config is uninstalled here?

Yes, It did call global config's `uninstall()`, but did not call `config.maya.uninstall()`.

`config.maya` registered some maya pyblish plugin path with it's `install()`, when I run `api.uninstall()`, the pyblish *global* plugin path did remove, but *maya* plugin path still exists.

I checked that with
```python
pyblish.api.registered_paths()
```

mottosso:
> Yes, It did call global config's uninstall(), but did not call config.maya.uninstall().

Aah, I see. What do you think about [this](https://github.com/getavalon/core/blob/79991d3b61877385c1cbf3704b2c24dcc615404a/avalon/pipeline.py#L71)?

```python
    # Optional host install function
    if hasattr(host, "install"):
        host.install(config)
```

It looks like what's missing is passing the config to the [host `uninstall` below](https://github.com/getavalon/core/blob/79991d3b61877385c1cbf3704b2c24dcc615404a/avalon/pipeline.py#L99).

[This one](https://github.com/getavalon/core/blob/79991d3b61877385c1cbf3704b2c24dcc615404a/avalon/maya/pipeline.py#L85) should take a config argument, and call the [corresponding uninstall](https://github.com/getavalon/core/blob/79991d3b61877385c1cbf3704b2c24dcc615404a/avalon/maya/pipeline.py#L57) for the particular config for this host, such as Maya.

Let me know if that makes sense!

davidlatwe:
Yeah, I think this make more sense, avalon will only store the main config module in core, and have the same logic with `host.install()`.

Then, for passing the config to `registered_host().uninstall(config)`, which one should I query for ?
[`self._config`](https://github.com/getavalon/core/blob/79991d3b61877385c1cbf3704b2c24dcc615404a/avalon/pipeline.py#L79) or `registered_config()` ?

I think `registered_config()` is a good go, having a more readable name.

```python
# core/pipeline.py
def uninstall():
    """Undo all of what `install()` did"""
    config = registered_config()
    try:
        registered_host().uninstall(config)
    except AttributeError:
        pass

    try:
        config.uninstall()
    except AttributeError:
        pass

    deregister_host()
    deregister_config()

    # more undo
```
[core/pipeline.py#L96-L101](https://github.com/getavalon/core/blob/79991d3b61877385c1cbf3704b2c24dcc615404a/avalon/pipeline.py#L96-L101)

Can we add a function for host finding host-config ?

```python
def find_host_config(config):
    try:
        config = importlib.import_module(config.__name__ + ".maya")
    except ImportError:
        config = None
    
    return config
```

mottosso:
> I think registered_config() is a good go, having a more readable name.

Sure, sounds ok.

> Can we add a function for host finding host-config ? in separate commit.

How do you mean? This is the function for finding it? Have you got another way of finding it in mind? Then sure, let's pop up a separate PR for it.

davidlatwe:
> This is the function for finding it?

Yes, I thought there will be duplicated code for finding host config,
```python
config = importlib.import_module(config.__name__ + ".maya")
```
if passing config to host's install/uninstall.

Not another way, just putting them together. It's in the last commit now :)

mottosso:
I see, this looks good! Let me know when you're happy and I'll merge.

davidlatwe:
Thank you, @mottosso !
Merge, please :D

Issue #293, opened 2017-10-25
Implement Qt styling for Avalon tools


BigRoy:
# Issue

Currently we've been enjoying the Avalon tools most of the time in Maya only where they took over the Maya styling and looked fine. Outside of Maya though they use the default OS styling and seem to be lacking the good looks that we want, and in some cases it even becomes a somewhat unreadable mess (like in `cbloader` currently)

Let's set up a Qt stylesheet for Avalon to always look great no matter the circumstances.

mottosso:
I remember you had a stylesheet set-up?

BigRoy:
Yes, but that was based on http://discourse.techart.online/t/release-qt-dark-orange-stylesheet/2287

Not sure if we can mimick that with the licensing we've got going. Especially if we change it here and there too.

mottosso:
Don't worry about that, it doesn't apply here.

BigRoy:
> Don't worry about that, it doesn't apply here.

How come?

mottosso:
Because we aren't using it, we are building something that once drew inspiration from it. Still, in the worst of cases, what kind of repercussions are you expecting? That the author will sue? Let's include his name if you'd like. Or better yet, why not ask him for it?

BigRoy:
Actually I just found the source repo I started from at the time: https://github.com/ColinDuquesnoy/QDarkStyleSheet

I really don't have much experience with styling Qt-applications and I'm not even sure whether the workflow in that repository is well suited for something like Avalon.

My "changed" version has the exact same code layout and only has some changed colors and padding here and there, they are more additional tweaks than a complete overhaul. That's why I also never used it aside from some personal apps here. So I think getting the licensing right (or at least us altering it without copyright issues) would be important too. Or to start from scratch, or at least redesign it largely.

mottosso:
Ah, but that's even better, because that's actually got a license, MIT. :) So we're able to use it as-is, bundle it, ship it, distribute it, make a commercial product of it even. No limits!

mottosso:
> I really don't have much experience with styling Qt-applications and I'm not even sure whether the workflow in that repository is well suited for something like Avalon.

No problem. I'm happy to help. CSS is a good route to standardising the look of applications. All we'd need is either take [this](https://github.com/ColinDuquesnoy/QDarkStyleSheet/blob/master/qdarkstyle/style.qss) file, and add another file for our customisations. We can use multiple files by simply adding them together..

```python
style1 = """QWidget {background: red}"""
[...]

PR #307, opened 2017-12-06
Implement avalon style

BigRoy:
This adds an Avalon stylesheet (#293)

> Note that this is based on QDarkStyleSheet as discussed in the issue - I'm mentioning it here to ensure we check whether we would be safe regarding licensing. The files I have pushed here do **not** have any license information included.

Here's an example of the stylesheet on the project manager:
afbeelding

---

Currently the [application context manager in the tools lib](https://github.com/getavalon/core/compare/master...Colorbleed:style?expand=1#diff-592b29662f6fe9ce137804910b8f6bd5R25) automatically applies the stylesheet for new QApplication instances when running outside of software. As such the stylesheet is not applied if a QApplication is already running. If we were to apply the stylesheet to a running QApplication instance it would also apply it to e.g. Maya itself. To apply the stylesheet to the windows or dialogs themselves we'll need to apply the stylesheet on the widgets themselves instead of the application.


BigRoy:
The massive list of Hound errors above are on compiled resources files for Qt. I assume we should safely ignore those.

mottosso:
Whop!

Yes, try adding an `ignore_file: ` section to `.hound.yml`

- https://houndci.com/configuration

It might not work with Python, but worth a go.

mottosso:
Should we just merge this? :P

BigRoy:
> Currently the application context manager in the tools lib automatically applies the stylesheet for new QApplication instances when running outside of software. As such the stylesheet is not applied if a QApplication is already running. If we were to apply the stylesheet to a running QApplication instance it would also apply it to e.g. Maya itself. To apply the stylesheet to the windows or dialogs themselves we'll need to apply the stylesheet on the widgets themselves instead of the application.

What did you think of this? When and where would you like to apply the stylesheet?

mottosso:
Oh, I see. We can just apply it to the window of each tool. That should be straightforward enough. In the show function.

BigRoy:
That should cover it. The hound is shouting about the same old things which shouldn't be an actual problem. @mottosso should we merge like this?

mottosso:
@BigRoy Yes

Issue #310, opened 2018-03-01
Switch subset with Loader


aardschok:
## Goal
Allow user to change the subset of a loaded asset.

## Motivation
A situation has arrived which can perfectly describe the support for this feature.
We want to automate the step from render to slap comp. This will require the following things:

* Rendered image sequences
* A predefined comp (created per project)
* A logic to update the subset / version / representation of any Loader in the comp

The current demands for this logic are that it will need to be able to clear a container from data and re-imprint it with new data. This will result in the new subset being loaded with the latest version, key here is __latest__.
The logic will only be able to use the Loader from the existing container to update the subset/version/representation, this will also keep us away from headaches and over complex functions.

It will look something along the lines of this, function in the `api.Loader` of `core/avalon/pipeline` for Fusion:

```python
# PSEUDO FUNCTION OF api.Loader
def switch(container, representation):
    from avalon.fusion import comp_lock_and_undo_chunk, imprint_container, clean_container
    
    host =  api.registered_host()
    tool = container["_tool"]
    assert tool.ID == "Loader", "Must be Loader"
    comp = tool.Comp()

    root = api.get_representation_path(representation)
    context = api.get_representation_context(representation)
    path = self._get_first_image(root)
    
    
    namespace = context['asset']['name']
    subset = context["subset"]["name"]
    version = context["version"]
    family = version["data"]["families"][0]

    # Ensure this subset can be loaded
    assert family in self.families, "Unsupported family for loader `%s`" % self.label

    clean_container(tool)  # here we strip all the attributes of it's values
    imprint_container(tool,
                                subset,
                                namespace,
                                context,
                                self.__class__.__name__
```

## Other uses
This will also allow the user to swap different subsets of let's say, cameras. Instead of creating a new camera version for stereo cameras the user can simply swap a `cameraDefault` subset for `cameraStereo` or `stereoCamDefault`, assuming they are using the same Loader.

Another example is swapping Level of Detail models of the same character. Let's say the camera is so close to an object that you will want to update it to the high resolution. `modelLOD1` is loaded but we need `modelLOD0` (lod0 being closest to the cam and lodX the most distant) 

## Conclusion
Implement swapping of subset when it meets these requirements:

* Family supported by Loader used
* Representation supported by Loader used

This will mean a new function to strip containers of their values.

BigRoy:
> The logic will only be able to use the Loader from the existing container to update the subset/version/representation, this will also keep us away from headaches and over complex functions.

[...]

PR #312, opened 2018-03-02
Introduce switch logic

aardschok:
* Introduces switch logic for #310 
* Moved assertion up to ensure it works as intended
* Import `switch` and add it to __all__

Example code to test function where we have a scene with only the asset hero loaded

```python
import avalon.io as io
import avalon.api as api

host = api.registered_host()

# Example 1:
# Swap subset modelDefault of hero for modelLod0 of hero
asset = io.find_one({"type": "asset", "name": "hero"})
subset = io.find_one({"type": "subset", 
                      "name": "modelLod0",
                      "parent": asset["_id"]})

version = io.find_one({"type": "version", 
                       "parent": subset["_id"]},
                        sort=[('name', -1)])

representation = io.find_one({"type": "representation", "parent": version["_id"]})

container = list(host.ls())[0]
api.switch(container, representation)


# Example 2:
# Scene contains model of hero but model needs to be of villain
asset = io.find_one({"type": "asset", "name": "villain"})
subset = io.find_one({"type": "subset", "name": "modelDefault",
                      "parent": asset["_id"]})

version = io.find_one({"type": "version", 
                       "parent": subset["_id"]},
                      sort=[('name', -1)])

representation = io.find_one({"type": "representation",
                              "parent": version["_id"]})

container = list(host.ls())[0]

api.switch(container, representation)
```


BigRoy:
This looks good. Just to be sure, this is functional - right? All that's needed is to implement a `switch` method on the Loader classes that the studio's  configuration has implemented, correct?

--- 

Aside from that, I think the next step is to have one of the tools expose this switch functionality to the artist. Which is slightly more involved and could maybe become a PR of its own to keep this one lightweight? @mottosso what do you think?

aardschok:
> Just to be sure, this is functional - right?

Yes it is functional, as you said, a `switch` method is needed in a Loader in order to make it work.

aardschok:
Here is an example of how it can work. You can fill in all three of the fields to switch to a specific representation.

Switch

mottosso:
Kick ass :D

BigRoy:
This is magic. I believe this is mergable. :muscle:

BigRoy:
Also, should we bump the minor version of Avalon? (I believe we haven't done that recently with other PRs).
I also noticed just now that the Avalon version is set to `5.0.0` - any reason it's already at 5?
Should we maybe bump this to 5.1.0 then?

@mottosso what do you think?

mottosso:
Yeah it's semantic versioning and major version increments is for when it breaks backwards compatibility. That's how many times that has happened so far. Ideally it wouldn't happen so frequently, but I'd recommend we keep doing it as both a reminder and notifier for others using Avalon.

BigRoy:
I bumped the minor version. I'd say, @aardschok, feel free to merge once the checks have passed.

PR #318, opened 2018-06-07
Add "ls" command to list projects in database.

tokejepsen:
This will list all projects in the database.

```bat
> python -u -m avalon.inventory --ls
Projects in database:
- batman
- myProject
```

### DIFF
--- a/avalon/inventory.py
+++ b/avalon/inventory.py
@@ -408,12 +408,11 @@ def _cli():
     root = kwargs.root or os.getcwd()
     name = kwargs.load or os.path.basename(root)
 
-    install = False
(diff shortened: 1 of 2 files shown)


BigRoy:
@tokejepsen this looks good.

I'm not sure the changes on the _install_ lines are necessary, I liked the brief previous version which was just a single if statement. I believe it operates the same, so it's just cosmetics. Just wanted to point this out.

mottosso:
Would it be possible to include some example output?

tokejepsen:
Yup, sorry. Bit sloppy PR. Have updated the description.

mottosso:
Haven't tested whether it technically works, but the tests pass so that's a good sign.

tokejepsen:
Confirmed here, and its looks good to me visually.

Issue #319, opened 2018-06-29
Work Files


tokejepsen:
**Goal**

Alleviate work files navigation.

**Motivation**

Loading and saving work files can becomes messy fast.

With taking control over where and how work files should be saved, not only makes a tidier file system but also makes users more independent and faster as they don't have to think about files on a disk but can concentrate on content creation.

**Implementation**

Opening/creating work files should happen between choosing which application to launch and actually launching the application.
Most applications are designed to open files directly, so its just a matter of adding to the launch command.

The location of the work files would need to be defined prior to asking the user to open/create.

Choosing which files to open or create, would require some kind of GUI. This could be a simple list with a ```Create``` button.

The naming of the work files would need defining as well. I'm not confident with the Avalon object model, but guess it'll be defined through that somehow.

mottosso:
There's been talk about this at some point and I'm personally not a fan - I think creation need to be as flexible and open ended as possible to account for the many and diverse personalities and preferences of said personalities in a given project - but if it's optional I don't see a problem with it.

tokejepsen:
Think the GUI will need to be as easy as a file browser, if not better.

Have previously done a dialog with ```Open Latest Version``` or ```Open Previous Version```. ```Open Previous Version``` would just simply open a file browser in the folder with all the work files.

PR #322, opened 2018-07-05
Work Files

tokejepsen:
This PR is to discuss an implementation of #319 

capture

The initial question is how would this fit into the Avalon pipeline?

My suggestion would be that you supply Work Files with an application path, root path and optional filters.

```bash
python -m avalon.tools.workfiles --root /path/to/work/area --application /path/to/maya.exe --filter .ma --filter .mb
```

Once the user is done with creating or selecting a work file, press "Open" and the Work Files just outputs the path. It'll then be up to the launcher to generate and execute a command.

```Browse``` will just be a file browser, so the user can pick any file according to the filters.

### CODE REVIEW

**Comment by tokejepsen** [comment]
A question that pops up is whether we should overwrite when creating new work files?

**Comment by BigRoy** [comment]
> A question that pops up is whether we should overwrite when creating new work files?

Let's not overwrite without question. It could pop up "file already exists" or something? But preferrably I wouldn't want it to overwrite current things.

**Comment by tokejepsen** [comment]
> Let's not overwrite without question. It could pop up "file already exists" or something?

Cool, have been raise errors so far. So will just raise can error for the time being.

**Comment by tokejepsen** [comment]
I'm trying to work out where the workfiles app fits into the Avalon launching processes.

The best place I think of is in the [launch method](https://github.com/getavalon/core/blob/899b8896cda9d6c5426a21eee01a17a1e277e8fe/avalon/lib.py#L208), as another argument; ```workfiles_app=True```.
By default this would be ```False```, so Actions would have to explicitly opt-in. For Actions coming from ```.toml``` files, these could opt-in with a config flag ```workfiles_app = True```.

I will also make the workfiles app available from with Maya, so people can generate new versions.

**Comment by BigRoy** [comment]
I'm not sure that `launch` method is the correct place to put that particular code - I like that specifically that method is so generic. :) What other options do we have?

**Comment by BigRoy** [comment]
> Actually working on it some more revealed it was better to have it in the default action launch. Still working through it, but seems like a much better option atm.

Yes, makes more sense there. :+1:

**Comment by tokejepsen** [comment]
First preview of the final application:

jul 19 2018 5_47 pm

Feedback is welcome! There is also some [documentation](https://github.com/tokejepsen/core/blob/work_files/avalon/tools/workfiles/README.md)

**Comment by mottosso** [comment]
Looks very pretty, but I'm confused; does this save a new Maya file from the Launcher? What is in the file? And why does it take so long to create it? (I noticed a few seconds between hitting Create and getting interactivity back)

I thought this was an app to be used from within Maya (and others) to use when you save/open files?

**Comment by mottosso** [comment]
Oh, I see now, it launches a Maya in the background, and saves an empty scene.

- https://github.com/tokejepsen/core/blob/work_files/avalon/tools/workfiles/maya_workfile.py

Still a little unsure of the benefit over saving it from an actual Maya session, post-launch.

**Comment by tokejepsen** [comment]
> I thought this was an app to be used from within Maya (and others) to use when you save/open files?

You can also use it within Maya.

> Still a little unsure of the benefit over saving it from an actual Maya session, post-launch.

Consistency in naming and getting the user away from fiddling in the file system.

**Comment by tokejepsen** [comment]
We can also take this further and have the latest version or last modified work file, be selected so users are just one click away from continuing the work.

**Comment by BigRoy** [comment]
Does it really need to start maya in the background to create a first file? I can see it make sense for consistency but might be a tad slow? Of course it could be swapped with a "standard template file" or something alike down the line. (Way too much optimization in my head now)
Should this particular logic on saving the maya scene (or whatever format) actually be in core?

Anyway, it looks good. I can see it be of use when you quickly want to open maya on a specific scene file too. But I'd really have to try it and see whether we would use it in the launcher too.

Nice prototype! :)

**Comment by tokejepsen** [comment]
> Does it really need to start maya in the background to create a first file? I can see it make sense for consistency but might be a tad slow? Of course it could be swapped with a "standard template file" or something alike down the line

In my experience you need to have the application make the file so you don't get into complications about application versions etc.
Optimizations wise you could maybe pre-generate a file per application version.

> Should this particular logic on saving the maya scene (or whatever format) actually be in core?

Thought it was nice to have the code inside the workfiles app. Whether it needs to be in core, I don't know.

> Anyway, it looks good. I can see it be of use when you quickly want to open maya on a specific scene file too. But I'd really have to try it and see whether we would use it in the launcher too.

I think where the quickness will come from (at least in my experience) will be from pre-selecting the latest version or last modified file. The user would be two clicks away from loading a work file, rather than launch the application first then finding the work file.

**Comment by tokejepsen** [comment]
> I thought this was an app to be used from within Maya (and others) to use when you save/open files?

Just realized I have not given the user an option to use the app when saving files within a host. Will rethink the design a bit for this.

**Comment by tokejepsen** [comment]
> Just realized I have not given the user an option to use the app when saving files within a host. Will rethink the design a bit for this.

Maybe the "New" button should turn into a "Save" button when you are within a host?

There is no real need for a "New" button within a host, because this is already provided.

**Comment by mottosso** [comment]
Yeah, should work. Doesn't even need a GUI to save. Just for opening. For saving, maybe it could also be helpful to include something to distinguish files, apart from just a version. Such as "blocking" or "toRender". A user comment of sorts.

But that could be just me; I'm not the target audience in this case.

**Comment by tokejepsen** [comment]
> For saving, maybe it could also be helpful to include something to distinguish files, apart from just a version. Such as "blocking" or "toRender". A user comment of sorts.

Yeah, I have thought if including an optional ```comment``` input and data member for the template. Don't know how many people use this, but since two people (me and you) so far have thought of it, I'll give the option to include it in the template.

**Comment by BigRoy** [comment]
> Yeah, I have thought if including an optional comment input and data member for the template. Don't know how many people use this, but since two people (me and you) so far have thought of it, I'll give the option to include it in the template.

Yes, some artists love that too - it's not a studio requirement yet but I can see it be of use.

**Comment by tokejepsen** [comment]
> Yes, some artists love that too - it's not a studio requirement yet but I can see it be of use.

I could make both ```version``` and ```comment``` optional inputs based on the template?
If the template have ```{comment}```, a line edit will appear. Same for ```version``` but with spinbox.

**Comment by tokejepsen** [comment]
New update:

jul 20 2018 5_19 pm

Inside Maya:

jul 20 2018 6_04 pm

[Updated docs](https://github.com/tokejepsen/core/blob/315bf681195334fb91ddc8cff75243863d536809/avalon/tools/workfiles/README.md)

New features:
- Last modified file is pre-selected
- Save files within hosts (Maya)
- Comment input
- Optional template groups

**Comment by tokejepsen** [comment]
I agree that in an ideal world, all content that is shared between two people should have been validated and published.
What I don't agree with is that this tool encouraged not publishing content.

In the current version of Avalon there is nothing stopping people from bypassing publishing and just opening/referencing each others work files. This is not something we can control in any way.
With the work files app we could actually control which work files are available to open for a user.

**Comment by mkolar** [comment]
We're hearing this request from artists very often. Most of the time this isn't used to handover files at all. It's simply making the artist life easier, by removing the need to be double checking that he's creating the correct name for his file. 

We've actually handled it till now by at least having the first workfile creation automated. after that it's in artist hands, but this simple GUI that Toke wrote is really what quite a few people are asking for. 

I'm throwing a big thumbs up.

**Comment by BigRoy** [comment]
I can confirm what Milan/Toke say. I have been holding off implementing this as I still am not sure what the best interface is for this but I believe this already goes a long way.

Previously we had a full blown "explorer" replacement that acted as something like this where artists could easily save/browse/open but functionality wise it never becomes a real OS explorer without a lot of code. I think having it simpler like this could already help.

To be sure, could this also be used in an app (e.g. Maya) to open an existing file? Or is it only intended for saving there?

**Comment by tokejepsen** [comment]
> To be sure, could this also be used in an app (e.g. Maya) to open an existing file? Or is it only intended for saving there?

Yeah, the in app part is not just for saving but you can select and open a workfile as well. See the open button in this comment:
https://github.com/getavalon/core/pull/322#issuecomment-406648528

**Comment by davidlatwe** [comment]
Also, could we make this app able to accepts something like `on_open` / `on_save` callbacks for us to do some custom file locking and unlock ?

**Comment by BigRoy** [comment]
Regarding the event, I am wondering if you are looking for something other than the `on_open` and others that Avalon can already register for a host, such as maya for example? Or that you have additional ideas for events that specifically this work files app would emit? And if so, could you elaborate why you would like them separate? For example with a specific use case.

**Comment by davidlatwe** [comment]
> Are you referring to emitting an event? https://getavalon.github.io/2.0/reference/#events

Yes, `events` emitting !

>  looking for something other than the on_open and others that Avalon can already register for a host, such as maya for example?

I only having some burly idea for now, it's mainly about checking some condition and decide whether or not to `SaveAs` the file, or restrict the `subset instance` publish ability in that file been copied or saved, and the conditions may based on *user name* or *department name* or else.

And perhaps the work file app could be other tool's prompt window and able to do a bit more control with event emitting. The core motivation was to be able to keep bad hand away by each studio's needs in the future :P

But I maybe off topic or not keeping things simple : ( what do you think ?

**Comment by mottosso** [comment]
Thanks guys for taking into consideration my wall of text; I think you all make good points and have no further objections. Full steam ahead!

**Comment by tokejepsen** [comment]
> I only having some burly blurry idea for now, it's mainly about checking some condition and decide whether or not to SaveAs the file, or restrict the subset instance publish ability in that file been copied or saved, and the conditions may based on user name or department name or else.

Maybe we should leave the app as simple as possible?
We can always add more features once use cases has been discovered in production. Problem is removing existing features.

> Inside Maya, the Save button only perform save as on current file, maybe we can have a Current File item always stay on top of the list, and change button name from Save to Save As, then the Save As can also include the files in workDir and do copy & rename just like in app save as.

I have had some thoughts about this.
I think I would like this app to be faster than going into an applications ```File > Recent Files```, in order to encourage adoption. This means faster than two clicks to get your last working file.

In the current version, where the latest modified file is pre-selected, to get to your last working file you are one click away; ```Open```.
If ```Current File``` was selected, then we are back to two clicks to get to the latest working file.

I may very well be overthinking this. Thoughts/suggestions are welcome :)

> And maybe we won't need WorkFile app to start at Launcher if we involve Current File option on list ?

In my experience with an application like TVPaint, where you don't have much control over the startup sequence or in-app menus, its good to have some control over the work files outside of the application.

Although having WorkFile app only inside of the applications would greatly simplify the code and impact on for example the Launcher.

**Comment by tokejepsen** [comment]
Update :)

I have decided to discard the outside of hosts functionality. It added a lot of complexity with very little benefits. Now it is only within a host (Maya) where you can use the Workfiles app. Thank you @davidlatwe for some perspective on this!

aug 3 2018 7_00 pm

[Updated docs](https://github.com/tokejepsen/core/blob/3ee04d8276e53fd07dd9d9c0e0fc1ee4061f75f0/avalon/tools/workfiles/README.md)

- Duplicate button for copy and rename a work file.
- "Next Available Version" checkbox for automatic finding the next version number to save as.
- Reorganized the layout

**Comment by tokejepsen** [comment]
Forgot to mention that if you wanted, you could show the Workfiles app as part of the startup sequence. This is added per config like so:

```python
workfiles.show(
    root=os.path.join(
        cmds.workspace(query=True, rootDirectory=True),
        cmds.workspace(fileRuleEntry="scene")
    ),
    executable=sys.executable
)
```

tokejepsen:
What do you guys think of the current version? Mergable?

mottosso:
Happy to merge ASAP. :)

tokejepsen:
Here we go :)

PR #326, opened 2018-07-25
Add list, tuple type support to `maya.imprint()` and other two minor tweak

davidlatwe:
Hi, I made some tiny changes, here's short brief:

#### 1) `avalon.maya.lib.imprint()` add `list`, `tuple` type support
* Motivation
Thought that would be nice to have `imprint` to create enum type attribute for us.
Example use case, choosing cache format (abc, gpu, fbx...) to export with pointcache creator. 

* Implementation
Implemented with Maya's `enum` type attribute.

#### 2) Change the type of `avalon.pipeline.Creator.data` from `dict` to `OrderedDict`
* Motivation
After adding some custom attributes into `creator.data`, those attributes visualized in GUI (Maya Attribute Editor) without order is not very comfort to see, custom attributes often mixed in core attributes.
* Implementation
Use `OrderedDict`.

#### 3) Larger creator icon border padding
* Motivation
Some icon still get cropped with 2 padding.
* Implementation
Bump to 4. :smile:

Please let me know what you think, thanks !

### DIFF
--- a/avalon/maya/lib.py
+++ b/avalon/maya/lib.py
@@ -195,6 +195,10 @@ def imprint(node, data):
         elif isinstance(value, float):
             add_type = {"attributeType": "double"}
             set_type = {"keyable": False, "channelBox": True}
+        elif isinstance(value, (list, tuple)):
+            add_type = {"attributeType": "enum", "enumName": ":".join(value)}
+            set_type = {"keyable": False, "channelBox": True}
+            value = 0  # enum default
         else:
             raise TypeError("Unsupported type: %r" % type(value))
 
--- a/avalon/maya/lib.py
+++ b/avalon/maya/lib.py
@@ -82,7 +82,7 @@ def read(node):
 
     for attr in cmds.listAttr(node, userDefined=True) or list():
         try:
-            value = cmds.getAttr(node + "." + attr)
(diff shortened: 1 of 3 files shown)


BigRoy:
I think this is great.

Does the [`read`](https://github.com/getavalon/core/blob/master/avalon/maya/lib.py#L78) also correctly read the new Enum value? What does it return in that case? An index or the string value that was chosen by the user? Should it return the string?

davidlatwe:
Thanks, @BigRoy !

`read()` in this case, will return the menuItem index of the value chosen by user.

> Should it return the string?

Good idea ! The implementation of this only require adding `asString=True` to this [line](https://github.com/getavalon/core/blob/899b8896cda9d6c5426a21eee01a17a1e277e8fe/avalon/maya/lib.py#L85), won't affect other value type of attribute query.
In my current use case, returning `int` or `str` doesn't matter, but could improve log reading.

BigRoy:
If `asString` doesn't break other queries then I think that'll be it and sounds like the best returned output.

Sounds good to me.

aardschok:
Yes! Tested it out, work like a charm :)

BigRoy:
Let's do this.

PR #328, opened 2018-07-26
Modules from path

tokejepsen:
Refactor the logic for getting modules from a path, to ```lib.py``` for public use by other apps.

This is an effort for https://github.com/getavalon/launcher/issues/35


tokejepsen:
Yeah, I don't know what is going with those commits.

Even tried to make this PR several times with the same results.

PR #329, opened 2018-08-06
Nuke workflow

tokejepsen:
This is a work-in-progress PR for Avalon integration with Nuke.

Initially the ```Avalon``` menu is implemented.


BigRoy:
Has anyone been able to play around some more? Maybe @mkolar?

tokejepsen:
For an initial Nuke integration this might actually be enough.
It's got the menu and the tools have been tested to work. The rest of the workflow is in the configs.

BigRoy:
Without `ls()` being implemented it could never list any loaded content, right? Which needs to be implemented in core.

Looks like a fine draft start - yet it'd need slightly more for all the tools to work. Fine to merge for me as long as we remember it need continued development for production.

tokejepsen:
I don't have time unfortunately to continue with this development until 6 months from now.

I have added a commit that raises errors when trying to use ```containerize()``` and ```ls()``` so it should be obvious to anyone picking up the gauntlet.

tokejepsen:
Closed in favour of #370

PR #330, opened 2018-08-07
Display project tasks when no asset tasks are found.

tokejepsen:
**Problem**

When browsing assets without tasks, no tasks are displayed. This is different from the Launcher, where it fetches the project tasks if no assets tasks are found.

**Solution**

Use project tasks in the Tasks model when no asset tasks are found.

This fixes both the Project Manager and Context Manager.

### DIFF
--- a/avalon/tools/projectmanager/model.py
+++ b/avalon/tools/projectmanager/model.py
@@ -230,7 +230,7 @@ class TasksModel(TreeModel):
 
         # If no asset tasks are defined, use the project tasks.
         if assets and not tasks:
-            project = io.parenthood(assets[0])[0]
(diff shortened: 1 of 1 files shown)


tokejepsen:
I tried using "io.find_one" but the environment doesn't seem to be set, so "parenthood" was the only other way I figure out how to get the project.

BigRoy:
Didn't `io.find_one ({"type": "project"})` work?

Anyhow, I am not saying current code is wrong. Was just wondering whether there was a reason to it.

tokejepsen:
> Didn't io.find_one ({"type": "project"}) work?

You are right! That does work. Shows how much I know (nothing) :)

PR #336, opened 2018-08-27
 Minor improvement on cbloader & Rename `api.load()` arg `data` to `options`

davidlatwe:
Little changes...

#### Motivation

I use `cbloader` and the `data` of representation object is required but the tool did not pass it.

#### Implementation

Add one line to extract `data` from representation and pass to `api.load`.

Also, I noticed that instead passing the `ObjectId` of representation to `api.load`, directly pass the representation object can reduce one database query, since the function [`get_representation_context`](https://github.com/getavalon/core/blob/b639ba3c00102d076fcab8c89da26cd744b20d09/avalon/pipeline.py#L830-L835) which been used [here](https://github.com/getavalon/core/blob/b639ba3c00102d076fcab8c89da26cd744b20d09/avalon/pipeline.py#L988), is able to accept both form.

### DIFF
--- a/avalon/pipeline.py
+++ b/avalon/pipeline.py
@@ -999,7 +999,7 @@ def load(Loader, representation, namespace=None, name=None, options=None,
     if options is None:
         options = kwargs.get("data", dict())  # "data" for backward compat
 
-    assert isinstance(options, dict), "Data must be a dictionary"
+    assert isinstance(options, dict), "Options must be a dictionary"
 
     # Fallback to subset when name is None
     if name is None:
--- a/avalon/pipeline.py
+++ b/avalon/pipeline.py
@@ -997,7 +997,7 @@ def load(Loader, representation, namespace=None, name=None, options=None,
 
     # Ensure options is a dictionary when no explicit options provided
     if options is None:
-        options = kwargs.get("data", dict())  # "data" for backward compact
+        options = kwargs.get("data", dict())  # "data" for backward compat
 
     assert isinstance(options, dict), "Data must be a dictionary"
 
--- a/avalon/pipeline.py
+++ b/avalon/pipeline.py
@@ -970,7 +970,8 @@ def load(Loader, representation, namespace=None, name=None, data=None):
 
     Args:
         Loader (Loader): The loader class to trigger.
-        representation (str or io.ObjectId): The representation id.
+        representation (str or io.ObjectId or dict): The representation id
+            or full representation as returned by the database.
         namespace (str, Optional): The namespace to assign. Defaults to None.
         name (str, Optional): The name to assign. Defaults to subset name.
         data (dict, Optional): Additional custom data to pass on to the loader.
--- a/avalon/pipeline.py
+++ b/avalon/pipeline.py
@@ -169,14 +169,14 @@ class Loader(list):
         fname = template.format(**data)
         self.fname = fname
 
-    def load(self, context, name=None, namespace=None, data=None):
+    def load(self, context, name=None, namespace=None, options=None):
         """Load asset via database
 
         Arguments:
             context (dict): Full parenthood of representation to load
             name (str, optional): Use pre-defined name
             namespace (str, optional): Use pre-defined namespace
-            data (dict, optional): Additional settings dictionary
+            options (dict, optional): Additional settings dictionary
 
         """
         raise NotImplementedError("Loader.load() must be "
@@ -965,7 +965,8 @@ def _make_backwards_compatible_loader(Loader):
     return type(Loader.__name__, (BackwardsCompatibleLoader, Loader), {})
 
 
-def load(Loader, representation, namespace=None, name=None, data=None):
+def load(Loader, representation, namespace=None, name=None, options=None,
+         **kwargs):
     """Use Loader to load a representation.
 
     Args:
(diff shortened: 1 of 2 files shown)


davidlatwe:
Correct me if I am wrong, in my understanding, `api.load` is *officially* require `representation id` to acquire `context`, so that it can pass `context` to `Loader.load`.

And the one who responsible to acquire `context` from `representation id` is the function `get_representation_context`, which currently work as follow pseudo code:

```python
def get_representation_context(representation):
    if isinstance(representation, (str, ObjectId)):
        representation = io.find_one({"_id": ObjectId(str(representation))})
    
    # acquire context
    ...
    return context

def load(Loader, representation, namespace=None, name=None, data=None):
    ...
    context = get_representation_context(representation)
    ...
```

The argument `representation` is actually able to be accepted as `str` or `ObjectId` or `dict`.

Thus, `api.load` should be able to accept `representation dict` directly as well, and prevent one extra unnecessary database query. 

> The data argument is intended for tool specific input, e.g. `with_selection=True` or `animation=False`. Almost like settings/options.

Sounds like we have different future use case on this ?
Is the `tool` you refer is Colorbleed's in-house tool ?

> but it has no function if it's used for the data that is housed in the database on the representation.

Yes, but I did write something to the `representation["data"]` with my extractor plugins, and those `data` is required for my loaders to process representations.

Thanks for the reply @BigRoy :)
Please let me know what you think

davidlatwe:
Hey, I am back on this :)

But I am still not quite understand on this :
> I wouldn't explicitly pass data from the representation if it's in the representation anyway.

Isn't that right the `data` of the `representation` should be accepted by the Loader by default ?
Since the field `data` is from the Avalon integrator plugin.

Thanks!

mottosso:
I think what Roy is saying is that the `data=` of a representation isn't the same as the `data` passed here. Maybe we should rename this `data=` argument to avoid confusion in the future. That argument is for tool options. Maybe `options=` or `opts=`?

The reason it isn't passed at the moment, is because there aren't any options for this particular tool, is that right @BigRoy? Does that make sense @davidlatwe?

BigRoy:
> The reason it isn't passed at the moment, is because there aren't any options for this particular tool, is that right @BigRoy? 

Yes, that is what I meant.

> Thus, api.load should be able to accept representation dict directly as well, and avoid one extra unnecessary database query.

If it technically works it's fine with me - I just recalled that the API wasn't intended to function that way but might have assumed wrong and jumped the gun on the answer.

---

The issue regarding the `data` is that it's indeed not clearly passed from the get go as "data" or "representation" - however the data is present even without explicitly passing it so you don't need to pass it yourself.

You can find it in the `load` function in the `context` argument, specifically:

```python
representation = context["representation"]
data = representation["data"]
```

That's what I meant with the fact that you wouldn't need to pass that explicitly, that particular data is already available to you in the load method. Making this change in code redundant, correct?

Does that help?

davidlatwe:
Ah ah! I see now !

Thanks for your help @mottosso & @BigRoy . :)
Let me try this on my plugins

davidlatwe:
Renaming `data` to `options` seems to be logical, since [`avalon.api.create`](https://github.com/getavalon/core/blob/b639ba3c00102d076fcab8c89da26cd744b20d09/avalon/pipeline.py#L773) has the same param interface, it has both `options` and `data` with different purposes.

```python
"""
Arguments:
        name (str): Name of subset
        asset (str): Name of asset
        family (str): Name of family
        options (dict, optional): Additional options from GUI
        data (dict, optional): Additional data from GUI
"""
```

I did not take too much attention on `Creator Plugin`, knowing that there are `options` param already exists on `api.create`, starting to make more sense now.

davidlatwe:
I tweaked my PR, added argument renaming (`data` -> `options` on `api.load`).
Works on my end, should also maintained backward compatibility.

What do you think ?

avalon/pipeline.py
```diff
@@ -993,11 +995,11 @@ def load(Loader, representation, namespace=None, name=None, data=None):
                                       "{}".format(Loader.__name__,
                                                   context["subset"]["name"]))
 
-    # Ensure data is a dictionary when no explicit data provided
-    if data is None:
-        data = dict()
+    # Ensure options is a dictionary when no explicit options provided
+    if options is None:
+        options = kwargs.get("data", dict())  # "data" for backward compact
```
BigRoy: `backward compact` should be `backward compatibility` or maybe `backward compat` if you really want it shorter? :)

BigRoy:
Sorry, I thought I replied to this. This looks good to me if it works. :) Just commented on one minor typo.

@aardschok what do you think?

---

@davidlatwe Nice work!

aardschok:
I have no objections to this, nice work!

davidlatwe:
Thanks !  I got them fixed :P

davidlatwe:
Sphinx's latest release `1.8.0`  will fail the document build.
Lock version to 1.7.9 for now.

BigRoy:
Are these latest commits supposed to be part of the same PR?

davidlatwe:
Sorry, that should be separate.
I'll reset it and do the right thing.

davidlatwe:
Okay it's done, I'll put that (*pin sphinx==1.7.9*) in next PR alone, sorry for the trouble :cry:

davidlatwe:
Please merge #338 first and I will rebase this PR so the travis will pass :)

mottosso:
Thanks @davidlatwe!

PR #340, opened 2018-09-18
Update Maya fileDialog's start-dir when task changed

davidlatwe:
### Before:
task_before

FileDialog's start-dir will stay at previous opened folder path after task changed.

### After:
task_after

Now it will point to the scenes folder of that task after task changed.
And, this is not recorded in GIF, but it will prompt a confirm dialog to notify artist to use Launcher, due to the workspace of that task is un-initialized (folder not exists).

### DIFF
--- a/avalon/maya/pipeline.py
+++ b/avalon/maya/pipeline.py
@@ -4,7 +4,7 @@ import errno
 import importlib
 import contextlib
 
-from maya import cmds, mel, OpenMaya
+from maya import cmds, OpenMaya
 from pyblish import api as pyblish
 
 from . import lib, compat
@@ -567,13 +567,10 @@ def _on_task_changed(*args):
     workdir = api.Session["AVALON_WORKDIR"]
     if os.path.exists(workdir):
         logger.info("Updating Maya workspace for task change to %s", workdir)
+
         _set_project()
 
-        # Update file dialog starting dir
-        # (NOTE) using `mel.eval` because this `workspace` Python cmd failed
-        #   in Maya 2017 +
-        frule_scene = mel.eval("workspace -q -fileRuleEntry \"scene\"")
-
+        frule_scene = cmds.workspace(fileRuleEntry="scene")
         cmds.optionVar(stringValue=("browserLocationmayaBinaryscene",
                                     workdir + "/" + frule_scene))
 
--- a/avalon/maya/pipeline.py
+++ b/avalon/maya/pipeline.py
@@ -577,15 +577,3 @@ def _on_task_changed(*args):
     else:
         logger.warning("Can't set project for new context because "
                        "path does not exist: %s", workdir)
-
-        asset = api.Session["AVALON_ASSET"]
-        task = api.Session["AVALON_TASK"]
-        cmds.confirmDialog(
-            title="Context Manager - Context Not Exists",
-            icon="warning",
-            messageAlign="center",
-            message=("Can't set project for new context because the workspace "
-                     "of context <{asset} - {task}> not exists yet.\nPlease "
-                     "use Avalon Launcher to enter."
-                     "".format(asset=asset, task=task)),
-        )
--- a/avalon/maya/pipeline.py
+++ b/avalon/maya/pipeline.py
@@ -4,7 +4,7 @@ import errno
 import importlib
 import contextlib
 
-from maya import cmds, OpenMaya
+from maya import cmds, mel, OpenMaya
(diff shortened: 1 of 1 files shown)


davidlatwe:
One thing that I am not sure of is, should we reset back to previous context when Avalon says new context's workspace not exists ?

To me, I think it should, and the workspace check should implement in `avalon.api.update_current_task`, right before emitting `taskChanged`, raise error if not exists, and the prompt dialog should be implement in context manager.

What do you think ?

BigRoy:
> One thing that I am not sure of is, should we reset back to previous context when Avalon says new context's workspace not exists ?

We have on task changed callback that creates the workspace if it does not exist yet (in Colorbleed config). So for us it's fine if it does not exist yet. :) Does that help?

Could port that somehow to core? (Thinking of it... that might even be in our core repo, but I am not sure)

davidlatwe:
Oh I see~
I found the implementation of workspace initialization in Colorbleed config, you use `avalon.Application` to initialize it, which making more sense ! Nice :P

Then I think I will remove the prompt dialog in this PR.

> Could port that somehow to core?

I think we can ship that logic to core in next PR, take Maya for example:

```python
# avalon/pipeline.py
...

def on_task_changed(*args):
    """Wrapped function of app initialize and maya's on task changed
    (copied from Colorbleed config)
    """

    # Inputs (from the switched session and running app)
    session = avalon.Session.copy()
    app_name = os.environ["AVALON_APP_NAME"]

    # Find the application definition
    app_definition = pipeline.lib.get_application(app_name)

    App = type("app_%s" % app_name,
               (avalon.Application,),
               {"config": app_definition.copy()})

    # Initialize within the new session's environment
    app = App()
    env = app.environ(session)
    app.initialize(env)
```

```python
# avalon/maya/pipeline.py
from .. import pipeline
...

def _on_task_changed(*args):
    pipeline.on_task_changed()

    # Do something app specific action..
    _update_menu_task_label()
    workdir = api.Session["AVALON_WORKDIR"]
    if os.path.exists(workdir):
        _set_project()
        ...
```

Then `avalon.maya.install()` will register this chained callback for us !

>  (Thinking of it... that might even be in our core repo, but I am not sure)

I'll go there and sniff around for good stuff :D

avalon/maya/pipeline.py
```diff
@@ -568,6 +568,27 @@ def _on_task_changed(*args):
     if os.path.exists(workdir):
         logger.info("Updating Maya workspace for task change to %s", workdir)
         _set_project()
+
+        # Update file dialog starting dir
+        # (NOTE) using `mel.eval` because this `workspace` Python cmd failed
+        #   in Maya 2017 +
+        frule_scene = mel.eval("workspace -q -fileRuleEntry \"scene\"")
```
BigRoy: I think in Python it somehow is:

```python
import maya.cmds as cmds
cmds.workspace(fileRuleEntry="scene")
```

Thus without the query. I think it was different in older versions of Maya where it would actually need the query flag, but haven't tested it.

Thanks for the comment here.

BigRoy:
> I'll go there and sniff around for good stuff :D

Sure do! ;) 

This Maya browser fix is already really good - nice work!

avalon/maya/pipeline.py
davidlatwe: So this is the command that fit to all version...
I gave it a test run, and [all passed](https://travis-ci.org/davidlatwe/maya-test/builds/429989661)
I will change to this cmd, much simpler.

davidlatwe:
Thanks @BigRoy :)
I'll leave it as is and merge in next couple day if no further objections.

BigRoy:
The warning message might still be conflicting with our "solution" to ensuring the workspace actually exists in our config. Because I think ours will do it directly after this callback - so the user gets warned for no apparent reason?

Can we avoid the pop-up but just have it log a sensical message?

davidlatwe:
> Because I think ours will do it directly after this callback

On the contrary, I believe it's initializing workspace before this callback, cause I just *stealing* your "solutions". :P

But since you have mentioned, I will remove that for good, I now aware it's a bit redundant for the future development heading.

davidlatwe:
Done. I leave the logger warning massage as is, should be fine right ?

BigRoy:
Sure is! Thanks again. Feel free to merge!

davidlatwe:
Merging this. Thanks @BigRoy  :)

PR #349, opened 2018-11-02
Fix #325

### LINKED ISSUE #325: Save project with visualparent
**Problem**

When having visualparent on assets the saving process errors.

State of inventory when loading project:
```toml
[[shots]]
label = "ep0010"
group = "Episode"
name = "ep0010"

[[shots]]
label = "sq0010"
group = "Sequence"
visualParent = 5b3dd2c0cd5f4c097c405697
name = "ep0010_sq0010"

[[shots]]
label = "sh0010"
group = "Shot"
visualParent = 5b3dd2c0cd5f4c097c405699
name = "ep0010_sq0010_sh0010"
```

Error message when ```python -m avalon.inventory --save```:
```
invalid literal for int() with base 10: '5b3dd2c0cd5f4c097c405697'
```

davidlatwe:
### Concept
Avoid `visualParent` get loaded to `.inventory.toml`, hide from user editing.

### Implementation
Pop-out the key `"visualParent"` from the asset data of the project document, once it loaded from database and before saving as `.toml`.

### DIFF
--- a/avalon/inventory.py
+++ b/avalon/inventory.py
@@ -224,6 +224,8 @@ def load(name):
             silo = asset["silo"]
             data = asset["data"]
 
+            data.pop("visualParent", None)  # Hide from manual editing
+
             if silo not in inventory:
                 inventory[silo] = list()
 


mottosso:
It looks like this should do the trick. By not being in the saved inventory, it won't get uploaded, and won't have any effect on the existing variable.

If so, I'm happy with this! Welcome to merge!

davidlatwe:
> It looks like this should do the trick. By not being in the saved inventory, it won't get uploaded, and won't have any effect on the existing variable.

No, it won't get uploaded nor effecting any existing variable !
Tested on my end :D

Will merge if no one has any objection on hiding `visualParent` this way :)

davidlatwe:
> Or maybe that is not the core repositorys responsibility.

Yeah, the right place to really *fix* the issue should be at `avalon.tools.projectmanager` *(still being the core repositorys responsibility, haha)*.

But it would require the asset document to move the field `"visualParent"` outside the field `"data"`, to avoid being loaded to `.toml` file. Perhaps would need a new asset schema `avalon-core:asset-3.0` ? Also needs to handle the backward compatibility.

This PR is more like a workaround, yes. For saving some troubles for the time being :)

mottosso:
> This PR is more like a workaround, yes. For saving some troubles for the time being :)

No, I disagree.

The `.toml` files are a (limited) user interface to the assets in the database, and it doesn't do a good enough job at representing these IDs. But there's no reason it needs to represent that. At the moment, it's pulling everything out of `.data{}` but doesn't really have to.

Rather than *excluding* members (which admittedly is a little hacky), maybe we could turn things around and *include* members instead. For members it's able to actually represent, like start/end frame, group and resolution. Then it'd be like any other GUI, where the GUI is the text editor. Then we could document those members, and that'd be it for `.toml` files; a basic interface where more complex functionality is left to other GUIs, like the Project Manager and Launcher.

davidlatwe:
Vulcan logic !

davidlatwe:
> Rather than excluding members (which admittedly is a little hacky), maybe we could turn things around and include members instead.

But if using include method, we couldn't adding more attribute to asset data freely, are we ? Unless we define a white list else where, in each studio's config maybe?

mottosso:
> But if using include method, we couldn't adding more attribute to asset data freely, could we ?

Exactly, it'd be limited to the exact things we'd specify. I figure, for more complex use, would it be preferable to edit that via a GUI?

I'm curious, what do you guys use the .toml files for? They were initially created as a stepping stone to a GUI, and now that we've got a GUI (Project Manager), are our efforts better spent there?

davidlatwe:
> what do you guys use the .toml files for?

I put some project-wide attributes like `fps`, `resolution`, `edit_in`, `edit_out`, and some deadline settings.
As for assets, there are `task`, and each shots' `edit_in`, `edit_out`.

mottosso:
Ah, and at the moment, there isn't any other way of editing that kind of data? Seems to be well suited for the Project Manager GUI.

davidlatwe:
Yeah, I remember @BigRoy mentioned that this was in thier internal roadmap, watting for implement, in gitter a while back.

davidlatwe:
Found it :P *(searching in Gitter was a bit hard)*
:point_up: [July 26, 2018 4:21 PM](https://gitter.im/getavalon/Lobby?at=5b598483db4bcb16e87a4ac5)

mottosso:
@davidlatwe Is this ready to merge?

davidlatwe:
For fixing #325, yes, this one is ready !

PR #353, opened 2018-11-10
Implement Hierarchy View in Scene Inventory App

davidlatwe:
This enables a better view for hierarchical (nested) subset type like `setdress` in Scene Inventory App.
Able to switch between normal *flat* view and hierarchy view.

hierarchy_view

### DIFF
--- a/avalon/maya/pipeline.py
+++ b/avalon/maya/pipeline.py
@@ -447,6 +447,14 @@ def ls():
 
     for container in sorted(containers):
         data = parse_container(container)
+
+        # Hierarchical container support
+        parent = cmds.listSets(object=container) or []
+        for node in parent:
+            if node in containers:
+                data["parent"] = node
+                break
+
         yield data
 
 
--- a/avalon/tools/cbsceneinventory/app.py
+++ b/avalon/tools/cbsceneinventory/app.py
@@ -475,6 +475,14 @@ class Window(QtWidgets.QDialog):
         control_layout.addWidget(text_filter)
         control_layout.addWidget(outdated_only)
         control_layout.addWidget(refresh_button)
+
+        option_layout = QtWidgets.QHBoxLayout()
+        hierarchy_view = QtWidgets.QCheckBox("Hierarchy View")
+        hierarchy_view.setToolTip("Display subsets hierarchy.")
+        hierarchy_view.setChecked(False)
+
+        option_layout.addWidget(hierarchy_view)
+
         # endregion control
 
         model = InventoryModel()
@@ -488,10 +496,12 @@ class Window(QtWidgets.QDialog):
         view.setItemDelegateForColumn(column, version_delegate)
 
         layout.addLayout(control_layout)
+        layout.addLayout(option_layout)
         layout.addWidget(view)
 
         self.filter = text_filter
         self.outdated_only = outdated_only
+        self.hierarchy_view = hierarchy_view
         self.view = view
         self.refresh_button = refresh_button
         self.model = model
@@ -500,6 +510,7 @@ class Window(QtWidgets.QDialog):
         # signals
         text_filter.textChanged.connect(self.proxy.setFilterRegExp)
         outdated_only.stateChanged.connect(self.proxy.set_filter_outdated)
+        hierarchy_view.stateChanged.connect(self.model.set_hierarchy_view)
         refresh_button.clicked.connect(self.refresh)
         view.data_changed.connect(self.refresh)
 
--- a/avalon/tools/cbsceneinventory/model.py
+++ b/avalon/tools/cbsceneinventory/model.py
@@ -58,7 +58,7 @@ class InventoryModel(TreeModel):
                 if node.get("isGroupNode"):  # group-item
                     return qta.icon("fa.folder", color=color)
                 else:
-                    return qta.icon("fa.folder-o", color=color)
+                    return qta.icon("fa.file-o", color=color)
 
             if index.column() == 3:
                 # Family icon
--- a/avalon/tools/cbsceneinventory/model.py
+++ b/avalon/tools/cbsceneinventory/model.py
@@ -55,7 +55,7 @@ class InventoryModel(TreeModel):
             if index.column() == 0:
                 # Override color
                 color = node.get("color", style.colors.default)
-                if not index.parent().isValid():  # group-item
+                if node.get("isGroupNode"):  # group-item
                     return qta.icon("fa.folder", color=color)
                 else:
                     return qta.icon("fa.folder-o", color=color)
@@ -202,6 +202,7 @@ class InventoryModel(TreeModel):
             group_node["family"] = family
             group_node["familyIcon"] = family_icon
             group_node["count"] = len(group_items)
+            group_node["isGroupNode"] = True
 
             self.add_child(group_node, parent=parent)
 
--- a/avalon/tools/cbsceneinventory/model.py
+++ b/avalon/tools/cbsceneinventory/model.py
@@ -20,6 +20,10 @@ class InventoryModel(TreeModel):
 
     UniqueRole = QtCore.Qt.UserRole + 2     # unique label role
 
+    def __init__(self, parent=None):
+        super(InventoryModel, self).__init__(parent)
+        self._hierarchy_view = False
+
     def data(self, index, role):
 
         if not index.isValid():
@@ -65,11 +69,23 @@ class InventoryModel(TreeModel):
 
         return super(InventoryModel, self).data(index, role)
 
+    def set_hierarchy_view(self, state):
+        """Set whether to display subsets in hierarchy view."""
+        state = bool(state)
+
+        if state != self._hierarchy_view:
+            self._hierarchy_view = state
+            self.refresh()
+
     def refresh(self):
         """Refresh the model"""
 
         host = api.registered_host()
         config = api.registered_config()
+
+        # host should have `pipeline`
+        if hasattr(host.pipeline, "find_host_config"):
+            config = host.pipeline.find_host_config(config)
 
         items = []
         containers = host.ls()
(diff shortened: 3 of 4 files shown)


avalon/maya/pipeline.py
```diff
@@ -447,6 +447,14 @@ def ls():
 
     for container in sorted(containers):
         data = parse_container(container)
+
+        # Hierarchical container support
+        parent = cmds.listSets(object=container) or []
+        for node in parent:
+            if node in containers:
+                data["parent"] = node
+                break
```
davidlatwe: I end up moving this logic from `config.collect_container_metadata` to here, since I thought the *hierarchy view* feature would apply to everyone, so the attribute `parent` which is required should also inside the `core`.

davidlatwe:
About this change on Scene Inventory, next I am going to do is implementing dependency view, for asset type like `lookDev` or `animCurve`.

But since the Scene Inventory is *software-agnostic*, those extra checkbox might not suitable for other DCC apps, perhaps these kind of change should be refactored at some point, like moving DCC specific code to `avalon.host`, and let the GUI load them on startup.

For now, I'll continue implement the feature for Maya and push them here, then see what need to be changed or refactored when the time comes.

Wondering what you think ? Any inputs are welcome :)

davidlatwe:
Thanks @tokejepsen  :D

Here's some screen shot and mockup on a scene which have a `setdress` type asset loaded:

* This is what we currently have now, every thing is flat
    <details>
    <summary>PIC</summary>

    sceneinventory_flat

    </details>

* Hierarchy view, just like gif above

    Since `lookDev` does not have the hierarchy, they stay on the ground.

    <details>
    <summary>PIC</summary>

    sceneinventory_hierarchy

    </details>

* Dependency view (Mockup)

    In dependency view, `lookDev` is nested under each `model` assets.

    sceneinventory_dependency

### Visualize relationships

So what I am trying to do is making Scene Inventory to be able to visualize some relationships between containers, since in *Maya*, depend on `subset` type, some of the container may have hierarchical relation, and some of them were in dependency.

davidlatwe:
Here's some latest progress on hierarchy view:
* Implemented version and text filtering.
* Implemented version outdated and child item outdated highlight.
* Able to properly passing containers to actions.

And here's some improvement on both view mode:
* Renamed column "name" to "Name" so that able to display namespace on GUI without losing container data.
* Able to select GUI items by actions.
* Moved action sub-menu to the top.

I think the most interesting function is able to select view items by action, the motivation I implementing this was, I was trying to find a way to quickly know which asset assigned that *look* in Scene Inventory, and the implementation of *Dependency view* I mentioned before was making things to complex, so I thought why not use `InventoryAction` to tell me and there it is. Hope you like this.

hierarchy_view_demo

In above GIF, the action "Lucky Pick" was randomly select one item in the GUI from the grouped containers.

tokejepsen:
Sorry, I'm not actually very useful when it comes to feedback on the Loader since I've yet to try it.

Unrelated to the review, what are those red boxes in your outliner?

BigRoy:
> Oh, it's the referenced subset's group node, nothing fancy except with a custom icon for artist to identify them. :)

Clever!

---

Have been unable to check the code yet though but visually it looks like there's quite some stuff here that could be interesting.

avalon/maya/pipeline.py
BigRoy: Sure. The only thing I am a bit wary about is whether being present in the container is the functionality that will be the same basis for parenting for "everyone" ever. However, I do feel this is a good basis. Plus a bit about the speed of things too - since it involves a bit more queries.

Could we maybe optimize this query a bit? Should `containers` maybe be a `set` so the lookup becomes `O(1)` instead of `O(n)`.

How does this differ in speed for big scenes?

mottosso:
Just want to say, this is stellar work @davidlatwe! I wish I had this in the pipeline I'm sitting in at the moment (shotgun toolkit, yuk!)

davidlatwe:
I have changed the implementation of *Hierarchy view* from affecting all containers with a checkbox, to able to cherry-pick with right-click action.

cherry-pick_demo

You can keep cherry-picking as long as there are item to select, and maintain the selection.
If the host does not have the concept of hierarchical containers, still able to use as cherry-pick only.

avalon/maya/pipeline.py
davidlatwe: Since I have changed to use different implementation, there should not have any over head on normal `ls()` query. :)

avalon/tools/cbsceneinventory/model.py
```diff
@@ -65,20 +95,99 @@ def data(self, index, role):
...
+                for name in [n for n in names if n not in selected_items]:
+                    selected_items.add(name)
+                    item = items_by_name[name]
+                    yield item
+
+                    for child in walk_children(item["children"]):
+                        yield child
+
+            items = list(walk_children(selected))  # Cherry-picked and extended
+
```
davidlatwe: Here's the heavy lifting part, since the hierarchy has been cherry-picked, in order to draw the graph, we need to scan all the containers, to see which is the final hierarchy and which is not.

This is the part that we need `"children"` data from host, for filtering.

And the `"parent"` data is required down below, for building the view.

avalon/tools/cbsceneinventory/app.py
```diff
@@ -527,15 +643,19 @@ def refresh(self):
         with preserve_expanded_rows(tree_view=self.view,
                                     role=self.model.UniqueRole):
             with preserve_selection(tree_view=self.view,
-                                    role=self.model.UniqueRole):
-                self.model.refresh()
+                                    role=self.model.UniqueRole,
+                                    current_index=False):
```
davidlatwe: I changed the `current_index` to `False`, this is for preserving the selection of multiple items.

davidlatwe:
Thanks @mottosso  :D, new year is coming, my new year gift from asset artists is counting on this ! (JK)

I have edited the previous comment to point out one minor tweak.

Finally, I hope the combing of cherry-pick and hierarchical view make sense to you guys, I personally pretty happy about this re-implement, since there's only a few types of subset have nested containers, and cherry-picking things could help on managing big scene I believe. And the cherry-pick works for all hosts, which I think it's really nice. :)

avalon/maya/pipeline.py
davidlatwe: I have changed to use `set` in some place in this implementation :)

davidlatwe:
Here's final sums up about this PR:

cbinventorycbinherry-pick_demo

### Major New Features

* Cherry-Pickable View with hierarchy visualization (Require `host` to support hierarchy)
* Custom inventory action able to select back if returning a  list (, set or dict) of container name.
* Enable custom inventory action to work without item selected. (dce0b51, 3c69e29)

### Minor Tweaks

* Changed group node name format from `"%s_%s: (%s)"` to `"%s . %s: (%s)"` (8135f92)
* Moved custom actions sub-menu to the top of right click menu for faster access. (3a55f62)
* Child item name and version color is dimmed. (7fbbc24, 0332c45)
* Changed column `name` to `Name`, remapped the container `"namespace"` to item `"Name"`. (d5b5ab8)
* Right click menu will not show all the actions if there's no highlighted (selected) items. (50037e5)
* When entering cherry-pick view, the widget border color will change to yellow as a hint.

BigRoy:
Great work @davidlatwe!

Some notes from my end on that last summary:

> Cherry-Pickable View with hierarchy visualization (Require host to support hierarchy)

So the Cherry-Pick is like an "isolate view", right? It seems nice. 

Would it be possible to somehow support the cherry pick being enabled by a specific search query in the filter? Like it would push in "pick:asset-x" in the Search bar so that the proxy automatically updates.

Did you Implement the Cherry Picking through a Proxy Model? (didn't check the code yet)

> Moved custom actions sub-menu to the top of right click menu for faster access.

I wouldn't be a fan of this actually. I'd still prioritize the other actions as the ones the Artist would most likely end up using.

> Child item name and version color is dimmed.

In the gif to me the children seem "greyed out" and I'm seeing them as "unloaded". Maybe it's just something to get used too - just wanted to let you know how it feels by just looking at the gif. It's a nice helper, but maybe it's slightly dimmed too far?

davidlatwe:
> So the Cherry-Pick is like an "isolate view", right? It seems nice.

Oh yes ! That's the word I should choose it at first !
I'd like to change the action label from `Cherry-Pick (Hierarchy)` to `Isolate (Hierarchy)`, it's a much common term for artist, right ?

> * Would it be possible to somehow support the cherry pick being enabled by a specific search query in the filter?
> * Did you Implement the Cherry Picking through a Proxy Model?

No, not by Proxy Model, it's by passing a set of `objectName` to the TreeModel's `refresh`, then filtering against all the containers collected from `host.ls()`.
Something like this:
```python
# app.py
class View(QTreeView):
    def enter_hierarchy(self, items):
        # Action to enter isolate view
        self._selected = set(i["objectName"] for i in items)
        self._hierarchy_view = True  # Will make `refresh` to take `selected` items
        self.hierarchy_view.emit(True)
        self.data_changed.emit()  # Trigger the `refresh`

# model.py
class InventoryModel(TreeModel):
    def refresh(self, selected=None):
        items = host.ls()
        if self._hierarchy_view and selected:
            # Filtering & child container collecting
            ...
            # Build Tree
            while items:
                ...
                    self.add_items(children, parent)
                ...
        else:
            # Build Tree
            self.add_items(items)
```

So no, I don't think search query can support this. Since it require to rebuild the model to visualize the hierarchy. :P

> I'd still prioritize the other actions as the ones the Artist would most likely end up using.

Yeah, it's true that the other default actions indeed use more often, but what I think was the custom action might be used repeatedly, depend on the function of that action, if there's some action is for selecting or diagnose-like function, then it would be repeatedly accessed.
Hmmm, having that said, or maybe I shouldn't put such action in there at first place ?

> maybe it's slightly dimmed too far?

Yeah, a bit too far I agreed, perhaps I'll try increasing the row spacing, instead of dimming the color.

Thanks for the feedback @BigRoy !!

davidlatwe:
> * I'd still prioritize the other actions as the ones the Artist would most likely end up using.

Hmmm, I'll revert it back to the original position, since I think the user experience is much more at Colorbleed of course. I shall make it stay that way  :)

davidlatwe:
I think I have a good start on implementing different view via search query, kind of fun.

Here's a rough demo.

inventory_search

With `$`, will display hierarchy view, and adding one digit will expand to that depth.
With `#` will auto expand normal view.
With `@` will enter headless (group-less) view.

I think we could define some simple query expression, like `$<depth> <asset1> <asset2> : <subset>` some sort.

What do you think ?

mottosso:
Would it be possible to keep PRs as small and contained as possible, before introducing more features?

I think the search is good, but am a little concerned it doesn't handle a busy network or server very well. A Proxy model would contain processing to the local machine, and there should probably also be some delay between typing the search query and actually performing the search, so that when a network takes 500 ms to reply typing doesn't appear sluggish.

davidlatwe:
> Would it be possible to keep PRs as small and contained as possible, before introducing more features?

Ah yes, the latest commit should be a good stop point. Thanks for the heads up. :)

> there should probably also be some delay between typing the search query and actually performing the search

About that, till the latest commit, it could happen when preforming cherry-pick action, since it will `refresh` the model, which will have database access.

As for the search query implementation above, the data model get built for all view modes only once at startup until next refresh or data change, when changing view with typing `$` `@` or `#`, there only preformed source model shifting.

But in both cases, when typing the keyword to filter out unwanted assets, it's handled in local pre-built model (filter proxy model), so won't be any latency. :P

tokejepsen:
The search query looks cool!

Is it based on any existing query language?

davidlatwe:
update

Final updates:
* Moved sub-menu back to bottom
* Brighten the child item color

If no other objections, I'll merge this next week (probably Tuesday) :)

davidlatwe:
About to merging this :)

PR #356, opened 2018-12-18
Add property 'families' into config schema 1.0

davidlatwe:
This PR will make the inventory able to save the *family configurations* via `$ avalon --save`,
which is required for `cbloader`.

https://github.com/getavalon/core/blob/c5dda32149e636a32d82ac06c09e279499a76402/avalon/tools/cbloader/lib.py#L14-L27

The property `families` won't be a *required* attribute, so instead of creating a new version of config schema, I chose to update current version.


BigRoy:
Just want to say, this looks perfect! 👍 Great work.

davidlatwe:
Merging this ! Thanks :)

PR #357, opened 2018-12-19
Handling database auto re-connection

davidlatwe:
This will handling MongoDB auto re-connection automatically ( 3 retry times ), especially useful if you work with MongoDB Replica Set ( Auto re-connect if the primary node went down ).

( I have changed to use Replica Set recently. )

### DIFF
--- a/avalon/io.py
+++ b/avalon/io.py
@@ -220,6 +220,24 @@ def requires_install(f):
     return decorated
 
 
+def auto_reconnect(f):
+    """Handling auto reconnect in 3 retry times"""
+    @functools.wraps(f)
+    def decorated(*args, **kwargs):
+        error = None
+        for retry in range(3):
+            try:
+                return f(*args, **kwargs)
+            except pymongo.errors.AutoReconnect as e:
+                error = e
+                log.error("Reconnecting..")
+                time.sleep(0.1)
+        else:
+            raise error
+
+    return decorated
+
+
 @requires_install
 def active_project():
     """Return the name of the active project"""
@@ -239,15 +257,21 @@ def projects():
         list of project documents
 
     """
+    @auto_reconnect
+    def find_project(project):
+        return self._database[project].find_one({"type": "project"})
 
-    for project in self._database.collection_names():
+    @auto_reconnect
+    def collections():
+        return self._database.collection_names()
+    collection_names = collections()
+
+    for project in collection_names:
         if project in ("system.indexes",):
             continue
 
         # Each collection will have exactly one project document
-        document = self._database[project].find_one({
-            "type": "project"
-        })
+        document = find_project(project)
 
         if document is not None:
             yield document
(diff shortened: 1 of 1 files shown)


BigRoy:
When exactly do you need this? When have you been facing the issue of disconnection? Somehow I've never seen an issue like that and I am wondering why this is a required workaround. Could you elaborate more on the case(s) where this is helpful?

BigRoy:
Ah, that makes sense. Though trying three reconnections in this brief time, is that sufficient? Taking a quick glimpse at Replica Sets in Mongo it seemst that the average election for a new primary MongoDB instance being completed takes around 10 seconds?

And to be sure, this change isn't tackling an issue you're currently facing but is a safety measure for a potential hazard that could come up in production. You're not facing any random connection losses?

I don't see this influencing the default behavior much, so it looks good to me.

avalon/io.py
```diff
@@ -220,6 +220,24 @@ def decorated(*args, **kwargs):
...
+def auto_reconnect(f):
+    """Handling auto reconnect in 3 retry times"""
+    @functools.wraps(f)
+    def decorated(*args, **kwargs):
+        error = None
+        for retry in range(3):
+            try:
+                return f(*args, **kwargs)
+            except pymongo.errors.AutoReconnect as e:
+                error = e
```
BigRoy: Do we need this `error = e`, couldn't you just ignore storing the error as variable and just in in the else statement `raise` without a variable to raise the last error that happened? Or am I misinterpreting what this is trying to do?

davidlatwe:
> Though trying three reconnections in this brief time, is that sufficient?

I think the re-connection is actually happening in the next `pymongo` operation, not so sure about the under laying mechanic, but I did simulate the situation by manually block the connection of the primary node and the code worked. :D

> is a safety measure for a potential hazard that could come up in production.  You're not facing any random connection losses?

No, no connection issue here, it's for the safety, yes :)

Thanks @BigRoy !

avalon/io.py
davidlatwe: Oh ! yes, you are right. It's redundant !

BigRoy:
> I think the re-connection is actually happening in the next pymongo operation, not so sure about the under laying mechanic, but I did simulate the situation by manually block the connection of the primary node and the code worked. :D

Did you try the same with a "write" operation? See: https://docs.mongodb.com/manual/replication/#automatic-failover - it states:

> The replica set cannot process write operations until the election completes successfully. The replica set can continue to serve read queries if such queries are configured to run on secondaries while the primary is offline.

So it's likely that read operations are easier to pass, but the write might still fail.
A bit unrelated to the code itself though. :)

davidlatwe:
> Did you try the same with a "write" operation?

Yap, I did try with `insert_one` ;) , and it did wait a bit longer ( must be waiting the election ! ) :P

davidlatwe:
Merging this !!

PR #359, opened 2018-12-21
Fix #358

### LINKED ISSUE #358: Apps launched with launcher don't format {AVALON_PROJECT} to the Session it launches in.
### Issue

I noticed some interesting behavior when a `.toml` for an application had something like:

```toml
[environment]
MY_CUSTOM_VARIABLE = {AVALON_PROJECT}
```

The `AVALON_PROJECT` value would become the value `placeholder`, not the actual project name that it launched in. This was due to how the applications are retrieved when going into a project, specifically in [control.py it uses its lib.get_apps()](https://github.com/getavalon/launcher/blob/aca32d6647cc755bc4cb0bc146ab3cd81ef4e4a9/launcher/control.py#L283) that triggers [avalon.lib.get_application](https://github.com/getavalon/launcher/blob/7053a6b2a8468662d81fbb66953bc5fc53c34354/launcher/lib.py#L65) to build the application from the `.toml` on disk.

And the [avalon.lib.get_application() directly embeds the "current environment"](https://github.com/getavalon/core/blob/d9ac16f32d353910c6a2ef51715992d27df0ba05/avalon/lib.py#L173) into the environment variables of that application definition.

### Solution

The application environment should only get injected upon launch (when it gets triggered) as opposed to when it gets loaded as it currently does. It should at the very least include the Session for the launch correctly so `placeholder` is not set like this but the actual project's name.

1. My proposal would be to change: `avalon.lib.get_application` to only read out the toml, but *not* perform the environment management and leave that up to the action. We will need to read the `.toml` at least before "running" it to load the icon and color to be displayed in the Launcher.

BigRoy:
This avoids `get_application` directly merging the environment but solely parses the `.toml` for #358. Whether this is the best solution I'm not too sure, but in my case this is exactly functional for our produtions and fixes the issue.

I've implemented in the way I described in the issue:

> My proposal would be to change: avalon.lib.get_application to only read out the toml, but not perform the environment management and leave that up to the action. We will need to read the .toml at least before "running" it to load the icon and color to be displayed in the Launcher

So visually it functions the same, it just now merges the environment at the end upon the call to `environ()` of the `Application` class.

### DIFF
--- a/avalon/lib.py
+++ b/avalon/lib.py
@@ -141,8 +141,16 @@ def which_app(app):
     return None
 
 
-def get_application(name, environment=None):
-    environment = environment or os.environ
+def get_application(name):
+    """Find the application .toml and parse it.
+
+    Arguments:
+        name (str): The name of the application to search.
+
+    Returns:
+        dict: The parsed application from the .toml settings.
+
+    """
     application_definition = which_app(name)
 
     if application_definition is None:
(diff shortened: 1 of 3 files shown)


BigRoy:
Ah, thanks. Well spotted - will look into that and see what it is for. :)

BigRoy:
Thanks, will still fix that thing @davidlatwe mentioned and then get ready to merge this.

BigRoy:
I changed the call in `session.py` - I believe currently that module doesn't do anything and at the time was a draft implemented by @mottosso as an idea of turning the current session into a class. Yet it was never thoroughly tested nor used I think.

Maybe we should drop it at some point (to remove the dead code) or implement it if it has a purpose in the foreseeable future.

PR #360, opened 2019-01-08
Fix app copy fail when path contain env var

davidlatwe:
### Problem

The files in `copy` section of app `.toml` did not get copied due to the file path contain envvar did not expand.
e.g. `"{AVALON_CORE}/res/workspace.mel"`

The app still be able to launch from Launcher even the copy fails, but will get error when you switching context in app.

### Fix

This PR ensure the paths from `config["copy"]` get expanded before doing `shutil.copy`.


BigRoy:
Thanks! Somehow I didn't catch this at all. This looks like a good and easy fix.

davidlatwe:
Merging this ! :D

PR #370, opened 2019-01-23
nuke workflow / pypeclub update

mkolar:
So we took #329 and pushed it a bit further, so this should have most that's needed for nuke implementation. We've closed our previous PR and reopening here as we had to do a lot of cleanup on our side. This builds on top of #329 so if everyone is fine with this PR, the old can be closed I think. 

**What's here:**
Avalon Menu in nuke with all the main options
integrated key functions in ./avalon/nuke/pipeline ( ls, publish, containerise, parse_container, update_container, viewer_update_and_undo_stop)
added some useful functions into ./avalon/nuke/lib (add_publish_knob, ls_img_sequence, maintained_selection)
workfiles integration for Nuke

We tried to keep the integration as close to fusion and maya standard as we could. If anyone spots inconsistencies please shout out, we'll gladly fix them. 

**Creator**
Attaches `avalon` tab to a node with `avalon` data string knob to which we serialize all the necessary data and then read it back during collection from this one attribute. we discussed instead adding more separate attributes like in maya, but this seemed a bit cleaner and user can't screw with the data on the node.

**Containers**
Work the same as above. When loading we add `avalon` data to a new tab on the node, so everything that has this tab with avalon data and correct id is listed as container and can be updated. 

I'm hoping there are no outside dependencies left at all now.


tokejepsen:
Awesome stuff! Unfortunately can't test it, but will close the other issue to clean up.

avalon/nuke/lib.py
```diff
@@ -0,0 +1,240 @@
...
+    try:
+        yield
+    finally:
+        # unselect all selection in case there is some
+        [n['selected'].setValue(False) for n in nodes]
+        # and select all previously selected nodes
+        if previous_selection:
+            [n['selected'].setValue(True)
+             for n in nodes
+             if n.name() in previous_selection]
```
BigRoy: Couldn't this just be `for n in previous_selection` if you don't store the `name` but just keep a reference to the node?

avalon/nuke/lib.py
```diff
@@ -0,0 +1,240 @@
...
+def maintained_selection():
+    nodes = nuke.allNodes()
+    previous_selection = [n.name()
+                          for n in nodes
+                          if n['selected'].value() is True]
+    try:
+        yield
+    finally:
+        # unselect all selection in case there is some
+        [n['selected'].setValue(False) for n in nodes]
```
BigRoy: This would act "wrong" in the situation where during the context new nodes are created which weren't originally there. These would never get deselected since they are not in `nodes`. This should actually retrieve the current selection, and clear that?

avalon/nuke/lib.py
```diff
@@ -0,0 +1,240 @@
...
+import logging
+import toml
+
+log = logging.getLogger(__name__)
+
+
+@contextlib.contextmanager
+def maintained_selection():
+    nodes = nuke.allNodes()
+    previous_selection = [n.name()
```
BigRoy: Can't you get the selected nodes in nuke with: `nuke.selectedNodes()`? I see it being used [here](https://stackoverflow.com/a/35048809/1838864).

avalon/nuke/lib.py
```diff
@@ -0,0 +1,240 @@
...
+    Returns:
+        tuple: head, padding, tail (extension)
+
+    """
+    filename, ext = os.path.splitext(path)
+    print(filename)
+    print(filename, ext)
+    # Find a final number group
+    if '%' in filename:
+        print("I am here")
```
BigRoy: These print statements likely need to be removed, looks like debugging lines.

avalon/nuke/lib.py
```diff
@@ -0,0 +1,240 @@
...
+        node = node[0]
+
+    add_avalon_tab_knob(node)
+    add_publish_knob(node)
+    node['avalon'].setValue(toml.dumps(data))
+
+    return node
+
+
+def ls_img_sequence(dirPath, one=None):
```
BigRoy: Just wondering, could this be simplified/resolved by using `clique` that is vendorized with avalon core?

Also, if there's a "list image sequence" method and it's non-trivial code (like this) I believe it is somewhat universally needed - could be a good one for `avalon.lib` as opposed to `avalon.nuke.lib`? Or is this specifically targeted at Nuke?

BigRoy:
Great stuff @mkolar! I commented one some things that stood out to me, though note that I am not a Nuke user.

mkolar:
> I commented one some things that stood out to me

awesome thanks. I'm pinging @jezscha who did most of it about those points. But all of them sound like very good catches and should be fixed I think.

avalon/nuke/lib.py
mkolar: Very true

avalon/nuke/lib.py
mkolar: absolutely

avalon/nuke/lib.py
mkolar: Yes!

avalon/nuke/lib.py
mkolar: Actually I can't figure out why this is even there. It does exactly the same as clique and we shouldn't need it at all.

BigRoy: Cool, let's remove and reduce unnecessary code. :)

avalon/nuke/lib.py
```diff
@@ -0,0 +1,230 @@
...
+        # unselect all selection in case there is some
+        current_seletion = nuke.selectedNodes()
+        [n['selected'].setValue(False) for n in current_seletion]
+        # and select all previously selected nodes
+        if previous_selection:
+            [n['selected'].setValue(True) for n in previous_selection]
+
+
+def reset_selection():
+    nodes = nuke.allNodes()
```
BigRoy: If this command is required... note that it could be as simple as:
```python
[n['selected'].setValue(False) for n in nuke.selectedNodes()]
```

_Personal note:  Also, somehow I'm more inclined to the name `clear_selection` instead of `reset_selection`, but I guess that's a random preference. :)_

mkolar: I've actually removed both reset selection and select nodes as they weren't used anywhere and are not needed. Must have stayed behind from testing.

mkolar:
has anyone had a chance to try this? we have it deployed and it works ok. Albeit over time we'll certainly find places for many improvements. I think if no-one has any complaints we could merge. I've fixed all the comments above I believe

BigRoy:
Unfortunately I'm unable to test this as I don't have any Nuke licenses. Maybe @davidlatwe is able to chime on this one to do a test?

If it works for you and your clients, noone else can test it and the code looks clean I guess it should be ready to merge. It looks great.

By the way, do you have any pictures/gifs/videos of it in action?

mkolar:
> By the way, do you have any pictures/gifs/videos of it in action?

nope but will do.

mottosso:
Also can't test this, but happy to leave it with you guys. Agree with @BigRoy that we need gifs or video of what it looks like and how it works.

BigRoy:
@mkolar Is this still the up-to-date state of a working Nuke integration? Plus, are you able to produce GIFs/JPEGS or some other type of image/video to showcase the working integration.

BigRoy:
@tokejepsen @mkolar @jezscha is any of you able to share some GIFs/videos of a working Nuke integration of Avalon plus make sure the up-to-date code for core is ready to go on this PR or another? As part of [Avalon 6.0](https://github.com/getavalon/core/projects/3) I think it'd be great to get the Nuke integration merged and ready, plus potentially documented - yet unfortunately I myself do not work with Nuke. ;)

tokejepsen:
As far as I know Nuke integration is still being worked on. Or maybe it was NukeStudio?

BigRoy:
Well, there do seem to be some updates not in this PR, like for Nuke https://github.com/pypeclub/avalon-core/commit/3c4ae51b1ced638c851896f38409608fe5933bd8 and for Nuke Studio https://github.com/pypeclub/avalon-core/commit/a9b794a81d39c61c67140d97a6b59cd207715589

mkolar:
I believe that this PR is still valid for the basic implementation. Yes we were adding some more details in different branches and features on our side, but it's not production tested yet. I'll double check with @jezscha today

mottosso:
Having a look at this one now. The `workfiles` app seems to have been updated inbetween this PR and now, I'll have a look at these now.

mottosso:
I've implemented the `workio.py` file consistent with `maya`, `fusion` et. al. based on the implementations found in in the workfiles app from this PR.

The only one I wasn't sure of was `workio.work_root()`, which I've made `os.path.dirname(workfile.current_file())`, @tokejepsen do you know if that is right? :S

BigRoy:
> The only one I wasn't sure of was `workio.work_root()`, which I've made `os.path.dirname(workio.current_file())`

@mottosso that's definitely wrong. See [documentation here](https://github.com/getavalon/core/blob/master/avalon/tools/workfiles/README.md#work-files-scenes-root-avalon_scenedir)

mottosso:
Ah, yes I was looking at Maya for guidance, but couldn't translate its idea of "workspace".  That doc is great, but isn't it saying to return the parent directory of he current file? What should the function return instead?

BigRoy:
It should not use a current file as it could actually even be queried when there is no currently saved file.

Anyway, documentation there does seem to lack the code example for those cases, however it should be similar to [this](https://github.com/getavalon/core/blob/master/avalon/fusion/workio.py#L42)

mottosso:
Thanks, it's been fixed.

tokejepsen:
> The only one I wasn't sure of was workio.work_root(), which I've made os.path.dirname(workio.current_file()), @tokejepsen do you know if that is right? :S

We have worked with this; https://github.com/pypeclub/avalon-core/blob/develop/avalon/nuke/workio.py

mottosso:
Thanks, fixed. Curious, is `nuke.root()` the same as `nuke.Root()`? The previous version used lower-case. :S

tokejepsen:
> Thanks, fixed. Curious, is nuke.root() the same as nuke.Root()? The previous version used lower-case. :S

Honestly dont know. Think both works.

Issue #351, opened 2018-11-08
Using representation config instead of parents when formatting template in loader


mkolar:
This is directly related to #121

We have more complex templates than what is expected by default. That means our paths are not formatted correctly when trying to load a representation. 

The issue seems to be here
https://github.com/getavalon/core/blob/79ea85a26f0296d7c1f72670d377611d4502307b/avalon/pipeline.py#L161-L167

We are storing context in the db, but the data used to format the publish template is explicitly set to be only made up of parents. That results in technically hardcoding how we can work with templates and is not very flexible. 

I suggest using `data = context['representation']['context']` instead to make this more arbitrary and open to customisation in different studios. 

I've done a quick test and it seems to solve all our issues and shouldn't break any backwards compatibility (we might have to do it via condition though in case old db data doesn't have the `context` field in the representation)

tokejepsen:
What are your thoughts on the issue of updating old assets when a parent is modified?

mkolar:
Not sure what exactly you mean. Can you give example?

tokejepsen:
Sorry ignore that comment. I misunderstood the issue.

mkolar:
On top of this we would actually also save the template that was originally used to format the path. This is to prevent mismatch during the import in case the publish template had to be tweaked during the production. (No it shouldn't happen, Yes it might happen on longer productions)

So the representation itself would hold all the data needed to format the right path, and still allow for overriding some keys if the project is for example moved to a different root.

BigRoy:
> I suggest using data = context['representation']['context'] instead to make this more arbitrary and open to customisation in different studios.

I think allowing to make it more arbitrary is perfect! It should allow formatting with custom data or other required keys a studio might introduce in their configuration so that they take more control over their template.

However, we should choose a simple interface for doing so that will remain consistent. That could be `context` or maybe just the representation's `data`. It should be something that makes sense to customize. Both `context` and `data` sound like good options to me, if at least the functionality is documented clearly for new developers.

@mottosso might have some good input on keeping this lean and mean and where to put it. Also especially since he originally created #121.

mottosso:
Thanks for pinging me on this issue, I think @mkolar is right about storing `context` inside of the representation. My initial concern was about whether a parent is allowed to change; for example, if the project is called `hulk` and the name `hulk` is stored in a representation, than if `hulk` were to change and become e.g. `hulkRising` then suddenly every representation would be out of date.

[...]

Issue #308, opened 2018-01-27
Deeper hierarchy in templates and paths


mkolar:
## Issue
Currently it's doesn't seem possible to create more complex hierarchies using avalon than the one in the default config https://getavalon.github.io/2.0/reference/#project-configuration-api

A deeper hierarchy is often needed, especially for bigger projects like features and episodic TV. 
`{root}/{project}/{silo}/{asset}/work/{task}/{app}`
with simple episodes this could still work, having each episode as a silo
`{root}/{project}/episodes/{silo}/{asset}/work/{task}/{app}`
But there are also more complicated scenarios:
```
- Development
    - Concept
    - Teaser
        - Animatic
        - Previz
        - Shots
            - sh010
            - sh020
```
now to get to those shots in a template it might even be hard to specify the member names
`{root}/{Folder1}/{Folder2}/{Folder3}/{shot}/work/{task}`

## Proposal

The system we use  outside avalon right now is to save anything above `shot/asset` to a member called `hierarchy`. Our templates then look like this
`{root}/{hierarchy}/{shot}/work/{task}` where `hierarchy = "development/teaser/shots`

In term of avalon this path member could always contain `silo` and could prepend any other visual parents or categories. It would allow creating completely arbitrary folder structures.

This whole thing stems from the fact, that we're technically unable to convert our ftrack project structures to avalon, because it's lacking the granularity we need.

BigRoy:
Visually it's possible to have a deeper hierarchy by hooking it up with a `visualParent`. For example the project manager can create "nested" items by selecting a previously existing asset and then "adding" an asset. These new nested assets get their parent prefixed in the asset name `{parent}_{asset}` - this could also be swapped to `{parent}/{asset}` -> where if the parent would also be prefixed with a potential parent would allow any deep nesting, for example the parent could be `s01e01/seq01` and then this asset is shot 010 within that sequence. The full name then becomes `s01e01/seq01/010`.

I'm not sure if the forward slash is supported in the asset's name, but if it is then even the folders created on disk would allow to "contain" assets. The only problem there would be what if you create an asset called `work` within a parent asset and your template also defines the work folder as `work` within an asset. They would end up being the same folder. That's why I had opted to, at least on disk, make it flat, yet make the browseable hierarchy visible in the interfaces.

That's also one of the reason to "prefix" the parent in the asset's name. But the other reason is: that asset names in the database have to be unique, so for example `010` couldn't end up twice in the database - even if nested because in the database it would always remain flat.

[...]

PR #371, opened 2019-02-03
More flexible templates 

mkolar:
PR relates to both #308  and #351 

## New Functionality

This PR allows Avalon to work with arbitrary hierarchies in the system. This is currently used in 2 places.

Filling work path when launching an app and in 'workfiles' app and getting the path of published representation during loading. 

### Getting representation path

Instead of hardcoding what keys we allow to be used in the template and also which template has to be used for publishing (as it is currently), store all the information we need directly on the representation. Proposed solution uses three possibilities to get the path which we try in sequence until one is successful.
 1. Get `representation['data']['path']` directly. We assume this is pre-filled full path to the file. If the folder exists on the disk, we use this
2. get `representation['data']['template']` and format it with `representation['context']. During this process we make sure that `root` from the context get's replaced with registered root in avalon to keep it dynamic. 
3. fallback to current way if none of the above was successful. 

example of what representation can look like in the db then
```
id: 5c41dcd19e4ee0671c7e594a
name: "exr"
parent : 5c41dcd19e4ee0671c7e5949
type : "representation"
dependencies : Array
context : Object
    subset: "render_writeDefault"
    family: "write"
    hierarchy: "sequences\sq01"
    version: v003
    project: 
        code: "mln"
        name: "super_project"
    task: "compositing"
    asset: "sh010"
    representation: "exr"
    root: "C:\projects\_AVALON"
    padding: #####
data: Object
    path: "C:\projects\_AVALON\milan_testing\sequences\sq01\sh010\publish\render\compositing\render_writeDefault\v003\mln_sh010_compositing_v003_render_writeDefault.#####.exr"
    template : "{root}/{project[name]}/{hierarchy}/{asset}/publish/{family}/{task}/{subset}/{version}/{project[code]}_{asset}_{task}_{version}_{family}_{subset}.{padding}.{representation}
```

### Formatting work path

The issue here was allowing any hierarchy to be filled into the path without putting too many restriction at what keys (sequence, shot, episode etc) can be used. The solution is introducing concept of `AVALON_HIERARCHY` which get's constructed on the fly from new `parents` data on the given asset. 

`asset['data']['parents']` is expected to be a correctly ordered list of all the hierarchical parents of the asset. This list get's parsed and joined with `os.path.sep` ending up as a simple string that can be used in the template as `{hierarchy}` key.

We've used this in production for a while now, but it needs to be tested for backwards compatibility of course. As far as we can tell, nothing should break on previous projects, but better safe than sorry. 

We found this feature crucial as different studios can have some seriously different preferences on what their folder structure looks like :). The implementation itself of course is up for discussion. The priority of individual ways of getting the path from representation for example. I find storing the absolute path in data useful as it might later allow extra checks to happen.


avalon/pipeline.py
```diff
@@ -967,6 +956,9 @@ def update_current_task(task=None, asset=None, app=None):
     _session = Session.copy()
     _session.update(changed)
     changed["AVALON_WORKDIR"] = _format_work_template(template, _session)
+    changed['AVALON_HIERARCHY'] = os.path.sep.join(
+            asset_document['data']['parents']
```
BigRoy: This will break pipelines that don't have the `parents` data, right? This should allow for backwards compatibility.

avalon/pipeline.py
```diff
@@ -1003,6 +995,7 @@ def _format_work_template(template, session=None):
         "root": registered_root(),
         "project": session["AVALON_PROJECT"],
         "silo": session["AVALON_SILO"],
+        "hierarchy": session['AVALON_HIERARCHY'],
```
BigRoy: Will this error out on pipelines who haven't set `AVALON_HIERARCHY` yet?

1. Should this be in the Session schema if it's required?
2. What is this value and when would you set it? How would you compute it for any asset?

avalon/pipeline.py
```diff
@@ -1198,20 +1191,39 @@ def get_representation_path(representation):
...
-        "asset": asset["name"],
-        "silo": asset["silo"],
-        "subset": subset["name"],
-        "version": version_["name"],
-        "representation": representation["name"],
-        "user": Session.get("AVALON_USER", getpass.getuser()),
-        "app": Session.get("AVALON_APP", ""),
-        "task": Session.get("AVALON_TASK", "")
-    })
+    output = None
```
BigRoy: Could you add in the docstring the order of which it resolves the paths in the simplest sense. That way one can easily figure it out without reading the lines of code. It seems to have 3 ways of getting the data and I think it would be good to keep that order and reasoning clear for anyone trying out the function without looking at the code implementation.,

BigRoy:
_Just a note, not a direct comment on your code of implementation._

More dynamic paths I think would be warmly welcomed when it doesn't complicate things too much - since it allows for more customization for a specific studio. We just need to ensure that the template we give and allow isn't something that would easily break and introduces issues for simple pipelines or newcomers. (E.g. just storing the full path in a multiple OS pipeline will tend to introduce issues.)

avalon/pipeline.py
mkolar: yes I'd say that making is required in the schema is reasonable. however there is a caveat here. Because it hold multiple key in the path joined together with a path.sep, it will be quite hard to parse the opposite way. So for example what you are using in your colorbleed config (get context from path) will likely not work. 

AVALON_HIERARCHY resolve to a list of parents joined by path.sep so for example If asset is in this hierarchy: `project1/assets/characters/hero`, then the template would be `{project/{hierarchy}/{asset}. 

`hierarchy` = 'assets/characters`

mkolar:
> (E.g. just storing the full path in a multiple OS pipeline will tend to introduce issues.)

very true. it's bothering me too. we should probably make the stored full path as that last resort if all else fails.

avalon/pipeline.py
mkolar: true. we'll put a .get() there with default of empty string to make sure it always resolve to something.

mottosso:
Hey guys, should we merge this?

mkolar:
Now I believe we can. I've changed the order of resolution based on the comments and it is now.

1. Get template from representation['data']['template'] and data from
       representation['context']. Then format template with the data.
2. Get template from project['config'] and format it with default data set
3. Get representation['data']['path'] and use it directly

If any is successful, the following won't be tried. 1 is the new way of getting all needed from representation, 2. is exactly the same as before and 3. is fallback to a full path potentially stored on the representation.

I've also added this into the docstring. We're quite happy with this and have it deployed in multiple places now.

tws0002:
Hi Mkolar,

i having this error

```
Running action: maya2018
  File "K:\test\avalon-setup\git\launcher\launcher\control.py", line 418, in trigger_action
    popen = action.process(api.Session.copy())
  File "K:\test\avalon-setup\git\core\avalon\pipeline.py", line 405, in process
    environment = self.environ(session)
  File "K:\test\avalon-setup\git\core\avalon\pipeline.py", line 316, in environ
    workdir = _format_work_template(template, session)
  File "K\test\avalon-setup\git\core\avalon\pipeline.py", line 935, in _format_work_template
    "hierarchy": session['AVALON_HIERARCHY'],
KeyError: 'AVALON_HIERARCHY'
avalon.py: Finishing up..
```

iLLiCiTiT:
> Hi Mkolar,
> 
> i having this error
> 
> ```
> Running action: maya2018
> Traceback (most recent call last):
>   File "K:\test\avalon-setup\git\launcher\launcher\control.py", line 418, in trigger_action
>     popen = action.process(api.Session.copy())
>   File "K:\test\avalon-setup\git\core\avalon\pipeline.py", line 405, in process
>     environment = self.environ(session)
>   File "K:\test\avalon-setup\git\core\avalon\pipeline.py", line 316, in environ
>     workdir = _format_work_template(template, session)
>   File "K\test\avalon-setup\git\core\avalon\pipeline.py", line 935, in _format_work_template
>     "hierarchy": session['AVALON_HIERARCHY'],
> KeyError: 'AVALON_HIERARCHY'
> avalon.py: Finishing up..
> ```

@mkolar should we also use:
```
"hierarchy": session.get('AVALON_HIERARCHY', ''),
```
in _format_work_template for case when hierarchy is empty?

mkolar:
>@mkolar should we also use:
>"hierarchy": session.get('AVALON_HIERARCHY', ''),
>in _format_work_template for case when hierarchy is empty?

yes this should be fixed. we didn't catch it before

avalon/pipeline.py
mottosso: This needs to be made optional, as per the comments.

avalon/pipeline.py
mottosso: This seems to have been addressed.

avalon/pipeline.py
```diff
@@ -1198,20 +1201,50 @@ def get_representation_path(representation):
...
+    try:
```
mottosso: This try/except block is too broad. At the moment, if the first formatting fails, the second doesn't happen. Also that first `if output is None:` is superflous, can we remove that?

mottosso:
Mentioning #417 here, as this is one of those features that are important but difficult to test. For a test to pass in each permutation of this functionality, we need:

1. No project template, no representation template and no representation path, returning `None`, and what to do with that; this is up to tools currently to visualise to the user.
2. No project template, no representation template, but a representation path, and how that should work cross-platform.
3. No project template, but representation template. Pretty straight forward
4. Finally, a project template, the prior default and current fallback.

In each case, we need both a database entry per pair, but also existing paths on the file-system, as these won't return anything unless the path actually exists; unsure of whether that's actually the responsibility of this function, but am not sure how else to "automatically" cycle through the options (or whether this is actually better suited for handling by tools that can visualise this).

mottosso:
I made a mistake on this one, the paths returned currently are the directories, not filenames! :O

**Original**

https://github.com/pypeclub/avalon-core/blob/da55d0f4f601e613a03e05ee3db33eaece19ec0f/avalon/pipeline.py#L1214

**Refactored**
https://github.com/getavalon/core/blob/9b89842047fec06d221f178191004544d5660837/avalon/pipeline.py#L1227-L1229

Fixing this now.

PR #372, opened 2019-02-03
Support silo assets

mkolar:
Somewhat related to #308

## New Functionality

This PR turns silos into full assets with option to hold tasks as a first step towards hopefully eradicating them altogether at some point :). 

### Implementation

We propose simply considering every asset with `silo` data member set to `None` as a silo itself. That solution is not 100% foolproof, but it's certainly a straightforward and simple one. It's fully backwards compatible as it simply combines any 'asset silos' with previously defined 'attribute silos' in all the GUIs. 

To be able to display silo tasks, we had to tweak the display logic a bit. To display tasks on the silo (which presents itself as a tab in the GUI) user simply needs to deselects all in the treeview. This behaviour overrides the previous where you would see project tasks by default.

### DIFF
--- a/avalon/tools/cbloader/app.py
+++ b/avalon/tools/cbloader/app.py
@@ -170,12 +170,11 @@ class Window(QtWidgets.QDialog):
 
         asset_item = assets_model.get_active_index()
         if asset_item is None or not asset_item.isValid():
-            type = "asset"
             silo = assets_model.get_current_silo()
             if len(silo) == 0:
                 return
             document = io.find_one({
-                "type": type,
+                "type": "asset",
                 "name": silo
             })
         else:
--- a/avalon/tools/cbloader/app.py
+++ b/avalon/tools/cbloader/app.py
@@ -106,6 +106,8 @@ class Window(QtWidgets.QDialog):
 
         refresh_family_config()
 
+        self._refresh()
+        self._assetschanged()
         # Defaults
         self.resize(1330, 700)
 
@@ -168,9 +170,19 @@ class Window(QtWidgets.QDialog):
 
         asset_item = assets_model.get_active_index()
         if asset_item is None or not asset_item.isValid():
-            return
+            type = "asset"
+            silo = assets_model.get_current_silo()
+            if len(silo) == 0:
+                return
+            document = io.find_one({
+                "type": type,
+                "name": silo
+            })
+        else:
+            document = asset_item.data(DocumentRole)
 
-        document = asset_item.data(DocumentRole)
+        if document is None:
+            return
         subsets_model.set_asset(document['_id'])
 
         # Enforce the columns to fit the data (purely cosmetic)
--- a/avalon/tools/cbloader/app.py
+++ b/avalon/tools/cbloader/app.py
@@ -3,7 +3,7 @@ import time
 
 from ..projectmanager.widget import AssetWidget, AssetModel
 
-from ...vendor.Qt import QtWidgets, QtCore, QtGui
+from ...vendor.Qt import QtWidgets, QtCore
 from ... import api, io, style
 from .. import lib
 
@@ -23,7 +23,7 @@ class Window(QtWidgets.QDialog):
     def __init__(self, parent=None):
         super(Window, self).__init__(parent)
         self.setWindowTitle(
-            "Asset Loader 2.0 - %s/%s" % (
+            "Asset Loader 2.1 - %s/%s" % (
                 api.registered_root(),
                 api.Session.get("AVALON_PROJECT")))
 
@@ -56,7 +56,7 @@ class Window(QtWidgets.QDialog):
         split.addWidget(asset_filter_splitter)
         split.addWidget(subsets)
         split.addWidget(version)
-        split.setSizes([225, 925, 0])
+        split.setSizes([180, 950, 200])
         container_layout.addWidget(split)
 
         body_layout = QtWidgets.QHBoxLayout(body)
@@ -101,12 +101,13 @@ class Window(QtWidgets.QDialog):
 
         families.active_changed.connect(subsets.set_family_filters)
         assets.selection_changed.connect(self.on_assetschanged)
-        subsets.active_changed.connect(self.on_versionschanged)
+        subsets.active_changed.connect(self.on_subsetschanged)
+        subsets.version_changed.connect(self.on_versionschanged)
 
         refresh_family_config()
 
         # Defaults
-        self.resize(1150, 700)
+        self.resize(1330, 700)
 
     # -------------------------------
     # Delay calling blocking methods
@@ -117,12 +118,16 @@ class Window(QtWidgets.QDialog):
         lib.schedule(self._refresh, 50, channel="mongo")
 
     def on_assetschanged(self, *args):
-        self.echo("Fetching results..")
+        self.echo("Fetching asset..")
         lib.schedule(self._assetschanged, 50, channel="mongo")
 
-    def on_versionschanged(self, *args):
-        self.echo("Fetching results..")
+    def on_subsetschanged(self, *args):
+        self.echo("Fetching subset..")
         lib.schedule(self._versionschanged, 50, channel="mongo")
+
+    def on_versionschanged(self, *args):
+        self.echo("Fetching version..")
+        lib.schedule(self._versionschanged, 150, channel="mongo")
 
     def set_context(self, context, refresh=True):
         self.echo("Setting context: {}".format(context))
@@ -265,7 +270,7 @@ class Window(QtWidgets.QDialog):
         return super(Window, self).closeEvent(event)
 
 
-def show(root=None, debug=False, parent=None, use_context=False):
+def show(debug=False, parent=None, use_context=False):
     """Display Loader GUI
 
     Arguments:
@@ -327,7 +332,9 @@ def show(root=None, debug=False, parent=None, use_context=False):
 
 
 def cli(args):
+
     import argparse
+
     parser = argparse.ArgumentParser()
     parser.add_argument("project")
 
@@ -336,6 +343,17 @@ def cli(args):
 
     io.install()
 
+    # Store settings
     api.Session["AVALON_PROJECT"] = project
+
+    from avalon import pipeline
+
+    # Find the set config
+    _config = pipeline.find_config()
+    if hasattr(_config, "install"):
+        _config.install()
+    else:
+        print("Config `%s` has no function `install`" %
+              _config.__name__)
 
     show()
--- a/avalon/tools/projectmanager/model.py
+++ b/avalon/tools/projectmanager/model.py
@@ -229,12 +229,12 @@ class TasksModel(TreeModel):
             asset_tasks = asset.get("data", {}).get("tasks", [])
             tasks.update(asset_tasks)
 
-        # If no asset tasks are defined, use the project tasks.
-        if assets and not tasks:
-            project = io.find_one({"type": "project"})
-            tasks.update(
-                [task["name"] for task in project["config"].get("tasks", [])]
-            )
+        # # If no asset tasks are defined, use the project tasks.
+        # if assets and not tasks:
+        #     project = io.find_one({"type": "project"})
+        #     tasks.update(
+        #         [task["name"] for task in project["config"].get("tasks", [])]
+        #     )
 
         self.clear()
         self.beginResetModel()
--- a/avalon/tools/projectmanager/model.py
+++ b/avalon/tools/projectmanager/model.py
@@ -113,7 +113,8 @@ class TreeModel(QtCore.QAbstractItemModel):
                 key = self.COLUMNS[column]
                 node[key] = value
 
-                self.dataChanged.emit(index, index)
+                # passing `list()` for PyQt5 (see PYSIDE-462)
+                self.dataChanged.emit(index, index, list())
 
                 # must return true if successful
                 return True
--- a/avalon/tools/projectmanager/model.py
+++ b/avalon/tools/projectmanager/model.py
@@ -229,17 +229,14 @@ class TasksModel(TreeModel):
             asset_tasks = asset.get("data", {}).get("tasks", [])
             tasks.update(asset_tasks)
 
-        # # If no asset tasks are defined, use the project tasks.
-        # if assets and not tasks:
-        #     project = io.find_one({"type": "project"})
-        #     tasks.update(
-        #         [task["name"] for task in project["config"].get("tasks", [])]
-        #     )
+        # If no asset tasks are defined, use the project tasks.
+        if len(tasks) == 0:
+            project = io.find_one({"type": "project"})
+            tasks.update(
+                [task["name"] for task in project["config"].get("tasks", [])]
+            )
 
         self.clear()
-        # let cleared task view if no tasks are available
-        if len(tasks) == 0:
-            return
 
         self.beginResetModel()
 
--- a/avalon/tools/projectmanager/model.py
+++ b/avalon/tools/projectmanager/model.py
@@ -228,6 +228,13 @@ class TasksModel(TreeModel):
             asset_tasks = asset.get("data", {}).get("tasks", [])
             tasks.update(asset_tasks)
 
+        # If no asset tasks are defined, use the project tasks.
+        if assets and not tasks:
+            project = io.find_one({"type": "project"})
+            tasks.update(
+                [task["name"] for task in project["config"].get("tasks", [])]
+            )
+
         self.clear()
         self.beginResetModel()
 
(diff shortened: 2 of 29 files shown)


avalon/tools/cbloader/app.py
```diff
@@ -168,9 +170,19 @@ def _assetschanged(self):
 
         asset_item = assets_model.get_active_index()
         if asset_item is None or not asset_item.isValid():
-            return
+            type = "asset"
```
BigRoy: @mkolar could you put a comment here as to what this is achieving? Just looking at the code I'm having a hard time figuring it out how this solves it.

Also, to not shadow the `type` function I'd recommend just passing `"asset"` to `io.find_one` directly on line 178.

BigRoy: Ah, wait... is this working in such a way that the loader still displays "silo" as tabs in the loader. Visually nothing changes, except that when you select nothing at all (no asset in selection) that you are operating on the 'silo asset'?

avalon/tools/projectmanager/widget.py
```diff
@@ -180,6 +185,16 @@ def __init__(self, silo=None, parent=None):
     def set_silo(self, silo, refresh=True):
         """Set the root path to the ItemType root."""
         self._silo = silo
+        try:
+            self.silo_asset = io.find_one(
+                {'$and': [
```
BigRoy: This doesn't need to be an "$and" query, right? Couldn't this be just one regular query?
```python
io.find_one({
    'type': 'asset',
     'name': silo,
     'silo': None
})
```

BigRoy:
Went over the code, seemed alright - but at first sight it might need a clarification here or there on how it works.

> We propose simply considering every asset with silo data member set to None as a silo itself. 

This helps with explaining it though - it's just that the code took me a bit to grasp why all the if statements were needed.

avalon/tools/projectmanager/model.py
```diff
@@ -229,14 +229,18 @@ def set_assets(self, asset_ids):
             asset_tasks = asset.get("data", {}).get("tasks", [])
             tasks.update(asset_tasks)
 
-        # If no asset tasks are defined, use the project tasks.
-        if assets and not tasks:
-            project = io.find_one({"type": "project"})
-            tasks.update(
-                [task["name"] for task in project["config"].get("tasks", [])]
-            )
+        # # If no asset tasks are defined, use the project tasks.
```
BigRoy: Why did you end up commenting this out? :)

iLLiCiTiT: That's because I wasn't sure why Project manager showed project's tasks values if there are no tasks to show on selected asset and was scared to remove :) It had any reason?

BigRoy: > That's because I wasn't sure why Project manager showed project's tasks values if there are no tasks to show on selected asset and was scared to remove :) It had any reason?

Ha, good - I like it your way too. ;) It's so confusing!

Anyhow, it got implemented here: https://github.com/getavalon/core/pull/330

I guess it was to match the behavior of the Launcher, which allowed to still go into an asset and work on a task when no explicit tasks where assigned. (Which I guess is backwards compatibility, since originally tasks were per project, not assigned per task.)

avalon/tools/projectmanager/widget.py
BigRoy: Hmm, could you set up a reproducable example? It should work.

Also, the `try... except`  shouldn't be needed since `find_one()` will return None if there's no match. Unless you're expecting to pass specific values to the function that will break the query - but I wouldn't know why that would happen?

avalon/tools/cbloader/app.py
iLLiCiTiT: > `Also, to not shadow the type function I'd recommend just passing "asset" to io.find_one directly on line 178.`

True. It doesn't make sence to store it into variable here.

avalon/tools/projectmanager/widget.py
iLLiCiTiT: You're absolutely right... just realized what was the reason why it didn't work correctly when I tried it. That was also the reason why I put that part into `try...except.` Now, I see! Thanks

avalon/tools/projectmanager/model.py
mkolar: I for one can't figure out what was the reasoning the in the first place. The only reason I can think of is the lack of project manager GUI so it was had to add individual tasks to assets, hence relying on a global task list. 

maybe @mottosso or @tokejepsen can shed some light on what is the use case?

tokejepsen: @mkolar sorry not sure what is the confusion?
Is it that the project manager shows the global tasks, or that you can see tasks per asset?

mkolar: What was the reason for project manager to show global tasks? Is the question. It seems that both us and colorbleed prefers only seeing tasks per asset, so I was wondering where the feature came from.

tokejepsen: Pretty sure that is a legacy feature from not having tasks per asset.

I would prefer having tasks per entity in Avalon, so you can explicitly say what tasks are available for the user.

BigRoy: Great - my preference would also be then to remove them here too.

mkolar: In that case we'll remove the commented code completely as the resulting behavior is exactly as agreed here.

mottosso:
Sorry, just saw this PR.

It isn't clear to me what problem was this solving? There seems to be some confusion about what "silo" is and why it exists; it's a filter, much like `families` for Pyblish plug-ins. It was put in place to avoid separating between what was a shot and what was an asset (or episode, or sequence etc.) such that every object in the database could be just an "asset". But the user still needed to visualise shots differently from assets, and so `silo` was born.

I've seen screenshots now of additional silos, like `editorial` which makes a lot of sense I think. In theory, we could have lots of filters for the various GUIs, the `visualParent` is another one. Things that only apply to GUIs and not to our pipeline tools.

Dunno, does that change anything about this PR? Do we still want silos to be assets? (seems very strange to me!) Maybe we should rename them `filters`?

mkolar:
As we work with this many more problems with silos actually arose for us. This is a first step to solving them. 

1. Yes silo is a virtual concept that is technically just a filter. However when we try to relate to them in the overall project hierarchy we're constantly hitting walls. The main one was the fact we need to have tasks on the top level hierarchy members...so technically silos. This is for editorial, but also some development tasks. Hence we needed to add this possibility and the only way to do it is to turn silos in to normal asset.

That being said. We're not saying `silo` should be an asset, but rather "the highest asset in hierarchy should act as silo in the GUI. so effectively we're ignoring silo and just didn't want to do a rewrite of all the GUIs to support not having them.

2. Second step would be dropping them completely. Starting using filters as you said so we can build GUIs that visually reorganize it how we want. The tabs we have for silos now for example are only useable if you have 2-3 silos. Our projects regularly have 4-5 and more top folders, so it's suddenly not very user friendly.

Long story short I absolutely agree that all these filters (including silo) should only affect GUI. However because we have to act fairly fast with production requirement, we need to take these features step by step to make clients happy. In this case for example dropping silo gave us the ability to make sure that if we redo this categorisation later on, everything can be made backwards compatible very easily as we only added new assets to db.

tokejepsen:
> the highest asset in hierarchy

Sorry, could you clarify what you mean by this, since the asset are in a flat hierarchy on the dB?
Do you mean the ```visualParent``` hierarchy?

mottosso:
I have the same question.

Don't you already have a concept of (visual) hierarchy using the `visualParent` attribute, and could you not use that in-place of how you currently use `silo`?

mkolar:
Yes I mean visual hierarchy. And yes we do have it from visualParents.

And frankly Id be happier is we removed reference to Silo from this and instead found assets without visualParent and used those as tabs...

mottosso:
I think that could work!

It'd be compatible with the Mongo documents already, and look and act identically to silos as far as users are concerned (since it's basically just moving a value from `silo` to another attribute `visualParent`), plus it doesn't eliminate the possibility for filtering by things other than hierarchy, like arbitrary tags.

The only addition is I would probably keep `visualParent` in those "top-level" documents, but let it be None. That way, it's clear that they are meant to be part of a hierarchy, and that they don't have a parent and must be on top.

mkolar:
Yes keeping as none is the way to go.

Were doing that with silo now.

tokejepsen:
> The only addition is I would probably keep visualParent in those "top-level" documents, but let it be None

Would that mean that ```visualParent``` is turning into a required attribute of the schema?

mottosso:
> Would that mean that `visualParent` is turning into a required attribute of the schema?

Good question; there's probably no rush to do so, we can see how things play out the way things are currently.

But if it does end up in the schema, I wouldn't want it to refer to anything visual. Visuals should interpret abstract data, not the other way around. So we may want to consider renaming it to e.g `parentHint` or the like, and interpret that as a "visual hierarchy" wherever it's used. Just so we don't start blurring the line between data and what is strictly related to drawing that data.

mkolar:
Even though it's technically not required now, Without it none of the GUIs make any sense, so it's almost sorta required already ;)

mottosso:
Ok, so this issue is about making the tabs of the Library GUI refer to top-level assets with `visualParent` set to `None`. A GUI change, effectively, rather than anything in the DB or application logic.

We could then technically keep `silo`, for the time being, even though it wouldn't be represented anywhere.

Does that capture the requirement fully? If so, let's either update the description or make a new issue.

mkolar:
yes. for us that captures it and would be preferable to us. We'll update this PR and the code to reflect.

avalon/tools/projectmanager/widget.py
```diff
@@ -143,7 +143,18 @@ def preserve_selection(tree_view,
 
 def _list_project_silos():
     """List the silos from the project's configuration"""
-    silos = io.distinct("silo")
+    # Backwards compatible way to get silos/top assets from DB:
```
BigRoy: Wouldn't it be easier to do this:

```python
silos = io.distinct("silo")   # Backwards compatible silo support

# New style silos getting assets without parent and silo
new_silos = [s['name'] for s in io.find({
        'type': 'asset',
        'data.visualParent': None,
        '$or': [{'silo': {'$exists': False}}, {'silo': None}]
    })]
silos.extend(new_silos)

return silos
```

- It supports backwards compatible silos together with new style silos in one project.
- It does less database queries.
- It's more readable (to me at least)

iLLiCiTiT: That won't work actually because io.distinct("silo") will contain None in new way. I had it similar way before:
```
silos = [s['name'] for s in io.find({"type": "asset", "silo": None})]
silos_old = io.distinct("silo")
for silo in silos_old:
     if silo not in silos and silo is not None:
         silos.append(silo)
```
But the point of this commit was prepare to not use 'silo' column in queries at all

mottosso: Do we have to change `silo` at all? I thought this change was about ignoring silo, and using the top-level asset instead?

BigRoy: > That won't work actually because io.distinct("silo") will contain None in new way.

Then like this?

```python
silos = io.distinct("silo") or []
```

mkolar: > Do we have to change `silo` at all? I thought this change was about ignoring silo, and using the top-level asset instead?

that's exactly what we're trying to avoid. I'll try to explain what @iLLiCiTiT was trying to achieve here :).

The key point here is how to make it backwards compatible while also making sure that the new way of determining what is the 'top-level asset' doesn't need any silo data whatsoever. 

I think the confusion here with the naming. Silo is just data, but what we're trying to do is make a list of Tabs in the GUI. This can be made either from silo data if the project has it, or finding assets with `visualParent == None`. The problem is that old project also have asset with visualParent None, but in those case, this project shouldn't become a Tab. So we actually need to distinguish between old type and new type of project to ensure backward compatibility. 

here's slight change to code that should make the distinction clear. This way one wouldn't need silo at all and its clear we're not actually collecting silos, just data that forms tab in the GUI.

```
def _list_project_silos():
    # Let's find whether we're dealing with old or new type of project (having no silos):
    silo_existence = {
        'type': 'asset',
        'silo': {'$exists': False}]
    }
    # if we have new project (one with no silo data at all), get tabs from visualParent == None
    if len(list(io.find(silo_existence))) > 0:
        tabs = [s['name'] for s in io.find({
            'type': 'asset',
            'data.visualParent': None
        })]
    # if we have old project get tabs the old way
    else:
        tabs = io.distinct("silo")

    if not silos:
        project = io.find_one({"type": "project"})
    return list(sorted(tabs))
```

BigRoy: You wouldn't need to check for the full project having one of it. Whenever something does *not* have a visual parent and does *not* have silo data, then it's your top most silo tab - it's as simple as that, right? ;)

It will even allow compatibility inside a single project to mix it up. (if you work on older projects)

iLLiCiTiT: And what about:
```
silos = io.distinct("silo")

# New style, works if 'silo' column doesn't exists
if not silos:
        silos = [s['name'] for s in io.find({
                'type': 'asset',
                'data.visualParent': None
            })]
```
- easy to separate, strictly dividing
- it doesn't allow to mix it up but I don't think it's good idea to allow mix different hierarchy approache, could cause huge issues

BigRoy: > it doesn't allow to mix it up but I don't think it's good idea to allow mix different hierarchy approache, could cause huge issues

I'm actually not seeing something that could cause a big issue. ;) Even if you have a silo `film` and later you create that asset `film` that can act as silo (having no silo set and no visualParent) then I believe it would know how to parent it. Right?

I'm just more afraid of someone accidentally making an old style silo (or the other way around, a new one) and it not being picked up or suddenly discarding the ones you had. But yes, this is definitely already looking like more readable code.

mkolar: >I'm actually not seeing something that could cause a big issue

the main on is a bit down the line. Distinguishing what to display as the top asset in the hierachy widget appearing under the tabs. If you have old style silos, the top asset in the tree view is the one with visualParent:None, if you're however using that as tab , it should be the next one in the hierarchy. So if you allow mixed type in the project, it goes a bit down the rabit hole. Isn't it more trouble than it's worth? 

If we definitely want to allow mixed projects, then you're latest suggestion is the way to go. And we'd have to implement it on a few more places.

BigRoy: > the main on is a bit down the line. Distinguishing what to display as the top asset in the hierachy widget appearing under the tabs. If you have old style silos, the top asset in the tree view is the one with visualParent:None, if you're however using that as tab , it should be the next one in the hierarchy. So if you allow mixed type in the project, it goes a bit down the rabit hole. Isn't it more trouble than it's worth?

I see what you mean - but in essence only the data differs but with the code it should still behave exactly the same to the end user. It shouldn't break, even if only later you create the actual parent asset for it. :) The only danger might be that you have a `film` asset that is not supposed to be a silo. So say you have asset "film" with its silo set to "film". ;)

Basically, in backwards compatibility all it means is that:
- When something has a silo set and *no* visual parent. It's parented under a silo tab.
- When something has *no* silo set and *no* visual parent. It is the silo asset itself - it is the tab.
- Any asset with a visualParent is never a silo and always only visible under the parent.

---

- Odd case: When something has a silo set and *no* visual parent but there is an asset for that silo then it is actually "parented" to the tab - it just doesn't have the `visualParent` data set to it. In the data it might look funny, but functionally it's the same as the above cases.

mottosso:
Happy to pull the trigger on this to keep the ball rolling.

mkolar:
> Happy to pull the trigger on this to keep the ball rolling.

Same for me. Just a confirmation that it doesn't currently break anything in their current project from @BigRoy would be good

BigRoy:
Sorry, unable to test right away - but possibly near end of day or tomorrow. Bit crazy days here. Am keeping an eye on this though, hopefully soon I can join the Russian Roulette.

avalon/tools/projectmanager/widget.py
```diff
@@ -536,13 +564,24 @@ def refresh(self):
 
         silos = _list_project_silos()
         self.silo.set_silos(silos)
-
+        if len(silos) > 0:
+            self.silo.set_current_silo(self.silo.tabText(0))
```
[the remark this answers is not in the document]
iLLiCiTiT: there's missing comment `# set first silo as active so tasks are shown on load`

mkolar:
after large cleanup, we decided to make this PR again, but a bit cleaner. to properly get rid of silos, we first need to deal with restructuring GUIs to more flexible organisation and remove silos after, to avoid any ugly surprises. We'll PR fresh very soon. 

functionality will stay the same.

mkolar:
I'm closing this PR as it is not relevant anymore. We'll follow up with a much simpler and backwards compatible one.

PR #379, opened 2019-03-29
Only pop warning when module path not found

davidlatwe:
### Problem

I have some loader plugins that is only for me to diagnose the pipeline health or testing, like printing representation ID.
But those plugins are in a place that only I can have access, which will raise error to others when the Avalon is registering plugin path that could not be found.

### Solution

Instead of raising error, just pop up warning.

### DIFF
--- a/avalon/lib.py
+++ b/avalon/lib.py
@@ -243,7 +243,9 @@ def modules_from_path(path):
 
     path = os.path.normpath(path)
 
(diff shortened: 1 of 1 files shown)


mottosso:
That makes sense.. wonder why it hasn't been a problem until now. It should act like PYTHONPATH, and it also ignores things like this.

Free to merge!

davidlatwe:
Merge !

Issue #365, opened 2019-01-17
Workfiles: Multiple hosts support


BigRoy:
### Issue

Supporting multiple hosts in the Workfiles app was originally laid out in the code, but relatively spread across the `app.py` file. Making it:

- hard to find the exact implementation details needed for a host.
- and continuously harder as the amount of hosts would grow.

Now I've done some ground work on support Houdini + Fusion, see [this commit](https://github.com/BigRoy/core/commit/241864fa5b82fa1348540679bed2618038356745) and wanted to discuss on the implementation of the `Host` class I set up solely for Workfiles tool.

The issue I'm seeing is that it mimics (it being called Host) the `avalon.maya`, `avalon.fusion`, `avalon.houdini` the basic host integrations that Avalon already has - yet they don't require the `save`, `open`, etc. API to be implemented for a Host integration to work with Avalon - just the Workfiles app wouldn't work.

### Discussion

We can go about this in two ways:
- Continue down the route of [this commit](https://github.com/BigRoy/core/commit/241864fa5b82fa1348540679bed2618038356745) and keep save/open/current-file and similar functions there solely for Workfiles for simplicity even though it might introduce some confusion for newcomers.
- Move these API methods to the Avalon globally supported hosts and have these methods available in their api.

The methods currently required for Workfiles can be found [here](https://github.com/BigRoy/core/blob/241864fa5b82fa1348540679bed2618038356745/avalon/tools/workfiles/host.py#L5) and comes down to: `save`, `open`, `current_file` and `root`. Their docstrings should elaborate somewhat on what they intend to do.

---

Referenced on Gitter: https://gitter.im/getavalon/Lobby?at=5c403bcdcb47ec3000586afe

davidlatwe:
I would vote for option 2 : *Move these API methods to the Avalon globally supported hosts and have these methods available in their api.*

Like, for example, having a sub-module `workio.py` in every host module.

BigRoy:
> Is that what you're suggesting here, @BigRoy?

Yes, the second way I proposed indeed means removing this `Host` class that I now implemented with [this commit](https://github.com/BigRoy/core/commit/241864fa5b82fa1348540679bed2618038356745) and move those methods to `avalon.maya`, `avalon.houdini`, etc. (In that commit I already separated it purely away from `avalon.tools.workfiles.app` into its own file.

BigRoy:
**Refactor to avalon registered host**

[...]

PR #382, opened 2019-04-27
Expose "open_latest_workfile" method for public use.

tokejepsen:
**Goal**

To enable developers to open the latest workfile on launch of application.

**Motivation**

Instead of hacking a solution together using the methods inside of the workfiles app, thought it better to expose a dedicated method that might benefit others.

### DIFF
--- a/avalon/tools/workfiles/README.md
+++ b/avalon/tools/workfiles/README.md
@@ -21,6 +21,18 @@ workfiles.show(
 )
 ```
 
+Or open the latest workfile.
+
+```python
+workfiles.open_latest_workfile(
+    os.path.join(
+        cmds.workspace(query=True, rootDirectory=True),
+        cmds.workspace(fileRuleEntry="scene")
+    )
+)
+# Returns None if no latest file were found.
+```
+
 ## Naming Files
 
 Workfiles app enables user to easily save and create new work files.
--- a/avalon/tools/workfiles/__init__.py
+++ b/avalon/tools/workfiles/__init__.py
@@ -1,7 +1,9 @@
 from .app import (
-    show
+    show,
+    open_latest_workfile
 )
 
 __all__ = [
-    "show"
+    "show",
+    "open_latest_workfile"
 ]
--- a/avalon/tools/workfiles/app.py
+++ b/avalon/tools/workfiles/app.py
@@ -322,7 +322,7 @@ class Window(QtWidgets.QDialog):
         if not float_list:
             return None
 
-        return data[str(max())]
+        return data[str(max(float_list))]
 
     def refresh(self):
         self.list.clear()
@@ -370,11 +370,10 @@ class Window(QtWidgets.QDialog):
         else:
             return None
 
-    def open_maya(self, file_path):
+    def open_maya(self, file_path, force):
         from maya import cmds
 
-        force = False
-        if cmds.file(q=True, modified=True):
+        if cmds.file(q=True, modified=True) and not force:
             result = self.save_changes_prompt()
 
             if result is None:
@@ -389,14 +388,14 @@ class Window(QtWidgets.QDialog):
 
         return True
 
-    def open(self, file_path):
+    def open(self, file_path, force):
         func = {"maya": self.open_maya}
 
         work_file = os.path.join(
             self.root, self.list.selectedItems()[0].text()
         )
 
-        return func[self.application](work_file)
+        return func[self.application](work_file, force)
 
     def on_duplicate_pressed(self):
         work_file = self.get_name()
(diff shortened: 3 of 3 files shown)


BigRoy:
Honestly, I'm not a fan... it's a bit too high-level without lower level being publicly available to. In my repo I have a setup where the work files api is exposed per host and I'd recommend first going that way, where it does:

```
#psuedocode
import avalon.api

host = api.registered_host ()
# understand where the work files are, dependent on software plus pipeline?
path = avalon.api.get_workfiles()[-1]

host.open (path)
```
Or something along those lines? From a dev perapective it just makes more sense to expose the files from the workfiles easily and how to open those in the host aside of only exposing to open the latest work file?

tokejepsen:
> Or something along those lines?

Sure yeah, that sounds great. Just been catering for my own needs atm, which it to open the latest workfile on application launch, so any improvements to that is good.

More specifically I get why the ```host``` should live in the Avalon API, but should the ```get_workfiles``` not live in the workfiles module?

mottosso:
>More specifically I get why the host should live in the Avalon API, but should the get_workfiles not live in the workfiles module? 

Ah, didn't spot you were exposing it via `workfiles`. I think we should keep any external access to Avalon accessible via `api` rather than via the package `__init__`. Anything in `api` is expected to be well maintained and last forever (so take care!), and anything not in `api` can't be relied upon for external use.

What Avalon does *internally* however is nobodies business, but it seems like you expect external users and developers building tools on-top of Avalon to use this, in which case api is the place to be.

tokejepsen:
> Anything in api is expected to be well maintained and last forever (so take care!), and anything not in api can't be relied upon for external use.

Good point. Will update this PR with the changes.

tokejepsen:
Actually on that point, should the tools be exposed in the API as well?

tokejepsen:
@BigRoy I agree with this method, but it seems like we should get https://github.com/getavalon/core/issues/365 sorted first then?

BigRoy:
Once #387 is merged likely it's best to close off this PR. However, @tokejepsen please still have a critical eye at whether the availability of "getting the last workfile" is simple enough or we still need to explicitly expose something to the core api.

mottosso:
With #387 merged, did we still think this should not get merged?

I agree with Roy that it's a little too high-level, even magical. Maybe this is better suited as a sorting function to the workfiles model/view itself? Such that you can click "Open Latest", as in, whichever is last in the list. Then sorting is both visible to the user and customisable, either by custom function implemented in a config, or by the user clicking on a relevant section of the view, like ascending modified date.

tokejepsen:
As stated the goal was for developers to launch the latest workfile for users instead of presenting the workfiles app on launch. One button click less and no waiting for the workfiles app.
The workfiles app already preselects the last modified file in the list.

I think all the talk about API exposure needs to be settled before this one can continue.

mottosso:
> One button click less and no waiting for the workfiles app.

Can you provide a usecase and/or example of how you envision the function to be used? I thought it was meant for either this GUI, or someone elses GUI, but I'm not seeing how that would reduce the number of clicks.

tokejepsen:
> Can you provide a usecase and/or example of how you envision the function to be used? I thought it was meant for either this GUI, or someone elses GUI, but I'm not seeing how that would reduce the number of clicks.

The idea was to open the latest workfile on launch on the application, rather than launching the application then open the latest workfile.
Launching either using Avalon Launcher or Ftrack.

mottosso:
> The idea was to open the latest workfile on launch on the application, rather than launching the application then open the latest workfile.

Ahaaa. But isn't that a little invasive? What if you wanted to open Maya without automatically loading the latest file? What if the file is huge and takes several minutes?

The idea is sound, how about we combine this with #412 and have the Context Manager show up on Maya launch instead?

tokejepsen:
> Ahaaa. But isn't that a little invasive? What if you wanted to open Maya without automatically loading the latest file? What if the file is huge and takes several minutes?

This is subjective, so would be implemented per config, and hopefully the config will also make it project based.

mottosso:
Ok, I see what you mean. You want this method exposed for external configurations to Avalon. I can see how that makes sense, and how other common convenience functions have a place in Avalon too.

Because there's no limit to how much convenience we can bake into Avalon, I'd like to make a strict separation between what is convenience and what is core functionality. Convenience being built on-top of core. David's suggestion to add a `avalon.util` seems a good place for this, something we can add to freely and something that can grow, and worst case require a manual implementation using core features if convenience breaks. Something not bound by the "everlasting support" of the public API.

- https://github.com/getavalon/core/issues/425

Let's move this PR into a new `util.py` module, would you like to take the lead on that @tokejepsen?

tokejepsen:
> Let's move this PR into a new util.py module, would you like to take the lead on that @tokejepsen?

My time schedule means I wont be able to look at this for weeks, but yeah.

PR #387, opened 2019-05-17
Workfiles API for Hosts

BigRoy:
It seems my branch is ahead or differs from main getavalon/core here and there so this PR might be a bit messy. To see the clean changes that need to be made for this to actually work, see this PR: https://github.com/Colorbleed/core/pull/42

_I will investigate the rest of the differences here and comment on this PR_

---

## Implement Workfiles API

This implements a Workfiles API for the hosts.

- [How to implement a Workfiles API for a host is documented in the README of the Workfiles tool](https://github.com/BigRoy/core/tree/workfiles/avalon/tools/workfiles#implementing-a-new-host-integration-for-work-files)
- Note: Nuke is still missing its Workfiles API here.

#### Example implementations

- https://github.com/BigRoy/core/blob/workfiles/avalon/fusion/workio.py
- https://github.com/BigRoy/core/blob/workfiles/avalon/houdini/workio.py
- https://github.com/BigRoy/core/blob/workfiles/avalon/maya/workio.py

#### References

- Issue: https://github.com/getavalon/core/issues/365
- Gitter discussion: https://gitter.im/getavalon/Lobby?at=5cdd928583ae782aee7b3046


avalon/houdini/pipeline.py
```diff
@@ -20,7 +20,7 @@
 self._parent = None
 self._events = dict()
 
-AVALON_CONTAINERS = "AVALON_CONTAINERS"
+AVALON_CONTAINERS = "/obj/AVALON_CONTAINERS"
```
BigRoy: This was a fix for where Houdini sometimes would do a "relative query" without an absolute path and would fail to resolve the container in some contexts.

avalon/houdini/pipeline.py
```diff
@@ -175,16 +172,18 @@ def containerise(name,
 
     """
 
-    # Get main object network node
-    obj_network = hou.node("/obj")
+    # Ensure AVALON_CONTAINERS subnet exists
```
BigRoy: Refactored code to take in account the new non-relative path of `AVALON_CONTAINERS` - also matched naming of containers somewhat more with how Maya was doing things. _Basically solely cosmetic changes_

avalon/houdini/pipeline.py
```diff
@@ -304,7 +302,7 @@ def process(self):
             node_type = "geometry"
 
         # Get out node
-        out = hou.node("out")
+        out = hou.node("/out")
```
BigRoy: Again, made sure path wasn't fetched relative and then in some contexts failed to work - this fixes it.

avalon/houdini/pipeline.py
```diff
@@ -313,16 +311,15 @@ def process(self):
...
-    api.emit("open", *[])
-
-
-def _on_scene_new(*args):
-    api.emit("new", *[])
-
-
-def _on_scene_save(*args):
-    api.emit("save", *[])
+def on_file_event_callback(event):
```
BigRoy: Actually made Houdini callbacks work - wait... aren't these already in core master?

avalon/maya/lib.py
```diff
@@ -115,11 +115,19 @@ def export_alembic(nodes,
...
     Arguments:
         nodes (list): Long names of nodes to cache
+
         file (str): Absolute path to output destination
+
         frame_range (tuple, optional): Start- and end-frame of cache,
             default to current animation range.
-        uv_write (bool, optional): Whether or not to include UVs,
+
+        write_uv (bool, optional): Whether or not to include UVs,
```
BigRoy: Fixed docstring. - Cosmetics mainly.

avalon/maya/lib.py
```diff
@@ -428,7 +447,10 @@ def lsattr(attr, value=None):
     """
 
     if value is None:
-        return cmds.ls("*.%s" % attr)
+        return cmds.ls("*.%s" % attr,
```
BigRoy: Ensure that when querying without a value the result actually returns the object's node name _without!_ the attribute in it, so it's consistent with how ` lsattrs` works with values.

avalon/maya/lib.py
```diff
@@ -439,12 +461,13 @@ def lsattrs(attrs):
         attrs (dict): Name and value pairs of expected matches
 
     Example:
-        >> lsattr("age")  # Return nodes with attribute `age`
-        >> lsattr({"age": 5})  # Return nodes with an `age` of 5
-        >> # Return nodes with both `age` and `color` of 5 and blue
-        >> lsattr({"age": 5, "color": "blue"})
+        >> # Return nodes with an `age` of five.
+        >> lsattr({"age": "five"})
```
BigRoy: Fixed docstring since function itself actually only worked on string attributes.

avalon/maya/pipeline.py
```diff
@@ -500,11 +500,12 @@ class Creator(api.Creator):
     def process(self):
         nodes = list()
 
-        if (self.options or {}).get("useSelection"):
-            nodes = cmds.ls(selection=True)
+        with lib.undo_chunk():
```
BigRoy: Ensure undoing in Maya actually undoes the full instance creation instead of the attributes separately from the node creation, etc. - this fixes undoing potentially leaving "incorrect" instances around that don't have all attributes.

avalon/tools/creator/app.py
```diff
@@ -299,22 +299,19 @@ def on_create(self):
...
 
         item = listing.currentItem()
-        useselection_chk = self.data["Use Selection Checkbox"]
-
-        if item is not None:
-            subset_name = result.text()
-            asset = asset.text()
-            family = item.data(FamilyRole)
-        else:
+        if item is None:
```
BigRoy: Purely cosmetic changes for readability. (somewhat subjective)

avalon/tools/creator/app.py
```diff
@@ -324,6 +321,8 @@ def on_create(self):
             self.echo("Program error: %s" % str(e))
             raise
 
+        self.echo("Created %s .." % subset_name)
```
BigRoy: In the Creator UI show that the creation has actually happened - in some applications this can be hard to notice, e.g. in Houdini you might not be in the context where you will see the creation happen and will be left wondering whether the UI actually did anything.

avalon/tools/projectmanager/widget.py
```diff
@@ -138,7 +138,8 @@ def preserve_selection(tree_view,
                 selection_model.select(index, flags)
 
             if current_index_value and value == current_index_value:
-                tree_view.setCurrentIndex(index)
+                selection_model.setCurrentIndex(index,
```
BigRoy: Actually correctly set the current index - not entirely sure why this was needed anymore or where the behavior felt off. I believe on creating assets previously it would change the current index you were working on - this fixed that I believe.

BigRoy:
See [files changed overview](https://github.com/getavalon/core/pull/387/files) for my comments on the "other" commits that are in this branch. Sorry for the mess. :)

avalon/maya/__init__.py
```diff
@@ -26,6 +26,15 @@
 
 )
 
+from .workio import (
```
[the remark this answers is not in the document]
BigRoy: This was purely to expose it for the API. Will add it to `__all__` too like the others so it will shush the hound.

avalon/fusion/workio.py
```diff
@@ -0,0 +1,44 @@
...
+    if not current_filepath:
+        return None
+
+    return current_filepath
+
+
+def work_root():
+    from avalon import api
+
+    return os.path.join(api.Session["AVALON_WORKDIR"], "scenes")
```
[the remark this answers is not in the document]
BigRoy: Interesting. I recall having tested this one in Fusion - but maybe one last refactor I did afterwards and I missed this. Thanks hound! _Will fix_

BigRoy:
#### About the conflicting files
```
avalon/houdini/pipeline.py
avalon/maya/pipeline.py
```

So the conflict in this PR seems to be on a change from `self.name` to `self.data["subset"]` in the creation of instances for the node names - which has been discussed [that `self.name` was actually the preferred pick here](https://github.com/getavalon/core/pull/362#issuecomment-469616712)

To resolve the conflict thus the best pick would be go for `self.name` - which is interesting to see that it wasn't actually resolved in avalon/core (yet). Let's also make sure the Nuke and Fusion integration to that too while we're at it.

davidlatwe:
> To resolve the conflict thus the best pick would be go for self.name - which is interesting to see that it wasn't actually resolved in avalon/core (yet).

Oh, nooo.. I forgot this one :(
Looks like if the conflict get resolved here and we are safe ? :thinking:

BigRoy:
Fixed the conflicts as described above. 

> Looks like if the conflict get resolved here and we are safe ? 

Yup! :)

BigRoy:
@tokejepsen have you been able to play with this yet? :)

tokejepsen:
Not yet :( Hope to get onto pipeline next week/month.

tokejepsen:
Seems to work fine here. Only tested Maya though as I don't have Houdini or know Fusion.

 Happy to merge this :)

BigRoy:
@davidlatwe is this something you can test too? Just want to make sure things behave as everyone wants at this stage. :)

davidlatwe:
Yes, for sure :)
I was planning to do the test this weekend.
I can test on Maya and Houdini tomorrow.

davidlatwe:
Tested, working good on both Maya and Houdini :D

One thing though, I did not set any `default_dirs` for Houdini, so at first it gives me an error saying that `path/to/houdini-workspace/scenes` dir not exists, I think this worth be mentioned in workfile's README ?
Or auto `makedirs` if not exists ?

BigRoy:
> One thing though, I did not set any default_dirs for Houdini, so at first it gives me an error saying that path/to/houdini-workspace/scenes dir not exists, I think this worth be mentioned in workfile's README ?
Or auto makedirs if not exists ?

Ah - wait. That's definitely a good point. I'm wondering if there's a more customizable approach to this for a studio, without altering avalon core... any ideas? Maya has the workspace you can customize. But what would Houdini have? Or even Fusion, or Nuke? Should we have an environment variable for applications that don't have a default fallback for "workspace customization"?

tokejepsen:
Maybe it should use the first entry in the ```default_dirs``` if it exists?

Could also have a workfiles_dir entry in the toml files?

BigRoy:
> Maybe it should use the first entry in the default_dirs if it exists?
> Could also have a workfiles_dir entry in the toml files?

I believe inside the Host the `.toml` file itself isn't readily accessible - or at least it's non-trivial. So maybe it's best to have lik a `AVALON_SCENEDIR` kind of environment variable so that hosts that do not have their own workspace management that this *scene dir* name gets appended to the `AVALON_WORKDIR` path for work files to be saved into.

```
set AVALON_WORKDIR=houdini
set AVALON_SCENEDIR=scenes
```

Makes `houdini/scenes`. When `AVALON_SCENEDIR` is not set then Workfiles will operate directly in the root of the avalon workdir?

tokejepsen:
> Makes houdini/scenes. When AVALON_SCENEDIR is not set then Workfiles will operate directly in the root of the avalon workdir?

Sounds good to me.

davidlatwe:
Sounds good to me, too :D

tokejepsen:
Nobody wants to hit the merge button ? :)

@Bigroy are you happy to merge this?

BigRoy:
Sorry, I'm a bit swamped with other things.

The only remainder is still: https://github.com/getavalon/core/pull/387#issuecomment-500879708 specifically the `AVALON_SCENEDIR`

BigRoy:
Implemented `AVALON_SCENEDIR` to define the scenes directory for hosts that do not have their own mechanism. So Fusion and Houdini now fall back to using that variable. Maya still uses its own `workspace` mechanism to define where to save scene files.

It's important to note that when the scenes directory does *not* exist the Work Files tool will still fail to launch. When using a custom folder name in your own pipeline, please make sure the folder is created prior it, e.g. through the use of `default_dirs` in Application `.toml` files.

davidlatwe:
Looks nice ! Will have a test later today !

tokejepsen:
How about we use ```AVALON_WORKDIR``` for all apps?

We could essentially get rid of the ```AVALON_SCENEDIR``` and just rely on a single environment variables for all apps.

BigRoy:
> How about we use AVALON_WORKDIR for all apps?
>
> We could essentially get rid of the AVALON_SCENEDIR and just rely on a single environment variables for all apps.

That way one could never set the "root" folder in `AVALON_WORKDIR` where you might also save images/references/other content for a specific shot. It really is the root of the asset's work directory... not necessarily where the scenes get saved. So, what you are saying is to always save the files in the root of `AVALON_WORKDIR` - which I think wouldn't be the best way...

And if we'd never have the "project root" then we'd miss out on that.

In a way, it's similar to Maya's workspaces like:

```python
cmds.workspace(query=True, rootDirectory=True) == AVALON_WORKDIR
cmds.workspace(fileRuleEntry="scene") == AVALON_SCENEDIR
```

BigRoy:
> How about just unifying using AVALON_WORKDIR in all apps?

We could, but..

1. I'm just afraid of backwards incompatibility for studios operating on Maya for which workfiles currently works as they'd expect since it relies on Maya's workspace system.
2. Personal preference. ;)

tokejepsen:
> I'm just afraid of backwards incompatibility for studios operating on Maya for which workfiles currently works as they'd expect since it relies on Maya's workspace system.

Good point! Continue as is :)

BigRoy:
@davidlatwe once you have tested it feel free to merge.

davidlatwe:
I have tested in Houdini and Maya, and they both worked smoothly :D
Merging this !

PR #388, opened 2019-06-04
Fix #364: Make cbloader and cbsceneinventory the new defaults 

### LINKED ISSUE #364: Remove old loader/manager and cleanup dead code
### Issue

Currently the default Loader and Manager you get when running avalon is the old and currently unmaintained one. The current updates being over respectively a year ago and 11 months ago - I believe no one is currently using them anymore.

The new tools however got prefixed `cb` as they were originally a port from the older pipeline at Colorbleed.

To actually run the maintained, up-to-date and new tools newcomers must set the [`AVALON_EARLY_ADOPTER`](https://github.com/getavalon/core/blob/5f552efdc3f5cdf02c320a4aef2a02531444d498/avalon/io.py#L164) environment variable to get the new tools (at least in Maya, the other hosts default to the new one).

_This issue is part of simplifying getting started with Avalon and even some development as old dead code would be removed._

### Proposal

1. Remove the old tools, also remove `avalon.tools.build` which is empty.
2. I believe the new tools should be the standard tools to use.
3. Refactor the tools to remove the `cb` prefix.
4. Remove any fallback to the older loader (since we would remove it)

We need to keep in mind backwards compatibility. Allowing to replace the tool with the new one should cover the Artist use cases, as they will have the same functionality - only through the new standard and updated UI. However, potentially a studio could have written custom code or tools referring to `cbsceneinventory` which should then map to `sceneinventory`. However, since this is not part of the front-facing API I'd say this shouldn't be too big of an issue.

---

Please provide feedback on whether this would be good, why or why not?

BigRoy:
Changes:
- Remove empty builder tool
- Remove old manager and loader, replace with cbloader and cbsceneinventory
- Refactor cbloader -> loader
- Refactor cbsceneinventory -> sceneinventory

---

Reference issue: #364

**NOTE**
_Currently code is untested, only refactored it for now. This way it also allows others to check their dependencies on it plus potentially discuss the naming/refactoring of "sceneinventory" as opposed to the older "manager"._


avalon/__main__.py
```diff
@@ -29,8 +29,8 @@
         loader.show(debug=True)
 
     elif args.manager:
```
mottosso: Don't forget to rename the argument too.

mottosso:
Looks good, this is the right way forwards.

BigRoy:
@mottosso do you happen to know what kind of drugs the Hound sniffed? :) I'm assuming those lines are `{file}.py:{line}:{column}`? Is it referring to negative columns/characters? Wut?

avalon/__main__.py
BigRoy: Fixed with: https://github.com/getavalon/core/pull/388/commits/bf4c26907522a04ec77274d8f862e853a64f7d6a

mottosso:
> @mottosso do you happen to know what kind of drugs the Hound sniffed? :)

Haha, no idea. Could it be line-endings, Windows versus Linux? Didn't you have a new file-server, could that have had an effect?

BigRoy:
> Haha, no idea. Could it be line-endings, Windows versus Linux? Didn't you have a new file-server, could that have had an effect?

Everything is saved from a Windows workstations. Plus it's the only file showing the odd behavior and it doesn't even show any other changes than one change on line 97. Additionally, checking symbols in the file doesn't show me anything odd other than just line endings everywhere. It's really finding something magically hidden.

BigRoy:
It seems it failed on the backslashes in the docstring. Seems like a bug, anyway... changed it with https://github.com/getavalon/core/pull/388/commits/e2f21d3982d33615064d8e57573dcb5951baae92 and it seems to make it pass. :)

BigRoy:
@davidlatwe @tokejepsen @mkolar - tagging you guys to give this a test run - if you're referencing any of the tools in your own pipeline code note that they might need refactoring with these names changes. As such, if you are using these tools that are **NOT** public in Avalon API this could introduce a backwards incompatibility and will require you to update code along.

That is also described in this PR's issue #388

davidlatwe:
I have tested on my end and looks working nicely after I changed those module name :)
And I also made a PR in Launcher getavalon/launcher#42 for this to working.

BigRoy:
> And I also made a PR in Launcher getavalon/launcher#42 for this to working.

Thanks, totally forgot that action was native to the launcher currently.

BigRoy:
> And I also made a PR in Launcher getavalon/launcher#42 for this to working.

Thanks, totally forgot that action was native to the launcher currently.

mkolar:
This does not affect us much. We're actually preparing (and should be done this week @iLLiCiTiT? ) a refactor of the UIs which separates models, delegates and widgets from their tools into universally usable elements for UI building in avalon. As we worked with the tools it became a bit of a mishmash of tools cross-referencing widgets from other tools, so this is an effort to make it cleaner... #upcoming

BigRoy:
> a refactor of the UIs which separates models, delegates and widgets from their tools into universally usable elements for UI building in avalon. As we worked with the tools it became a bit of a mishmash of tools cross-referencing widgets from other tools, so this is an effort to make it cleaner... #upcoming

Perfect! :)

PR #391, opened 2019-06-12
Loader GUI subset grouping

davidlatwe:
### Feature - Visually grouping subsets in Loader GUI

loader-subset

This grouping feature is on subset level, and the group name is from the subset document's `data.subsetGroup` field.
So to trigger subset grouping, you need to inject `subsetGroup` entry to the subset document data during publish. Or maybe another little tool for it ?

Welcome to give any inputs :)

### Update

See comment [below](https://github.com/getavalon/core/pull/391#issuecomment-501718754)

### DIFF
--- a/avalon/tools/cbloader/model.py
+++ b/avalon/tools/cbloader/model.py
@@ -25,7 +25,11 @@ class SubsetsModel(TreeModel):
         super(SubsetsModel, self).__init__(parent=parent)
         self._asset_id = None
         self._icons = {"subset": qta.icon("fa.file-o",
-                                          color=style.colors.default)}
+                                          color=style.colors.default),
+                       "group": qta.icon("fa.object-group",
+                                         color=style.colors.default),
+                       "grouped": qta.icon("fa.file",
+                                           color=style.colors.default)}
 
     def set_asset(self, asset_id):
         self._asset_id = asset_id
@@ -104,9 +108,24 @@ class SubsetsModel(TreeModel):
             self.endResetModel()
             return
 
-        row = 0
-        for subset in io.find({"type": "subset",
-                               "parent": self._asset_id}):
+        filter = {"type": "subset", "parent": self._asset_id}
+
+        # Collect subset groups
+        group_nodes = dict()
+        for group_name in io.distinct("data.subsetGroup", filter):
+            group = Node()
+            data = {
+                "subset": group_name,
+                "isGroup": True,
+            }
+            group.update(data)
+            group_nodes[group_name] = {"node": group,
+                                       "childRow": 0}
+            self.add_child(group)
+
+        # Process subsets
+        row = len(group_nodes)
+        for subset in io.find(filter):
 
             last_version = io.find_one({"type": "version",
                                         "parent": subset["_id"]},
@@ -118,16 +137,28 @@ class SubsetsModel(TreeModel):
             data = subset.copy()
             data["subset"] = data["name"]
 
+            group_name = subset["data"].get("subsetGroup")
+            if group_name:
+                group = group_nodes[group_name]
+                parent = group["node"]
+                parent_index = self.createIndex(0, 0, parent)
+                row_ = group["childRow"]
+                group["childRow"] += 1
+                data["isGrouped"] = True
+            else:
+                parent = None
+                parent_index = QtCore.QModelIndex()
+                row_ = row
+                row += 1
+
             node = Node()
             node.update(data)
 
-            self.add_child(node)
+            self.add_child(node, parent=parent)
 
             # Set the version information
-            index = self.index(row, 0, parent=QtCore.QModelIndex())
+            index = self.index(row_, 0, parent=parent_index)
             self.set_version(index, last_version)
-
-            row += 1
 
         self.endResetModel()
 
@@ -146,7 +177,13 @@ class SubsetsModel(TreeModel):
 
             # Add icon to subset column
             if index.column() == 0:
-                return self._icons["subset"]
+                node = index.internalPointer()
+                if node.get("isGroup"):
+                    return self._icons["group"]
+                elif node.get("isGrouped"):
+                    return self._icons["grouped"]
+                else:
+                    return self._icons["subset"]
 
             # Add icon to family column
             if index.column() == 1:
--- a/avalon/tools/cbloader/model.py
+++ b/avalon/tools/cbloader/model.py
@@ -109,14 +109,14 @@ class SubsetsModel(TreeModel):
                                "parent": self._asset_id}):
 
             last_version = io.find_one({"type": "version",
-                                        "parent": subset['_id']},
+                                        "parent": subset["_id"]},
                                        sort=[("name", -1)])
             if not last_version:
                 # No published version for the subset
                 continue
 
             data = subset.copy()
-            data['subset'] = data['name']
+            data["subset"] = data["name"]
 
             node = Node()
             node.update(data)
@@ -146,7 +146,7 @@ class SubsetsModel(TreeModel):
 
             # Add icon to subset column
             if index.column() == 0:
-                return self._icons['subset']
+                return self._icons["subset"]
 
             # Add icon to family column
             if index.column() == 1:
--- a/avalon/tools/cbloader/model.py
+++ b/avalon/tools/cbloader/model.py
@@ -28,11 +28,11 @@ class SubsetsModel(TreeModel):
         self._asset_id = None
         self._sorter = None
         self._icons = {"subset": qta.icon("fa.file-o",
-                                          color=style.colors.default),
-                       "group": qta.icon("fa.object-group",
-                                         color=style.colors.default),
-                       "grouped": qta.icon("fa.file",
-                                           color=style.colors.default)}
+                                          color=style.colors.default)}
+        # Subset group item's default icon and order
+        self._default_group = {"icon": qta.icon("fa.object-group",
+                                                color=style.colors.default),
+                               "order": 0}
 
     def set_sorter(self, sorter):
         self._sorter = sorter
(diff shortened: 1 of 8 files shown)


BigRoy:
Visually this is great, very very nice work David! :) I will still need to check the code itself though.


#### Grouped icon

Anyway, first a note. The *grouped* entries have `fa.file` as icon as opposed to `fa.file-o` making them stand out a lot and also appear very differently then regular entries, even though they are not necessarily different. How about just toning the color a bit but keeping the same icon? (Like maybe tinted slightly darker). _It's not a big issue, it's just that the full opaque folder icon is so.... full. It really pops out._

I wonder what others think about that.

The group icon however looks totally fine to me. :)

tokejepsen:
> How about just toning the color a bit but keeping the same icon?

Does it even need a different icon?
Personally think the indentation is enough.

davidlatwe:
I made a few changes and updates, including current icon restored. :)

* Subset group could be defined and storing in project document's `config` field
* Groups are sortable and only sort with other groups in descending order

Here's the GIF:

loader-subset-fix

---

### Predefined groups (pretty groups)

Like `families` and `tasks`, this new element is named `groups`  and storing in project `config`.
A `group` has these attributes:
```python
{
    "name": "Trash Bin",  # Must
    "icon": "trash-o",  # Optional
    "color": "#c4cedc",  # Optional, icon color
    "order": -99  # Optional, `float` or `int`, default 0
}
```

### Sorting

Groups in Loader GUI are sorting by the "order" attribute, not by it's name, and only sort with other groups.
Also noted that groups are always sorting in descending order and always be on top of the view.


### Grouping subsets

By injecting the group name into `subsetGroup` entry in the *subset document `data`*, the Loader GUI will create a visual group item to grouping those subsets which has the same group name defined in the `data.subsetGroup`.

If the group name could not be found in the project `config`, it will use default icon and order.
So those predefined groups are just for customizing icons and orders.

BigRoy:
Visually that is so clever - very nice @davidlatwe!

> Like families and tasks, this new element is named groups and storing in project config.

Does this mean the `project` schema needs a new version/update? Or does it still pass validation with this new thing in `config`?

---

##### model.set_sorter()?

The `set_sorter()` method, is that required? Somehow Qt itself knows how to sort by that column on the SortRole even without having this `sorter` stored? So is there a way we could rely on that original behavior?

Like could we add the prefix of the group (0 or 1) to `super(SubsetsModel, self).data(index, role)`? For example:

```python
    return prefix + super(SubsetsModel, self).data(index, role)
```

So that we don't have to somehow go through the "Sorter" to get the right column and query the data of it ourselves. I wonder how Qt knows what column it needs to return in that case? Or is this only the case _because_ you are setting the Sort Role explicitly?

Somehow the assigning of the sorter feels off. Can this work simpler by avoiding it?

davidlatwe:
Sorry, forgot to push the latest commit after leaving the comment 0.0

davidlatwe:
> So that we don't have to somehow go through the "Sorter" to get the right column and query the data of it ourselves. 

Oh no, the reason why I need that was to query for which direction it's sorting ([here](https://github.com/getavalon/core/pull/391/files#diff-2aba52af6f402a3f3e5110176faec69dR225)).
And to provide the right `prefix` for fixing groups on top also not to be affected whether it's ascending or descending.

Just pushed the update, and it's a bit more complex then the time you saw it. :P

davidlatwe:
> Does this mean the project schema needs a new version/update? Or does it still pass validation with this new thing in config?

I haven't test it !
Those data were all directly write into the database currently.
Will have a look tomorrow :)

BigRoy:
> I could not find other way to get the current sortOrder, nor to make some of the item not sortable after the model is set. Also the feature requires not to sort group with other groups by clicking the on columns.
Maybe I missed some keywords ?

I see. I think the idea is to actually adjust the Sorter itself and ignore it inside the model. ;) So the Sorter model decides completely what to do. However, of course the model could include the `order` data so it's collected solely once from the database. (As I'm not sure how the sorter proxy would collect data like that only once - I think it should be able to do so too though)

davidlatwe:
Hehe, I think I've figured out a way to implement this in a cleaner way !
I'll update shortly :)

davidlatwe:
Ok, I believe this is better than previous implementation :)
No weird looking code now.

BigRoy:
Thanks for the update! I think we can slim it down even further, I'll try to pull your version and see if my idea could work out. Not sure if I get to it today though. :)

davidlatwe:
Welcome to give it a try :D

davidlatwe:
> Does this mean the project schema needs a new version/update? Or does it still pass validation with this new thing in config?

Turns out the project config schema indeed requires an update. Added !

davidlatwe:
Added one more commit for command `avalon --save` to include `groups` when saving to database.

davidlatwe:
The test has been fixed :D
@BigRoy , were you able to give it a try ?

davidlatwe:
### Subset grouping dialog

By pressing `Ctrl + G`, Loader GUI will pop-up a dialog for inputting group name and group or ungroup selected subsets.

This provides a convenient way to manage subset groups right in Loader GUI.

loader-subset-grouping

#### Note

* The drop down menu shows groups that were currently existed in subsets and pre-defined groups in project config.
* Dialog will not show if nothing selected.

BigRoy:
> By pressing Ctrl + G, Loader GUI will pop-up a dialog for inputting group name and group or ungroup selected subsets.

Sweet. Nice feature. Only thing I am slightly afraid of is... should anyone be allowed to do this? Or should this be a specific "admin" mode thing? I'm a bit scared of anyone on the team being able to alter the "subsets" in the database like that. 

Any opinions @mottosso @tokejepsen @mkolar?

davidlatwe:
How about adding a validation at [here](https://github.com/getavalon/core/pull/391/files?file-filters%5B%5D=.py#diff-a738abffc1b44618672aed7b4dea1fe2R302), and the validation will be implemented in each avalon-config.

So the code would be like:

```python
def show_grouping_dialog(self):
    subsets = self.data["model"]["subsets"]
    selected = subsets.selected_subsets()
    if not selected:
        self.echo("No selected subset.")
        return

    config = api.registered_config()
    validate = getattr(config, "subset_grouping_validation", None)
    # Also take selected subsets into calculation
    if validate and not validate(selected):
        self.echo("Can not group/ungroup selected subsets.")
        return

    dialog = SubsetGroupingDialog(items=selected, parent=self)
    dialog.grouped.connect(self._assetschanged)
    dialog.show()
```

tokejepsen:
I agree with @BigRoy that it needs to be an admin thing. Should also state clearly why the user cant group/ungroup. Something like ```User validation failed. Contact your administrator.```

Also dont know how common storing variables on the config is.

BigRoy:
> How about adding a validation at here, and the validation will be implemented in each avalon-config.

Preferrably I think, once we go down that line it's time to implement some sort of "authentication level". Maybe Avalon core should just have something in it's api like:

```python
avalon.api.get_user_rights()
```

Or `get_user_level` or `get_user_permissions` or `get_user_role`.
Then the tools in core could just always fall back to that.

For example:
```python
role = api.get_user_role()
if role < api.roles.ADMIN:
    # less than disallow
    return
# allowed
print("YAY!")
```

And we allow for some default roles, like:
```
roles.ADMIN = 5
roles.MANAGER = 4
roles.PRODUCER = 3
roles.ARTIST = 1
```
Etc?

Then any pipeline can set the current permissions, even when using an external "login"... e.g:
```python
api.set_user_role(api.roles.ADMIN)
```

However, this should become it's own "Permissions/User level of control issue" - so let's make sure that's open and an active discussion. Then for now I'm fine with two ways:

- Either include it now (as it is) with this PR and then later patch it to allow "permissions" to be set.
- Or exclude the grouping with CTRL + G for now until we have the "permissions" implemented.

---

Or do we see a simpler way?

tokejepsen:
> Either include it now (as it is) with this PR and then later patch it to allow "permissions" to be set.

I'm thinking this might be the best option.

Further to the implementation of permissions, it could also just be a simple environment variable to enable it.

davidlatwe:
How about option 3:
* Use `AVALON_EARLY_ADOPTER` to enable/disable this *group by `CTRL+G`* feature ?

davidlatwe:
> Preferrably I think, once we go down that line it's time to implement some sort of "authentication level". 

Yes, I think Avalon would need this. Once we implemented this in the api, should make us a bit easier to accept other feature that is convenient but dangers.

BigRoy:
I've started to give this a test run - and it's pretty sweet. Nice work @davidlatwe .

However, I did notice some issues.

- Whenever you "filter" a family and nothing remains inside a Group - the Group remains visible.
- The same happens when you type a filter at the top bar - the empty Groups always remain visible, but should be gone.
- When pressing "CTRL+G" when having the Loader open in Maya actually also runs the shortcut in Maya interface as if the Qt interface does not handle the key event as completed but lets it pass on to Maya.
- Pressing Enter after typing the name for the group does not work. You need to accept it by pressing the button. (A bit annoying)

Additionally, I'd love one other minor feature - the possibility to disable the grouping. I'm thinking of adding a `Group` checkbox to the right of the "Filter Subsets" line edit. What do you think?

Other than that, really amazing work :D

davidlatwe:
Thanks for testing out @BigRoy  !!

I will have a closer look on those issues tomorrow and see if I can resolve them :D

> Additionally, I'd love one other minor feature - the possibility to disable the grouping. I'm thinking of adding a Group checkbox to the right of the "Filter Subsets" line edit. What do you think?

I think this would be useful, let's do this.

davidlatwe:
* Group remain visible in view even all of it's members has been filtered out
    Firstly implemented in 98d2f85.
    Fixed for older (before 5.10) Qt version in b9914db.
    And a bit improvement in 093a0d7.

* "CTRL+G" key event keep passing to parent App
    Fixed in 8b868ef.

* Apply grouping by pressing Enter
    Implemented in 63ac36a.

* The groups are randomly ordered
    Should be fixed after 1af5e8d.

* Column not resized to fit group members' content
    Fixed in 65940c3.
    With this implementation, all columns will auto resized to fit the content. But the size is fixed, user may not change it.

* Auto expand group on text filtering
    Implemented in 66f678c.

* The possibility to disable the grouping
    Implemented in c8a0468.

* Minor fixes
    - faeb6de
    - 3d23f6e

davidlatwe:
Here's a quick preview:

loader-subset-grouping-fix

BigRoy:
That looks amazingly smooth @davidlatwe - perfect work!

avalon/tools/cbloader/model.py
```diff
@@ -21,16 +21,25 @@ class SubsetsModel(TreeModel):
                "handles",
                "step"]
 
-    def __init__(self, parent=None):
+    SortAscendingRole = QtCore.Qt.UserRole + 2
+    sortDescendingRole = QtCore.Qt.UserRole + 3
```
BigRoy: @davidlatwe only note - can we match the capital start? Either both `Sort` or `sort`? :)

davidlatwe: Well spot !! Fixed it :D

BigRoy:
Merge?

davidlatwe:
Merge !

Issue #246, opened 2017-07-24
Store on and lock the of "family" for a "subset"


BigRoy:
# Discussion

We've been having the need to filter for an asset the available "looks" (specifically the latest versions) and list them to the artist. Currently there's no way, by subset, to know whether it contains a specific family so it requires.

To me personally, a subset **is** a specific family, and during its lifetime of existence it shouldn't change to house any versions of a different family. For example subset `lookDefault` would never get a version of family `model` if the first version was a `look` family, I would disallow it changing the family over its lifetime.

# Problem

Currently listing all latest `look` families requires something along the lines of:
```python
subsets = io.find({"type": "subset", "parent": asset_id})
looks = list()
for subset in subsets:
    versions = io.find({"type": "version", "parent": subset['_id']})
    has_look = any(x for x in versions if x['data']['family'] == "look")
    if not has_look:
        continue

   looks.append(subset)

print looks
```
        
Still, this would also allow that the subsets returned _might possibly_ contain different families than a look. So one of the versions in the subset could in  practice be a `camera` family or `model` family.

By locking the subset to a specific family the subset presents itself more clearly to the artist, directly, plus it persists over time. Additionally this simplifies listing the looks for an asset to:

```python
looks = io.find({"family": "look", "type": "subset", "parent": asset_id})
```

What do you think?

mottosso:
Thanks @BigRoy.

I read two problems here.

1. Finding versions by family is hard
2. Families should be stored in `subset`, not version

For (1), you could use the `$in` operator.

```python
# All versions of family "mindbender.rig"
io.find({"type": "version", "data.families": {"$in": ["mindbender.rig"]}})
```

For (2), sure. That makes sense. Could we make it `families` though, as opposed to `family`?

BigRoy:
> For (2), sure. That makes sense. Could we make it families though, as opposed to family?

We could, sure. @aardschok what you think?

aardschok:
Perfect, enables a more broader search too!

mkolar:
For reviews specifically. we add `review` into `families` of the given subset, which triggers review extractor and attaches the quicktime as another representation. 

Across the board we use multiple families on many things during publishing. We then only store the main `family` on the version in the DB as all the additional ones are irrelevant once they've been extracted.

BigRoy:
> Across the board we use multiple families on many things during publishing. We then only store the main family on the version in the DB as all the additional ones are irrelevant once they've been extracted.

[...]

PR #392, opened 2019-06-12
Implement Loader families mask

davidlatwe:
### Families icon color

* Before

    image

* After

    image

This change was aimed to improve the icon's visibility when it's been highlighted in the GUI.

---

### Shadow Families

In some of our subset, may contain more then one family in the `data.families` document entry. For example:
```
[
    "reveries.imgseq",
    "reveries.imgseq.playblast",
]
```
Both family will be shown in Loader GUI's family widget, but only the first family `"reveries.xgen"` will be able to filter out subsets.

So I propose this, by adding a list of `"shadowFamilies"` into project data to mask out those families you might need to hide.

### DIFF
--- a/avalon/fusion/__init__.py
+++ b/avalon/fusion/__init__.py
@@ -15,7 +15,15 @@ from .pipeline import (
     get_current_comp,
     comp_lock_and_undo_chunk
 
+)
 
+from .workio import (
+    open,
+    save,
+    current_file,
+    has_unsaved_changes,
+    file_extensions,
+    work_root
 )
 
 from .lib import (
@@ -32,6 +40,14 @@ __all__ = [
 
     "get_current_comp",
     "comp_lock_and_undo_chunk",
+
+    # Workfiles API
+    "open",
+    "save",
+    "current_file",
+    "has_unsaved_changes",
+    "file_extensions",
+    "work_root",
 
     "maintained_selection"
 
--- a/avalon/fusion/workio.py
+++ b/avalon/fusion/workio.py
@@ -1,0 +1,50 @@
+"""Host API required Work Files tool"""
+import sys
+import os
+
+
+def file_extensions():
+    return [".comp"]
+
+
+def has_unsaved_changes():
+    from avalon.fusion.pipeline import get_current_comp
+
+    comp = get_current_comp()
+    return comp.GetAttrs()["COMPB_Modified"]
+
+
+def save(filepath):
+    from avalon.fusion.pipeline import get_current_comp
+
+    comp = get_current_comp()
+    comp.Save(filepath)
+
+
+def open(filepath):
+    # Hack to get fusion, see avalon.fusion.pipeline.get_current_comp()
+    fusion = getattr(sys.modules["__main__"], "fusion", None)
+
+    return fusion.LoadComp(filepath)
+
+
+def current_file():
+    from avalon.fusion.pipeline import get_current_comp
+
+    comp = get_current_comp()
+    current_filepath = comp.GetAttrs()["COMPS_FileName"]
+    if not current_filepath:
+        return None
+
+    return current_filepath
+
+
+def work_root():
+    from avalon import api
+
+    work_dir = api.Session["AVALON_WORKDIR"]
+    scene_dir = api.Session.get("AVALON_SCENEDIR")
+    if scene_dir:
+        return os.path.join(work_dir, scene_dir)
+    else:
+        return work_dir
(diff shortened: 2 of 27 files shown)


BigRoy:
(Subjective opinion) I'm not too much of a fan of the new icon color, I see why you decided for it against the selected color. But maybe a tad darker blue or lighter blue would already do wonders for this. Another option is having the green less bright/saturated. Also, keep in mind I'm used to the old... and it's always hard to force a new direction because you're dealing with opinions.

> Both family will be shown in Loader GUI's family widget, but only the first family "reveries.xgen" will be able to filter out subsets.

1. This is actually due to another issue, it's the fact that the SubsetWidget in the Loader -> the one showing the publishes you can load. Doesn't actually use the `families`, the same goes for its filter. My proposal would be to don't start using extra data like `shadowFamilies` but instead we somehow adapt the widget to allow filtering by any of the options. As such, when any of the families is toggled off maybe it's hidden?

2. Then if you want to "exclude" one filter to be shown, then perhaps that would be a Project-based or Configuration-based option that can say: "disallow-filter", "hideFilter" or alike. That way it's clear that just the filtering part is not shown. I believe that's somewhat like you've done with `shadowFilters` but admittedly the name kind of confused me. How about `hideFilter`? This could also be an option inside the family itself in the project configuration? Could that work nicely?

For example, families currently support this in project configuration:
```
            {
                "name" : "colorbleed.animation",
                "icon" : "male",
                "label" : "Animation"
            }
```

How about adding support for `"hideFilter": True`?

Somehow the name "shadowFilter" felt unclear to me as to why what it would do.

BigRoy:
_Sorry, pressed the wrong button - did not intend to close_

davidlatwe:
> How about adding support for "hideFilter": True?

Ah, I like this. :)

> adapt the widget to allow filtering by any of the options. As such, when any of the families is toggled off maybe it's hidden?

I did thought about this, works like `pyblish`'s `plugin.match = pyblish.api.Subset`, right ?

For example:

I have a subset like this

```python
{
  "type": "subset",
  "name": "xgenLegacyDefault",
  "families": [
  	"reveries.xgen",
  	"reveries.xgen.legacy",
  ]
}
```

and another like

```python

{
  "type": "subset",
  "name": "xgenInteractiveDefault",
  "families": [
  	"reveries.xgen",
  	"reveries.xgen.interactive",
  ]
}
```

And in the GUI filter widget

* [x] reveries.xgen
* [x] reveries.xgen.legacy
* [x] reveries.xgen.interactive

When I turn off `reveries.xgen.interactive`, only the `xgenInteractiveDefault` subset will be filter out,
but if I turn off `reveries.xgen`, they all go away.


If that's what option 1) is for, I think I like this better :D

davidlatwe:
Oh, one unrelated note for the example I described above, the `families` still storing inside the version document here. Not moving to subset, yet.

BigRoy:
> When I turn off reveries.xgen.interactive, only the xgenInteractiveDefault subset will be filter out,
but if I turn off reveries.xgen, they all go away.

Yes, that's basically what I meant on how it could work if the filtering was aware of the multiple families as opposed to one. ;) Visually the filters could even end up being shown indented, hehe. But that's more cosmetics than needed for it to function.

This functionality could then still of course be combined with `hideFilter` to decrease visual complexity in the filter widget.

> Oh, one unrelated note for the example I described above, the families still storing inside the version document here. Not moving to subset, yet.

Yes. That's correct. That is a different discussion though, see https://github.com/getavalon/core/issues/246

davidlatwe:
### Update

* Reverted the family icon color change. :)

* Dropped the `shadowFamilies` and re-implement the feature by introducing new family property `hideFilter`. Which will hide the family from filter widget if `hideFilter = true`.

* Enabled the family filter to interact with subset's all families, like this:

  loader-families-fix

davidlatwe:
Spot this issue a few days ago :

If one subset/version has family that has been set to `hideFilter = true`, it will not be shown in Loader. 

Example:
```python
# The families of the subset/version
subset_families = set([
    "config.image",
    "config.image.render"
])

# All families in database
existed_families = set([
    "config.image",
    # "config.image.render"  # been set to `hideFilter = true`, will not be collected
    ...
])

VISIBLE = subset_families.issubset(existed_families)
# Since "config.image.render" has been ignored, the `VISIBLE` is `False`
assert not VISIBLE

```

The commit 04b2f6a solved this by filtering out item families that has been ignored, before calculation.

BigRoy:
Quickly gave this a test run on our end. We don't have multiple families per publish yet I think, so regarding the masking I couldn't check much - I just checked for backward compatibility. It didn't break anything... so all good. 

Feel free to merge.

PR #393, opened 2019-06-12
Remove silo create tab from Loader and Context Manager GUI

davidlatwe:
Remove the `+` tab which will pop-up a silo create dialog even you are in the GUI that was not meant to create silo.

image


davidlatwe:
Merging this as well :)

PR #400, opened 2019-07-03
Fix #399: Maya 2019+ hide containers in outliner

### LINKED ISSUE #399: Maya set "Hidden in Outliner" for Avalon containers (Maya 2019+)
### Proposal

The default behavior for Avalon in Maya is to have the `AVALON_CONTAINERS` and the containers themselves visible in the outliner. 

However, since Maya 2018 an attribute `hiddenInOutliner` was introduced... yet Autodesk skipped support for it on objectSets. Now, since Maya 2019 it also supports this feature on `objectSet` nodes.

I'm thinking to, for Maya 2019 and up, by default set the `hiddenInOutliner` attribute to to True so:

- The outliner is slightly less cluttered, no `AVALON_CONTAINERS`
- ObjectSets that are insides the loaded container are not nested in `AVALON_CONTAINERS > container > objectSet_is_here` but show just in the outliner as you'd expect. (biggest plus)
- Less likely someone will ever accidentally delete the `AVALON_CONTAINERS` - never happened for us... but it's possible to do so.

To still debug the avalon containers in the scene one can enable `Display > Ignore "Hidden in Outliner"` in the Outliner to still quickly make them visible.

_Note: This is about *loaded* content/containers and not about publish instances. It's the INCOMING containers._

*This behavior would only work moving forward from Maya 2019 upwards*

BigRoy:
This implements #399 - simple as that.

- Backwards compatible, it first checks whether the attribute exists.
- Will only set the value on newly created containers/sets.

### DIFF
--- a/avalon/maya/pipeline.py
+++ b/avalon/maya/pipeline.py
@@ -406,6 +406,16 @@ def containerise(name,
 
     cmds.sets(container, addElement=main_container)
 
+    # Implement #399: Maya 2019+ hide containers in outliner
+    if cmds.attributeQuery("hiddenInOutliner",
+                           node=container,
+                           exists=True):
+        cmds.setAttr(container + ".hiddenInOutliner", True)
+    if cmds.attributeQuery("hiddenInOutliner",
+                           node=main_container,
+                           exists=True):
+        cmds.setAttr(main_container + ".hiddenInOutliner", True)
+
     return container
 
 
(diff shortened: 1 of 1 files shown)


BigRoy:
Note, this currently does hide the Main container for Maya 2019 scenes when loading a new instance into an existing scene, but not its members making them "all" visible in the outliner by themselves.

I'll change the code so it only hides the main container on initial creation...

davidlatwe:
Sorry, currently not able to access Maya 2019 for giving this a try, hopefully there will be some gif :P

BigRoy:
Here's a little preview.

outliner_hidden

It shows:

- Containers being hidden.
- Whenever a "set" is in the container, it now directly shows in the outliner. Yay!
- You can hide/unhide them via the `Display > Ignore "Hidden in Outliner"` option.

BigRoy:
Will merge this tomorrow if no objections..

Issue #401, opened 2019-07-10
Workfiles: doubleclick opens file.


tokejepsen:
**Goal**

When double-clicking an item in the workfiles app, it'll open. Same as select > "Open".

davidlatwe:
How about adding this one more goal:
* Accept file save by hitting Enter, currently you need to press `Ok` button to save file.

    ![image](https://user-images.githubusercontent.com/3357009/60982859-8170eb00-a36b-11e9-8695-7942e657df2d.png)

BigRoy:
> When double-clicking an item in the workfiles app, it'll open. Same as select > "Open".

Interesting, for me this already works. See:

![workfiles_double_click_gif](https://user-images.githubusercontent.com/2439881/60983268-f297aa80-a339-11e9-8c8f-2ee0869f90e7.gif)

Did I secretly hack this in without merging it?

> Accept file save by hitting Enter, currently you need to press Ok button to save file.

This could be nice - yes.

davidlatwe:
> Did I secretly hack this in without merging it?

I think this feature has been merged, in #387

cameronpd:
Double clicking to open works for me as well.

tokejepsen:
I can confirm this is already implemented as well :)

PR #402, opened 2019-07-16
Workfiles save: Allow enter key to accept the save.

BigRoy:
You can now accept file save by hitting Enter in the Workfiles app.

---

Based on request by @davidlatwe [here](https://github.com/getavalon/core/issues/401#issuecomment-510115040):

> Accept file save by hitting Enter, currently you need to press Ok button to save file.

### DIFF
--- a/avalon/tools/workfiles/app.py
+++ b/avalon/tools/workfiles/app.py
@@ -77,6 +77,10 @@ class NameWindow(QtWidgets.QDialog):
         # Allow "Enter" key to accept the save.
         self.ok_button.setDefault(True)
 
+        # Force default focus to comment, some hosts didn't automatically
+        # apply focus to this line edit (e.g. Houdini)
+        self.comment_lineedit.setFocus()
+
         self.refresh()
 
     def on_version_spinbox_changed(self, value):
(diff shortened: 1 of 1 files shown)


BigRoy:
I'll merge this by tomorrow if no one replies since it's such a minor change.

avalon/tools/workfiles/app.py
```diff
@@ -74,6 +74,9 @@ def __init__(self, root):
         self.ok_button.pressed.connect(self.on_ok_pressed)
         self.cancel_button.pressed.connect(self.on_cancel_pressed)
 
+        # Allow "Enter" key to accept the save.
+        self.ok_button.setDefault(True)
+
```
davidlatwe: I found out that in Houdini, the comment field somehow could not have focus when the dialog pops up.
Maybe we could also add the line
```python
self.comment_lineedit.setFocus()
```
so the UX in both DCC would align :P

BigRoy: Resolved with https://github.com/getavalon/core/pull/402/commits/d5c5d3736ba2cafaa81529d1e4b1c379060cd308

BigRoy:
@davidlatwe feel free to merge :)

PR #403, opened 2019-07-26
Fix asset task view

jasperges:
As discussed on [Gitter](https://gitter.im/getavalon/Lobby?at=5d3aafd2f3e76e13270c5326) this shows only the tasks that were specifically assigned to an asset in the Project Manager.

I also included 2 other (small) things:
- Make the `app` import of the Context Manager relative. Now it also import properly in Python 3.7
- Change the default task icon of the Project Manager to `fa.male`. This is to make it consistent with Avalon Launcher. If people prefer the `folder-o` then maybe we should change the Launcher.

### DIFF
--- a/avalon/tools/contextmanager/__init__.py
+++ b/avalon/tools/contextmanager/__init__.py
@@ -1,3 +1,3 @@
-from app import show
+from .app import show
 
 __all__ = ["show"]
--- a/avalon/tools/projectmanager/model.py
+++ b/avalon/tools/projectmanager/model.py
@@ -191,8 +191,7 @@ class TasksModel(TreeModel):
         super(TasksModel, self).__init__()
         self._num_assets = 0
         self._icons = {
-            "__default__": awesome.icon("fa.folder-o",
-                                        color=style.colors.default)
+            "__default__": awesome.icon("fa.male", color=style.colors.default)
         }
 
         self._get_task_icons()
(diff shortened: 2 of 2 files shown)


avalon/tools/contextmanager/__init__.py
```diff
@@ -1,3 +1,3 @@
-from app import show
+from .app import show
```
mottosso: Oh wow, this was not already the case? Surprised this hasn't broken in some way already.

mottosso:
Leaving this in your hands Roy

avalon/tools/contextmanager/__init__.py
BigRoy: Well spotted, and thanks. I'm quite surprised too. But 🥇!

BigRoy:
> Change the default task icon of the Project Manager to fa.male. This is to make it consistent with Avalon Launcher. If people prefer the folder-o then maybe we should change the Launcher.

Never noticed that difference because we have configured icons apparently for all tasks. :D But yes, clever. Looks good too.

---

Thanks, behavior now makes much more sense again as it's back to the original behavior where it shows no tasks when it has none.

jasperges:
Is this okay to be merged or do more people need to review/approve it first?

BigRoy:
Let's merge it! Thanks again.

PR #413, opened 2019-08-06
fix #406: Unified attributes names

### LINKED ISSUE #406: Unify Attribute names
During a bigger cleanup operation we ran into multiple inconsistencies in avalon-core when  it comes to data member names.

In particular some of the key ones. 

Maya uses `edit_in` and `edit_out` to set frame range, however loader expects `startFrame` and `endFrame`

There are a few key attributes without which many functions are either useless or don't work at all. We'd like to suggest unifying these and making them required on assets and project. On top of that adding functions for their retrieval that will make sure that if the attribute is not filled on the asset, the default value from project will be returned instead.

Based on client feedback and past year working with avalon we suggest these to be the required ones

```
resolutionWidth
resolutionHeight                  

handleStart
handleEnd

frameStart
frameEnd

fps
```

`edit_in` and `edit_out` are problematic, because they overlap with editorial metadata terminology and it became very confusing as soon as we had requests to collect edit specific information like `sourceIn/Out` and `clipIn/Out` 

We'd also like to split handles to `handleStart` and `handleEnd` this is also client request coming from live action VFX show, where this happens quite often. Making this backwards compatible is trivial so I suppose it won't be an issue. 

The questions is whether key attributes like this should remain in `data` or should be promoted to asset directly

iLLiCiTiT:
Changed attribute naming:
- `edit_in`, `startFrame` changed to `frameStart`
- `edit_out`, `endFrame` changed to `frameEnd`
- `resolution_width` changed to `resolutionWidth`
- `resolution_height` changed to `resolutionHeight`
- `handles` can be stored as `handleStart` and `handleEnd`
___
Reference issue: #406 

Everything should stay backwards compatible

### DIFF
--- a/avalon/maya/commands.py
+++ b/avalon/maya/commands.py
@@ -21,10 +21,12 @@ def reset_frame_range():
 
         frame_start = shot["data"].get(
             "frameStart",
+            # backwards compatibility
             shot["data"].get("edit_in")
         )
         frame_end = shot["data"].get(
             "frameEnd",
+            # backwards compatibility
             shot["data"].get("edit_out")
         )
     except KeyError:
@@ -59,16 +61,14 @@ def reset_resolution():
 
     try:
         resolution_width = project["data"].get(
-            "resolutionWidth", project["data"].get(
-                "resolution_width",
-                1920
-            )
+            "resolutionWidth",
+            # backwards compatibility
+            project["data"].get("resolution_width", 1920)
         )
         resolution_height = project["data"].get(
-            "resolutionHeight", project["data"].get(
-                "resolution_height",
-                1080
-            )
+            "resolutionHeight",
+            # backwards compatibility
+            project["data"].get("resolution_height", 1080)
         )
     except KeyError:
         cmds.warning("No resolution information found for %s"
--- a/avalon/maya/commands.py
+++ b/avalon/maya/commands.py
@@ -53,8 +53,18 @@ def reset_resolution():
     project = io.find_one({"type": "project"})
 
     try:
-        resolution_width = project["data"].get("resolution_width", 1920)
-        resolution_height = project["data"].get("resolution_height", 1080)
+        resolution_width = project["data"].get(
+            "resolutionWidth", project["data"].get(
+                "resolution_width",
+                1920
+            )
+        )
+        resolution_height = project["data"].get(
+            "resolutionHeight", project["data"].get(
+                "resolution_height",
+                1080
+            )
+        )
     except KeyError:
         cmds.warning("No resolution information found for %s"
                      % project["name"])
--- a/avalon/maya/commands.py
+++ b/avalon/maya/commands.py
@@ -17,9 +17,11 @@ def reset_frame_range():
     shot = api.Session["AVALON_ASSET"]
     shot = io.find_one({"name": shot, "type": "asset"})
 
+    shot_data = shot["data"]
     try:
-        edit_in = shot["data"]["edit_in"]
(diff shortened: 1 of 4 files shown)


avalon/maya/commands.py
```diff
@@ -18,8 +18,9 @@ def reset_frame_range():
     shot = io.find_one({"name": shot, "type": "asset"})
 
     try:
-        edit_in = shot["data"]["edit_in"]
-        edit_out = shot["data"]["edit_out"]
+
+        frame_start = shot["data"].get("frameStart", shot["data"]["edit_in"])
```
BigRoy: This code is wrong. :)

Unlike an `or` statement I believe Python does not know how to optimize the out the second argument as such the full `shot["data"]["edit_in"]` is always retrieved. As susch, whenever only `frameStart` is in the data this code will still raise a KeyError:
```python
data = {"frameStart": 1}

print data.get("frameStart", data["edit_in"])
# KeyError

print data.get("frameStart", data.get("edit_in", None))
# Yay! :)
```

avalon/tools/cbloader/model.py
```diff
@@ -72,16 +72,23 @@ def set_version(self, index, version):
         version_data = version.get("data", dict())
 
         # Compute frame ranges (if data is present)
-        start = version_data.get("startFrame", None)
-        end = version_data.get("endFrame", None)
-        handles = version_data.get("handles", None)
-        if start is not None and end is not None:
+        frame_start = version_data.get("frameStart", None)
```
BigRoy: This fails to fall back to the original `startFrame` and thus is not backwards compatible.

This should be:
```python
version_data.get("frameStart", 
                 # backwards compatibility
                 version_data.get("startFrame", None))
```

The same goes for `frameEnd` below.

mkolar: true. well spotted.

avalon/maya/commands.py
```diff
@@ -18,8 +18,15 @@ def reset_frame_range():
     shot = io.find_one({"name": shot, "type": "asset"})
 
     try:
-        edit_in = shot["data"]["edit_in"]
-        edit_out = shot["data"]["edit_out"]
+
+        frame_start = shot["data"].get(
+            "frameStart",
+            shot["data"].get("edit_in")
```
BigRoy: Consider this a question, but should we maybe explicitly comment with these fallbacks that they are due to "Backwards compatibility" - I can imagine newcomers or ourselves in 1-2 years looking at this code and be like "why the hell does it also allow `edit_in`?"

I'd rather have it clear that it's only allowing that fallback due to backwards compatibility.

BigRoy:
Note to everyone reading this. If you have Loader plug-ins or other features in your own pipeline relying on start frame or end frame, e.g. a ["Set Time Range" based on published data](https://github.com/Colorbleed/colorbleed-config/blob/13842d313f5ccb4875a0a5c294e117a3b3cba1ca/colorbleed/plugins/maya/load/actions.py#L29) make sure to include the backwards compatibility and gently move forward to the new unified naming conventions. ❤️

avalon/maya/commands.py
iLLiCiTiT: I can really imagine myself in 1-2 weeks looking at this code and be like "who the hell wrote this?" Agree

BigRoy:
Flawless, looking good now. Will try to roll this out tomorrow and test it. :) Nice work!

mkolar:
@mottosso any objections? If not, then this is ready to merge I believe

PR #418, opened 2019-08-20
Fix364

mottosso:
Continuing from https://github.com/getavalon/core/pull/388, for some reason I wasn't able to push directly to that branch.

As this PR merely deletes unused code and that it's been proposed for over two months, I'm merging this now.


BigRoy:
Just a heads up.

This was still remaining as an open PR as it would introduce backwards compatibility issues for studio tools that relied on the widgets/models that were in these tools.

Another reason #414 made *some* sensible choices as we rely a lot on the AssetWidget internally too.

Any way, I'll refactor internal code when I'm back from vacation to what it needs to be by then. ;)

mottosso:
Thanks Roy

mottosso:
For future reference, this is why it's important that external users only reference `api.py` and host-API's. Anything outside of those modules are implementation detail. That will allow us to safely make these kinds of changes, without breaking anything.

If there's something you need that isn't exposed via an API, submit an issue.

For us internally, we can avoid this becoming an issue as well by gracefully deprecating parts of our code. For example, we could reverse the `AVALON_EARLY_ADOPTER` flag, such that it defaults to `cbloader` but can be restored by a temporary variable, e.g. `AVALON_LATE_ADOPTER=true`. That way, we can deploy, people can test, and if it's a problem they won't have to wait for a PR to retain compatibility.

Issue #422, opened 2019-08-21
Resilience against missing paths


mottosso:
**What happened?**
When project is set, but no project files exists, odd errors start appearing.

```py
#   File "C:\Users\marcus\packages\avalon_core\5.3.2\python\avalon\pipeline.py", line 1082, in load
#     return loader.load(context, name, namespace, options)
#   File "<string>", line 124, in load
# RuntimeError: No file name given.
```

On trying to load an asset. (Yes I did give you a file name!)

```py
#   File "<string>", line 29, in copy_path_to_clipboard
#   File "C:\program files\autodesk\maya2018\bin\python27.zip\ntpath.py", line 419, in normpath
# AttributeError: 'NoneType' object has no attribute 'startswith'
```

On trying to copy that asset's path. (I wanted Bruce's model, what is a NoneType??)

**What did you expect?**
A warning or help message about what to do. Avalon can't make assumptions about where files are located, as users may be remote, disks may be unmounted or paths may not be cross-platform. The end-user is able to act on those, but can't do much with a stack trace.

**To Reproduce**
Set these variables, but leave an invalid `AVALON_PROJECTS`.

- `AVALON_PROJECTS`
- `AVALON_PROJECT`
- `AVALON_ASSET`
- `AVALON_TASK`

**Additional context**
I ran into this whilst locally testing Avalon and moving files around on disk. It's not expected to be common, but there isn't much required for it to be handled appropriately.

Also linking to #371 here, as it introduced a mechanism that doesn't return a path unless the path is found on disk; even though a path was available relative the representation/config.

PR #424, opened 2019-08-21
Resilience against missing paths

mottosso:
Related to #422.

This change changes the return value of `avalon.pipeline.get_representation_path()`, which returns a path from (1) representation context, (2) global config or (3) representation data, in that order.

- Previously, if the file didn't exist, it would continue checking the next source.
- Currently, a path is returned regardless of whether it exists on disk.

With this change, it's possible to *attempt* to open a missing path, which is then handled by the tool. It's also possible to copy the proposed path to the clipboard via the Loader, to see what the attempted path actually was.

Before, if a path didn't exist, there was no way of telling why, where it looked and what to do about it.

This still isn't perfect however, there's an architectural flaw; the function is almost able to operate entirely in isolation, given only the data present in the representation. That's easily testable and stable, but then there's is one feature that reaches into the database and breaks testability, which is the `io.parenthood` function.

What we could do instead, is separate the global config from paths stored in the representation. That way a tool can ask "Is there a path in this representation? No? Ok, look globally" and handle each error individually.


mkolar:
So the changed logic now is that if `path_from_representation` was successful (in a sense that it returned something), we won't be checking the next option, but rely on the path returned to be correct.  Right?

I don't see problem with that.  The truth is that if it finds `template` and `context`, fills the path, and that pat doesn't exists it would be good to know about it rather than falling back onto `baked` path right away. That has already caused some minor issues for us in the past. 

Thumbs up from me.

mottosso:
@davidlatwe Please do! Thanks guys.

avalon/pipeline.py
```diff
@@ -1269,16 +1268,14 @@ def path_from_config():
             log.debug("Template references unavailable data: %s" % e)
             return None
 
-        if os.path.exists(path):
-            return os.path.normpath(path)
+        return os.path.normpath(os.path.dirname(path))
```
mkolar: @iLLiCiTiT noticed that this is actually returning only the folder and not the full path. That will break all the loading won't it?

davidlatwe: O.o Good spot !
Looks like something happened while merging #371, but wasn't able to figure out how it changed to return dir path.

mkolar:
I'm wondering why this froze in space a bit. I reckon @mottosso simply didn't find time to change it from directory to the full path? Should we just fix it and merge?

BigRoy:
> Should we just fix it and merge?

I suppose so, yes. 

It could still be worth opening an additional issue for the future about what's stated in the initial comment:

> This still isn't perfect however, there's an architectural flaw; the function is almost able to operate entirely in isolation, given only the data present in the representation. That's easily testable and stable, but then there's is one feature that reaches into the database and breaks testability, which is the io.parenthood function.
>
> What we could do instead, is separate the global config from paths stored in the representation. That way a tool can ask "Is there a path in this representation? No? Ok, look globally" and handle each error individually.

Just to be sure we keep that in mind?

mkolar:
We're happy to fix this, I'm just not sure what is the best way. We can either make new PR that does the same thing without the issue of not returning the full path, or send a PR to @mottosso fork which will trickle down here.

mottosso:
New PR is fine, this one is stale.

PR #429, opened 2019-08-21
Remerge app.py

mottosso:
#388 mistakenly merged the wrong `app.py` for the new loader. See https://github.com/getavalon/core/pull/426#issuecomment-523377949

@davidlatwe At first glance, this appears to be working, however the style doesn't appear to work, and I'm not sure why? :S

**Before**

image

**After**

image

### DIFF
--- a/avalon/tools/loader/app.py
+++ b/avalon/tools/loader/app.py
@@ -1,13 +1,21 @@
 import sys
 import time
 
-from ..projectmanager.widget import AssetWidget, AssetModel
+from ..projectmanager.widget import (
+    AssetWidget,
+    AssetModel,
+    preserve_selection,
+)
 
 from ...vendor.Qt import QtWidgets, QtCore
 from ... import api, io, style
 from .. import lib
 
-from .lib import refresh_family_config, refresh_group_config
+from .lib import (
+    refresh_family_config,
+    refresh_group_config,
+    get_active_group_config,
+)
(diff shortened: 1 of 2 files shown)


mottosso:
These are both running Loader from outside of Maya, e.g.

```bash
python -m avalon --loader
```

davidlatwe:
I just checked-out your branch. and..
No matter I launcher Loader inside or outside of Maya, all failed to reproduce it 0.0

davidlatwe:
The only difference (compare to the commit before 388) I spotted was this [line](https://github.com/getavalon/core/pull/429/files#diff-8659765f202acfc73aaa64f9a5ce2394L40), it's from #393 [here](https://github.com/getavalon/core/pull/393/files#diff-a738abffc1b44618672aed7b4dea1fe2R40), the last missing piece.

```python
- assets = AssetWidget()
+ assets = AssetWidget(silo_creatable=False)
```

mottosso:
Ok, I can confirm this only happens with PySide2-5.13, and does not happen with..

- PySide
- PySide2-5.12
- PyQt4
- PyQt5-5.12

Very strange, but probably nothing we need to worry about.

mottosso:
> The only difference (compare to the commit before 388) I spotted was this line, it's from #393 here, the last missing piece.

Sorry, missed this. Pushed this directly to master just now.

PR #431, opened 2019-08-21
Fix #350, specify JSON schema draft version

### LINKED ISSUE #350: Avalon's schema validation module does not support Draft-6 validation keywords
### Problem

Some of Avalon's schema were using the keyword `const`, which was introduced in JSON schema Draft 6, which require `jsonschema==3.0.0` to work, and the current vendoring version of `jsonschema` in Avalon is version `2.4.0`.

Those not supported keyword will not functioning, which means no `jsonschema.ValidationError` raised.

See the following example:

```python
from avalon.vendor import jsonschema

data = {"foo": "bar"}
schema = {"properties": {"foo": {"const": "foobar"}}}
jsonschema.validate(data, schema)
# No error raised
```
Above example code will raise `jsonschema.ValidationError` if you use the latest version (3.0.0) downloaded from [jsonschema github repo](https://github.com/Julian/jsonschema).

> Note: version 3.0.0 only on github, the latest on `pip` is 2.6.0, which still not support Draft 6.
> [Difference between Draft4 and Draft6](https://json-schema.org/draft-06/json-schema-release-notes.html#q-what-are-the-changes-between-draft-04-and-draft-06)

So, if someone changed the default value which provided by Avalon, and thought it's okay since no error raised, will be an issue in the feature, when Avalon update the `jsonschema` vendor module, someone's pipeline will broke.

For example, in my case, I changed the `container id` from `pyblish.avalon.container` to `pyblish.avalon.sub-container`, for assets like `setdress` type, which has nested container.

### Implementation

Eventually, Avalon will update `jsonschema` vendor module at some point, and when the time comes, we could let TDs to use `Draft4Validator` for the time being and pop warning.

Example code:
```python

import os
import warnings
from avalon.vendor import jsonschema  # assume it's version 3.0


def validate(data, schema):
    SCHEMA_COMPAT = bool(os.environ.get("AVALON_SCHEMA_COMPAT"))

    try:
        jsonschema.validate(data, schema)  # validating with latest Draft
    except jsonschema.ValidationError as e:
        if SCHEMA_COMPAT:
            warnings.warn("Validating with Draft4...")
            validator = jsonschema.Draft4Validator(schema)
            validator.validate(data)
        else:
            raise e
```

davidlatwe:
### Changes

* Specifying our schema draft version to version 4, because the module `jsonschema` we vendorized was only supported up to version 4.

* Change keyword `const` to `enum`, because `const` was introduced in JSON schema Draft 6.

You may see #350 for detail :)

### DIFF
--- a/avalon/schema/application-1.0.json
+++ b/avalon/schema/application-1.0.json
@@ -1,5 +1,5 @@
 {
-    "$schema": "http://json-schema.org/schema#",
+    "$schema": "http://json-schema.org/draft-04/schema#",
 
     "title": "avalon-core:application-1.0",
     "description": "An application definition.",
--- a/avalon/schema/asset-1.0.json
+++ b/avalon/schema/asset-1.0.json
@@ -1,5 +1,5 @@
 {
-    "$schema": "http://json-schema.org/schema#",
+    "$schema": "http://json-schema.org/draft-04/schema#",
 
     "title": "avalon-core:asset-1.0",
     "description": "A unit of data",
--- a/avalon/schema/asset-2.0.json
+++ b/avalon/schema/asset-2.0.json
@@ -20,13 +20,13 @@
         "schema": {
             "description": "Schema identifier for payload",
             "type": "string",
-            "const": "avalon-core:asset-2.0",
+            "enum": ["avalon-core:asset-2.0"],
             "example": "avalon-core:asset-2.0"
         },
         "type": {
             "description": "The type of document",
             "type": "string",
-            "const": "asset",
+            "enum": ["asset"],
             "example": "asset"
         },
         "parent": {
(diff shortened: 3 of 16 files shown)


mottosso:
Nice one! For completeness, our vendored version of `jsonschema` won't actually take this into account, but once/if we update then these constraints should kick in? Or do they work already?

davidlatwe:
Yeah, the keyword `const ` was not working, for example, you could change the container id from `pyblish.avalon.container` into arbitrary string and it still pass the validation.

For example:
```python
from avalon.vendor import jsonschema

data = {"foo": "bar"}
schema = {"properties": {"foo": {"const": "not-bar"}}}
jsonschema.validate(data, schema)
# No error raised
```

mottosso:
Aaah so the current schema was using a too *new* keyword, and this brings it back to the version we're actually using to validate. Gotcha, that makes sense. Happy with this!

PR #433, opened 2019-08-21
Fix #411, change io.save to io.replace_one

### LINKED ISSUE #411: Pymongo collections.save() is deprecated
## Issue

The `avalon.io` module exposes [`save()` from Pymongo Collections, however this seems to be deprecated](https://api.mongodb.com/python/current/api/pymongo/collection.html#pymongo.collection.Collection.save) in favor of [`replace_one`](https://api.mongodb.com/python/current/api/pymongo/collection.html#pymongo.collection.Collection.replace_one) and [`insert_one`](https://api.mongodb.com/python/current/api/pymongo/collection.html#pymongo.collection.Collection.insert_one).

Adding this issue here to ensure we're aware of this deprecation and have any features we implement use `replace_one` and `insert_one` moving forward.

We currently have `replace_one` and `insert_one` exposed via `avalon.io` too, so it's recommended to use those. :)

---

Note that we'll need to swap out `save` in the inventory functions too:
- https://github.com/getavalon/core/blob/a007074c67fffb38e40e829bbd4c266320469853/avalon/inventory.py#L272
- https://github.com/getavalon/core/blob/a007074c67fffb38e40e829bbd4c266320469853/avalon/inventory.py#L302
- https://github.com/getavalon/core/blob/a007074c67fffb38e40e829bbd4c266320469853/avalon/inventory.py#L333

davidlatwe:
Pymongo [`collections.save()`](https://api.mongodb.com/python/current/api/pymongo/collection.html#pymongo.collection.Collection.save) is deprecated since [version 3.0](https://api.mongodb.com/python/current/changelog.html#collection-changes). According to the document, we should change to use `insert_one` or `replace_one`.

* Changed to use `replace_one` since we use `io.save` to overwrite full document.
* Add a deprecation note in `io.save` doc string.

#411

### DIFF
--- a/avalon/inventory.py
+++ b/avalon/inventory.py
@@ -258,7 +258,9 @@ def _save_inventory_1_0(project_name, data):
             print("Separating project metadata: %s" % key)
             metadata[key] = data.pop(key)
 
-    document = io.find_one({"type": "project"})
+    _filter = {"type": "project"}
+
+    document = io.find_one(_filter)
     if document is None:
         print("'%s' not found, creating.." % project_name)
         _id = create_project(project_name)
@@ -269,7 +271,7 @@ def _save_inventory_1_0(project_name, data):
     for key, value in metadata.items():
         document["data"][key] = value
 
-    io.save(document)
+    io.replace_one(_filter, document)
 
     print("Updating assets..")
     added = list()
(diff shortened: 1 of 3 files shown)


mottosso:
Thanks, I was also eyeing this one but wasn't sure what to replace it with. This looks good to me. 👍

davidlatwe:
Oh, wait. I forgot the case that we create new document.
Will patch this later.

davidlatwe:
Ah, no. We are safe to merge this :)
The time we call `replace_one`, documents were already exists in those cases.

davidlatwe:
Will merge this tomorrow if no objections :)

Issue #436, opened 2019-08-23
Refactor tools/


mottosso:
### Goal

Reduce duplication, simplify implementations of tools.

### Motivation

Our "tools" are GUIs written in Qt, and GUIs are a hard problem to solve. There is no PEP 008 for writing graphical applications, and the result is a lack of consistency and foresight.

- There's architecture - "MVC", do we need it, how do we apply it effectively?
- There's structure - pages, panels, widgets and layouts; how do we structure those?
- There's reactivity - which signals should exist and how do we handle them?
- There's communication - which widgets should talk to which, and which should go through a controller?
- There's controllers - What are they, what are their responsibilities?
- There's responsiveness - Anything made asynchronous multiplies every problem by 10x
- There's division of labour - do we create 3 small GUIs that do one thing each well, or 1 GUI that does everything?

Without appropriate answers to these questions, extending or editing a graphical application is a problem made even harder.

### Implementation

Let's see if we can run through these and establish a foundation upon which to build GUIs effectively. I'll write these as unambious as I can, but remember these are guidelines and though some are tried and true, some are not (I'll tag these appropriately).

- [Architecture](#architecture)
- Structure
	- [View](#structureii---view)
	- [Controller](#structure-ii---controller)
	- [Model](#structure-iii---model)
- [Responsiveness](#responsiveness)

#### Architecture

You generally won't need to divide an application into model, view and controller unless it is of a certain size. But as this is an article on consistency, let's move the goal posts to say "No application is small enough to not require MVC".

With MVC as our baseline, here's how to implement it.

| Module | Description
|:-------|:----
| `view.py` | The paintbrush. The "main" of our Window, the cortex of where all widgets come together. This is responsible for drawing to the screen and for converting user interaction to interactions with the "controller"
| `control.py` | The brain. The "main" of our program, independent of a display, independent of the data that it operates on.
| `model.py` | The memory. The container of data.

These three are all that is required to implement an application. Additional modules are for convenience and organisation only, such as..

| Module | Description
|:-------|:---------
| `widgets.py` | Independent widgets too large to fit in `view.py`, yet specific to an application (i.e. not shared with anything else)
| `util.py` | Like widgets, but non-graphical. Standalone timers, text processing, threading or delay functions.
| `delegates.py` | If relevant; these create a dependency between model and view, which typically only communicate through the controller.

**Open Questions**

[...]

PR #437, opened 2019-08-24
Add CONVENTION.md

mottosso:
Related to #436 

This PR contains a set of guidelines for writing graphical tools for Avalon. It's a starting point, along with an example program of what all guidelines look like together. It is meant as a reference for how to write tools that is..

1. Consistent (with the world, but foremost within Avalon)
1. Scalable (without hacks)
2. Resilient to user error (i.e. handle exceptions, provide messages)
3. Resilient to developer error (i.e. work when parts are broken, "hop on one leg")

- [CONTRIBUTING.md](https://github.com/mottosso/avalon-core/blob/convention-guidelines/CONVENTION.md)

The example implements some of the guidelines at the moment, with the idea of being both textual and codified reference of the guidelines (it's a "guideline browser").

It's currently missing:

1. Multiple models, why, and how to deal with that
2. Asynchronousity, threading
3. Reference to which guidelines are implemented where (e.g. `E3`, `S5` etc.)
4. External resources, like images (I noticed Avalon does not yet use any!)
5. CSS

```bash
python -m avalon.tools.convention \path\to\avalon-core\CONVENTION.md
```

image

More examples include..

- [allzpark](https://github.com/mottosso/allzpark/tree/master/allzpark)
- [pyblish lite](https://github.com/pyblish/pyblish-lite/tree/master/pyblish_lite)
- [pyblish qml](https://github.com/pyblish/pyblish-qml/tree/master/pyblish_qml)

Except these guidelines are a superset of what those were built with, an improvement.

As a guideline to the guidelines, let's refer to..

- http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines

For #436, I'd like for these guidelines to apply, but they're not perfect. What I'd like you to do next is have a look at the example program and see how you can improve it. It'll be the baseline for any future application, and what we'll refer back to for guidance on updates to existing GUIs or the creation of new GUIs. It'll be a test-case for the written English, to ensure that everything actually works together and isn't just pretty words. It should be short, yet complete and functional. It doesn't need to accomplish anything useful, it's only purpose is to excercise the guidelines. Similar to how the `pyblish_qml --demo` flag operates, a system stress-test.

#### Missing

How do we deal with multiple models? As mentioned in #436, we'll try and merge the controller (in Allzpark/Pyblish terminology) with the model, making the model both the brain and data container. But the model governs things like row and column count, and different views on different sets of data may/should have different counts. In this case, it's somewhat convenient that the second "view" - the text browser - doesn't have an official "model". But if it did, how can it share a model if it only had a single item, a single row, single column? :S

### DIFF
--- a/CONVENTION.md
+++ b/CONVENTION.md
@@ -1,0 +1,453 @@
+This is a living document under continous improvement. It outlines a set of guidelines for the structure of a typical Avalon graphical user interface.
+
+<br>
+
+### Goal
+
+Improve communication, maintenance and code quality.
+
+<br>
+
+### Example
+
+Each of the below guidelines are implemented in the `convention` GUI.
+
+```bash
+python -m avalon.tools.convention \path\to\avalon-core\CONVENTION.md
+```
+
+![image](https://user-images.githubusercontent.com/2152766/63637288-edd15080-c672-11e9-8697-a7eba1690d57.png)
+
+<br>
+
+### Motivation
+
+Our "tools" are GUIs written in Qt, and GUIs are a hard problem to solve. There is no PEP 008 for writing graphical applications, and the result is a lack of consistency and foresight.
+
+- There's architecture - "MV", do we need it, how do we apply it effectively?
+- There's structure - pages, panels, widgets and layouts; how do we structure those?
+- There's reactivity - which signals should exist and how do we handle them?
+- There's communication - which widgets should talk to which, and which should go through a controller?
+- There's models - What are they, what are their responsibilities?
+- There's responsiveness - Anything made asynchronous multiplies every problem by 10x
+- There's division of labour - do we create 3 small GUIs that do one thing each well, or 1 GUI that does everything?
+
+Without appropriate answers to these questions, extending or editing a graphical application is a problem made even harder.
+
+<br>
+
+### Implementation
+
+Let's see if we can run through these and establish a foundation upon which to build GUIs effectively. I'll write these as unambious as I can, but remember these are guidelines and though some are tried and true, some are not (I'll tag these appropriately).
+
+- [Naming Convention](#naming-convention)
+- [Architecture](#architecture)
+- Structure
+	- [View](#structureii---view)
+	- [Model](#structure-ii---model)
+- [Responsiveness](#responsiveness)
+
+<br>
+
+#### Naming Convention
+
+Python uses snake_case, Qt uses mixedCase; what do we do?
+
+Clearly Qt is being naughty; forwarding the conventions from one language to another. But we can't ignore it, so what are our options?
+
+1. [ ] Stand fast - Stick with snake_case where you can, mixedCase where you must
+2. [ ] Give in - Use mixedCase everywhere
+3. [x] Compromise - Use whichever convention is the majority of code per module or package
+
+##### Stand Fast
+
+This is what I typically do and see.
+
+```py
+from Qt import QtWidgets
+
+class Window(QtWidgets.QDialog):
+	def __init__(self, parent=None):
+		super(Window, self).__init__(parent)
+
+		my_button = QtWidgets.QPushButton("Press me")
+		some_layout = QtWidgets.QHBoxLayout(self)
+		some_layout.addWidget(my_button)
+
+		my_button.clicked.connect(self.on_button_clicked)
+
+	def mousePressEvent(self, event):
+		pass
+
+	def on_button_clicked(self):
+		pass
+```
+
+Which is clearly horrendous. The few occasions where Qt appears to play ball - `clicked` and `connect` - are circumstantial. That isn't snake case, it's mixedCase masquerading as snake_case, ready to jump out at you at a moments notice.
+
+One of the advantages (or so I tell myself) is that it's obvious which methods are overridden, and which are new. But is that enough to justify the overhead in reading ability? It's like reading-an ENGlish `_sen_tenc_e_` with 2 deffirent ConvenTions. Takes you out of the story.
+
+#### Give in
+
+Why not just write everything in mixedCase and get it over with?
+
+If your code consists of only Qt and your own code, that's a fine idea; but standard and third-party libraries don't work that way.
+
+```py
+import pymongo
+
+class Model(QtWidgets.QTableModel):
+	def setData(self, index, value, role):
+		item = pymongo.find_one({"type": "asset"})
+		...
+```
+
+<br>
+
+#### Compromise
+
+Optimise to reduce context switching. Use the majority convention; less granularity is better.
+
+1. Per project
+1. Per package
+2. Per module
+3. Per class
+3. Per method
+3. Per block
+
+For example, if a package (e.g. `avalon.loader`) is mostly Qt, use mixedCase. But if it isn't clear (e.g. `avalon.maya`) use mixedCase for modules where mixedCase is a majority, snake_case otherwise. If it isn't clear *within* a module which is a majority, stay consistent within a class, method or worst case, per block of code.
+
+**Bad**
+
+Sometimes you have no choice but to be consistent per block, but don't do this.
+
+- Each line has its own convention
+- One line has two conventions
+- Context switch count: 4
+
+```py
+class Window(QtWidgets.QDialog):
+	def __init__(self):
+		some_button = QtWidgets.QPushButton("Hello")     # 1
+		myLabel = QtWidgets.QLabel("World")              # 2
+		this_layout = QtWidgets.QVBoxLayout(self)        # 3
+		that_widget = QtWidgets.QWidget()                #
+		otherLayout = QtWidgets.QHBoxLayout(that_widget) # 4
+```
+
+**Good**
+
+Instead do this.
+
+- Context switch count: 2
+
+```py
+class Window(QtWidgets.QDialog):
+	def __init__(self):
+		some_button = QtWidgets.QPushButton("Hello")    # 1
+		this_layout = QtWidgets.QVBoxLayout(self)       #
+
+		thatWidget = QtWidgets.QWidget()                # 2
+		myLabel = QtWidgets.QLabel("World")             #
+		otherLayout = QtWidgets.QHBoxLayout(thatWidget) #
+```
+
+That's 50% less context switching for the same code, however structural consistency is now worsened, as the layout initialisation is now spread across two blocks.
+
+Structural consistency is more important than convention.
+
+<br>
+
+#### Architecture
+
+You generally don't need to divide an application into model and view (MV) unless it is of a certain size. But as this is an article on consistency, let's move the goal posts to say "No application is small enough to not require MV".
+
+With MV as our baseline, here's how to implement it.
+
+| Module | Description
+|:-------|:----
+| `view.py` | The paintbrush. The "main" of our Window, the cortex of where all widgets come together. This is responsible for drawing to the screen and for converting user interaction to interactions with the "controller"
+| `model.py` | The brain. The "main" of our program, independent of a view.
+
+These three are all that is required to implement an application. Additional modules are for convenience and organisation only, such as..
+
+| Module | Description
+|:-------|:---------
+| `widgets.py` | Independent widgets too large to fit in `view.py`, yet specific to an application (i.e. not shared with anything else)
+| `util.py` | Like widgets, but non-graphical. Standalone timers, text processing, threading or delay functions.
+| `delegates.py` | If relevant; these create a dependency between model and view, which typically only communicate through the controller.
+
+<br>
+
+#### Structure I - View
+
+This involves the `def __init__()` in `view.py`, the remaining methods being signal and event handlers.
+
+```py
+class Window(QtWidgets.QMainWindow):
+	title = "My Window"
+
+    def __init__(self, model, parent=None):
+    	super(Window, self).__init__(parent)
+    	self.setWindowTitle(self.title)
+        self.setWindowIcon(QtGui.QIcon(...))
+        self.setAttribute(QtCore.Qt.WA_StyledBackground)
+
+    	pages = {
+    		"home": QtWidgets.QWidget(),
+    	}
+
+    	panels = {
+    		"header": QtWidgets.QWidget(),
+    		"body": QtWidgets.QWidget(),
+    		"footer": QtWidgets.QWidget(),
+    		"sidebar": QtWidgets.QWidget(),
+    	}
+
+    	widgets = {
+    		"pages": QtWidgets.QStackedWidget(),
+    		"logo": QtWidgets.QWidget(),
+    		"okBtn": QtWidgets.QPushButton("Ok"),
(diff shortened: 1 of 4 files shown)


davidlatwe:
> But the model governs things like row and column count, and different views on different sets of data may/should have different counts. In this case, it's somewhat convenient that the second "view" - the text browser - doesn't have an official "model". But if it did, how can it share a model if it only had a single item, a single row, single column?

Happens to have an example on this one :O (should fit this context I think ?)

While I was working on the [Avalon Uploader](https://github.com/davidlatwe/avalon-sftpc) for my studio, I need two `QTreeView` to show different perspective ([staging](https://github.com/davidlatwe/avalon-sftpc/blob/79a65e66406a57c56ffb6b6bf7b1e545f6df0e44/avalon_sftpc/model.py#L377) & [uploading](https://github.com/davidlatwe/avalon-sftpc/blob/79a65e66406a57c56ffb6b6bf7b1e545f6df0e44/avalon_sftpc/model.py#L428)) on [one model](https://github.com/davidlatwe/avalon-sftpc/blob/79a65e66406a57c56ffb6b6bf7b1e545f6df0e44/avalon_sftpc/model.py#L110). And I found that it actually able to let each view has different column count.  Not sure if this implementation more like a hack, at least it looks and works logical to me :P

mottosso:
I also thought of that; of making one proxy model per view. Even those that don't do anything fancy, like sorting of filtering.

```

     ____________________
    |                    |
    |       model        |
    |____________________|
     /        |        \
 ___/____  ___|____  ___\____
|        ||        ||        |
| proxyA || proxyB || proxyC |
|________||________||________|
    |         |         |
 ___|____  ___|____  ___|____
|        ||        ||        |
| viewA  || viewB  || viewC  |
|________||________||________|
```

And I think that makes sense. The proxy would be able to provide more specialised `data()` and `setData`, along with unique columns and such. They may even be able to act as more specialised business logic too? E.g.

```py
editorModel.sortLines()
mainModel = editorModel.sourceModel()
mainModel.sortFiles()
```

davidlatwe:
> The proxy would be able to provide more specialised data() and setData, along with unique columns and such. They may even be able to act as more specialised business logic too?

Implement specialized business logic in proxies to provide processed data by that logic to view make sense to me, too.

But I am not sure about implementing the specialized `setData` in proxies, because isn't it just a proxy for viewing model differently ?

mottosso:
> But I am not sure about implementing the specialized setData

That's a good point.. doesn't feel quite right leaving this imbalance of one having `data` and `setData`, and the other only having `data`, but at first glance it makes sense.

BigRoy:
Sorry for dropping this in here, but can we wait with merging this a day or five? I am back from vacation around Wednesday and hope to have some more time on Thursday to look into this PR and want to make sure I understand what this does and why it's there.

Being on the sidelines and reading only a glimpse of what this is doing my first hunch is to ask the question: "are we overdoing this?" 

Also, the developer guide tool being in `avalon.tools` feels very odd to me? Can we keep that out of core or somewhere into a readme or alike? From the sidelines it feels very forced. Or do I have that feeling as I am missing some crucial part of the conversation? 

To me it doesn't look like it's actually solving anything in practice but just putting in more lines of code and rules. Or will there be another PR refactoring the tools based on the ideas presented here?

Sorry of if it comes across as harsh or if I am misinterpreting things. It's definitely not intended that way, so see it as a question mostly.

mottosso:
They're developer guides, like what you'd find for Blender, or Unreal, or Docker, or Go. I'd like to pass on what little knowledge I have to you, and bring out the knowledge in you in as structured of a manner as possible for all of our benefit and the benefit of the project. Avalon is growing up, and we're in a good spot to start thinking bigger than merely solving the problem right in front of us without considering the whole.

The goal is for us to both use these as reference, but also as knowledge base for when a new concept is introduced, like whether and how we should allow for and manage singletons, how we can queue tasks like in David's SFTP app, how to handle progress bars, threading and lots of concepts that run the risk of being reinvented or worse yet unconsidered for a given problem.

When we think of one idea that can be improved, we'll improve this documentation (and example). When we think of a new concept to a problem, this is where we'll document it (and implement it into the example). Then we'll have a lexicon of design patterns (and example code) to improve upon alongside the code, making Avalon scalable akin to bigger projects, like Blender, Unreal, Docker and Go.

At least, that's the dream!

davidlatwe:
`"Form is liberating."`, so I vote "Yes" to these :)

CONVENTION.md
```diff
@@ -0,0 +1,453 @@
...
+```py
+class Window(QtWidgets.QMainWindow):
+	title = "My Window"
+
+    def __init__(self, model, parent=None):
+    	super(Window, self).__init__(parent)
+
+    	...
+
+    	layout = QtWidgets.QHBoxLayout(panels["header"])
```
BigRoy: Are we sure it reads better to use `panels["header"]` than just `header` as a variable? It's not a standard practice I've seen before.

CONVENTION.md
```diff
@@ -0,0 +1,453 @@
...
+        self.setCentralWidget(widgets["pages"])
+        widgets["pages"].addWidget(pages["home"])
+
+```
+
+**Notes**
+
+- **S10** Every widget MUST be given a unique name
+- **S11** Every widget MUST be stylable with CSS
+- **S12** Central widget MUST be a `QStackedWidget` to facilitate multiple pages
```
BigRoy: This seems like an odd requirement? Small apps shouldn't need to have this, right? Seems to me that it's overly complicating apps that have require a single page only. Why would we force this?

CONVENTION.md
```diff
@@ -0,0 +1,453 @@
...
+    	layout.setContentsMargins(0, 0, 0, 0)
+    	layout.addWidget(panels["header"], 0, 0)
+    	layout.addWidget(panels["body"], 1, 0)
+    	layout.addWidget(panels["footer"], 2, 0)
+    	layout.addWidget(panels["sidebar"], 0, 1, 0, 2)
+```
+
+**Notes**
+
+- **S20** Layouts MUST all be called `layout`; don't bother with unique names or maintaining reference to them
```
BigRoy: How does this work if you need to do `layout.addLayout()`?

mottosso:
Thanks Roy.

> I'm still not a big fan of putting the convention tool in avalon.tools.convention - it's just cluttter inside there and I'd argue it's not an actual avalon tool but it is a coding example. 

Sure it's not ideal, but I didn't see an issue with it. If you've got a better idea, now's the time.

CONVENTION.md
```diff
@@ -0,0 +1,453 @@
...
+
+- Context switch count: 2
+
+```py
+class Window(QtWidgets.QDialog):
+	def __init__(self):
+		some_button = QtWidgets.QPushButton("Hello")    # 1
+		this_layout = QtWidgets.QVBoxLayout(self)       #
+
+		thatWidget = QtWidgets.QWidget()                # 2
```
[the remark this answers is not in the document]
mottosso: > This seems like a bad example. Why are we using this a "Good" example for a compromise? I think the compromises should be made solely where they cannot be avoided, whereas this one easily can.

What is a better example?

> I would opt actually for Avalon tools using snake_case as that is the code guidelines for Avalon itself, PEP08. 

For clarity, you are referring to "stand fast"? If not, how does your ideal case differ?

- https://github.com/mottosso/avalon-core/blob/convention-guidelines/CONVENTION.md#stand-fast

CONVENTION.md
mottosso: > Are we sure it reads better to use panels["header"] than just header as a variable?

Yeah, let's stick with that. It's something that has worked well in many differently sized UIs, and remains consistent and safe. The added bonus is that it also gives the reader a clear indication of what the name in CSS will be, and that when it's accessed from elsewhere - which you can expect it can - it'll be `self._panels["header"]` as opposed to.. `self._header?` What is that, a line of text? A widget, or what?

CONVENTION.md
mottosso: > This seems like an odd requirement? Small apps shouldn't need to have this, right? Seems to me that it's overly complicating apps that have require a single page only. Why would we force this?

It's only 1 extra line of code for not having to think about this choice and for encouraging additional pages being added; there are a lot of ways in which to add "pages" and without this it isn't clear which to use. Let's keep it.

CONVENTION.md
mottosso: You don't.

Every layout must have a widget.

CONVENTION.md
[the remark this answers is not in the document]
BigRoy: Just I was referring to the *Stand Fast* code where all variables are PEP08 lowercase with underscores. I don't see a reason to mix the variable naming conventions inside of Avalon, if it's a "sometimes" thing it'll only make room for discussion/debating that should just have gotten a clear cut rule.

CONVENTION.md
BigRoy: It's both fine for me. Just wondered whether others didn't feel this would turn into a clutter, including yourself. If not, fine with both to adapt as needed.

CONVENTION.md
BigRoy: Then maybe one of the notes should also be to avoid the use of [`layout.addLayout`](https://doc.qt.io/qt-5/qboxlayout.html#addLayout) which is also discussed in the [tips on layouting for Qt](https://doc.qt.io/qt-5/layout.html#tips-for-using-layouts)

BigRoy:
> Sure it's not ideal, but I didn't see an issue with it. If you've got a better idea, now's the time.

I think the best location where the code can be maintained in a single location would be another repository or a specific "contribution" branch. The last seems an odd use for a branch, so I think the first is still best. The `CONTRIBUTION.md` file could then just link to that repository with conventions and examples.

CONVENTION.md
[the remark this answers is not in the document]
mottosso: > Just I was referring to the Stand Fast code where all variables are PEP08 lowercase with underscores. I don't see a reason to mix the variable naming conventions inside of Avalon, if it's a "sometimes" thing it'll only make room for discussion/debating that should just have gotten a clear cut rule.

But it's not just variables; it's methods too, and supporting functions. The example highlights that (1) if we want snake_case, we must still have mixedCase and (2) that's ugly.

Consistent use of ugly, non-standard mixedCase is greater than inconsistent use of a nice and standard use of snake_case.

But I can see how we'd get the question "what should I use where?", as "majority of code" changes over time. It could start out with mostly mixedCase and turn into mostly snake_case. Do we then refactor to one over the other? The guideline could be "anything widgets use mixedCase, everything else use snake_case".

Not 100% sure, anyone else have any thoughts here?

davidlatwe:
> I think the best location where the code can be maintained in a single location would be another repository

Hmmm, how about put it in `core/conventions/gui` ? I personally lean to not separating them in different repositories. Also I think it might still be changed, so would be better if they were all in same repository when one making a PR ?

Anyway, I think the most important is easy to maintain/access for developers. :P

CONVENTION.md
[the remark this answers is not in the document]
BigRoy: > But it's not just variables; it's methods too, and supporting functions. 

Is it? It's really just methods and functions, right? Or at least, the things you reference from other libraries? What I'm saying is that wherever you don't **require** to match exactly the mixedCase or any other standard, then we don't.

So yes, we do override `mousePressEvent` exactly as it is, but otherwise it doesn't work. However, we do define the callback methods as `on_accept` or alike.

The rule then would be to use `snake_case` unless the code wouldn't work when using `snake_case`.

BigRoy:
> Hmmm, how about put it in core/conventions/gui ? I personally lean to not separating them in different repositories. Also I think it might still be changed, so would be better if they were all in same repository when one making a PR ?

Imagine someone working on an older version of avalon and working with older convention examples from their local deployed version. I'd rather have it be easily distinguishable that it's the latest and greatest conventions, almost like a manual/documentation page. Anyway, I've made my arguments and I'll leave the decision open for Marcus or someone else.

CONVENTION.md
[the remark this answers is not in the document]
mottosso: I think we're saying the same thing.

If we use snake_case where we can, mixedCase elsewhere; that's inconsistent. That's what this bullet point addresses. The question is whether to do that, or to do something about it. What to do? That's what the three options are for; stand fast, give in, or compromise.

mottosso:
> I personally lean to not separating them in different repositories.

Yes, unlike Pyblish, Avalon is a monorepo; everything goes into here. Otherwise we should have a conversation about whether tools and integrations warrant their own repos too, but had I done Pyblish all over again I would have made it a monorepo.

BigRoy:
> Yes, unlike Pyblish, Avalon is a monorepo; everything goes into here. Otherwise we should have a conversation about whether tools and integrations warrant their own repos too, but had I done Pyblish all over again I would have made it a monorepo.

I understand, but this particular conventions repository is not required for anything to work. It's just a code example. It should be part of documentation or somethings. Maybe a branch actually does make sense, similar to how `gh-pages` is used.

CONVENTION.md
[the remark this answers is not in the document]
davidlatwe: So the default is *stand fast*, and other options are just optional ?
I actually lean to all *stand fast* as default, since it’s most simple.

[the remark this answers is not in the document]
BigRoy: Yes, then I agree. It's stand fast that is the **Good** example. Also, the three examples and only one being the right one didn't stand out to me. How about just identifying them as **Good**, **Bad** and **Bad** similar to how you did with the others.

mottosso:
> It should be part of documentation or something

Documentation is also part of this repo, in the `master` branch. `gh-pages` is implementation detail for GitHub to publish it, it isn't something you clone or write into.

Maybe `docs/` is a better place for it?

> So the default is stand fast, and other options are just optional ?

I put the options there to cover all cases; to not only indicate which one to use, but also indicate what else where was and why those are less ideal. I put the little checkbox in `Compromise` as an indicator as to which one to use but it probably wasn't apparent enough.

Happy to stick with Stand Fast, I'll update the checkbox and point it out more explicitly.

BigRoy:
> Maybe docs/ is a better place for it?

To me that does make more sense, sorry if I'm making quite a bold and hard opposition. Feel free to ignore if you both feel different about it.

mottosso:
> Maybe docs/ is a better place for it?

Thinking about this, there is `docs/` which is the [api ref](https://getavalon.github.io/core/), but then there's https://github.com/getavalon/docs which is the [user docs](https://getavalon.github.io).

Would really like to merge these into here, so docs align with the version of Avalon, and so that PRs can be made to both together. And to make it more of a monorepo. That's separate, but means docs/ might not be the best place for this after all.

What we've got is a text document with associated executable code; if I could I would have embedded the Python into the markdown, and let us somehow run the markdown directly. But that's not possible, so the only reason for putting it elsewhere is technical mumbo jumbo. Still not 100% on where.

PR #438, opened 2019-08-26
Nuke: storing attributes in separated knobs with backward compatibility

jezscha:
**What's changed?**
- the way Avalon data are stored in nodes. Before it was text knob (attribute) with dictionary stored as toml dump, now each attribute has own knob which some of those are editable. 
- adding proper docstrigs to functions supporting Argument, Examples, etc
- adding new asset attributes and ensuring backward compatibility on `frameStart/End` and `resolutionWidth/Height` 
- some minor avalon.nuke api code improvments


avalon/nuke/pipeline.py
```diff
@@ -29,28 +25,25 @@ def reload_pipeline():
...
-                   "avalon.api",
-                   "avalon.tools",
-                   "avalon.nuke",
-                   "{}".format(AVALON_CONFIG),
-                   "{}.lib".format(AVALON_CONFIG)):
         log.info("Reloading module: {}...".format(module))
+
         module = importlib.import_module(module)
-        importlib.reload(module)
+        reload(module)
```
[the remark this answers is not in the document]
jezscha: this i am not sure how to do, reload is build in function for python2 which is used by Nuke

avalon/nuke/lib.py
```diff
@@ -3,82 +3,274 @@
...
+    else:
+        return False
 
 
 def fix_data_for_node_create(data):
+    """Fixing data to be used for nuke knobs
+    """
     for k, v in data.items():
+        try:
+            if isinstance(v, unicode):
```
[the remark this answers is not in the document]
jezscha: this is issue of python2-3 compatibility, code is written for python 2 as Nuke doesnt support python3 at the moment

avalon/nuke/pipeline.py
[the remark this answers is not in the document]
davidlatwe: I think we could use `six` :P
```python
from avalon.vendor import six
...
six.moves.reload_module(module)
```

avalon/nuke/lib.py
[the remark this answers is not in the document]
davidlatwe: I think `unicode` could change to `six.text_type`, by the [doc](https://six.readthedocs.io/#six.text_type), so you won't need the `try.. except`.

avalon/nuke/pipeline.py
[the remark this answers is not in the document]
jezscha: have change it to `importlib.reload()` as i t was before.

jezscha:
have resolved all Hound-bot issues but it seems it is still not happy with something.. I am not sure what is the issue..

avalon/nuke/pipeline.py
```diff
@@ -60,27 +53,27 @@ def containerise(node,
...
-        Node
+        node (obj): containerised nuke's node object
 
     """
 
     data_imprint = OrderedDict({
         "schema": "avalon-core:container-2.0",
-        "id": AVALON_CONTAINER_ID,
+        "id": "pyblish.avalon.container",
         "name": str(name),
```
davidlatwe: Is there any reason you need to revert this ? :O
It was from #352 , to avoid hard-coding container id everywhere. :)

avalon/nuke/pipeline.py
```diff
@@ -90,35 +83,45 @@ def containerise(node,
...
 
-    node['avalon'].setValue(toml.dumps(data_imprint))
+    lib.set_avalon_knob_data(node, data_imprint)
 
     return node
 
 
-def parse_container(node, validate=True):
+def parse_container(node):
     """Returns containerised data of a node
```
davidlatwe: Should we drop the validation ?

jezscha: good point, thanks for noticing :)

avalon/nuke/pipeline.py
jezscha: yes, this also makes sense ;)

avalon/nuke/lib.py
```diff
@@ -3,82 +3,283 @@
...
+                    node[name].setValue(str(value))
+
+        # adding closing group knob
+        knob = eval(
+            "nuke.{type}('{name}', '{value}', {group})".format(**knobs[-1]))
+        node.addKnob(knob)
+
+        return node
+
+    except Exception as e:
```
BigRoy: This seems like a very broad Exception, why do we need to do it like this. Can't we be more explicit as to what errors we are actually catching?

avalon/nuke/lib.py
```diff
@@ -3,82 +3,283 @@
...
+                if k["name"] not in node.knobs().keys():
+                    knob = eval((
+                        "nuke.{type}('{name}',"
+                        "'{value}', {group})"
+                    ).format(**k))
+                    node.addKnob(knob)
+
+        # add avalon knobs for imprinting data
+        for k, v in data.items():
+            label = k
```
BigRoy: If you are setting `label = k` and value = `str(v)` anyway , why not do `for key, value in data.items()` and just do `value = str(value)`?

avalon/nuke/lib.py
```diff
@@ -3,82 +3,283 @@
...
+    if "publish" not in node.knobs():
+        divider = nuke.Text_Knob('')
+        knob = nuke.Boolean_Knob("publish", "Publish", True)
+        knob.setFlag(0x1000)
+        node.addKnob(divider)
+        node.addKnob(knob)
+    return node
+
+
+def set_avalon_knob_data(node, data={}, prefix="ak:"):
```
BigRoy: I'm not familiar with Nuke. But is there a reason you went for prefix `ak:` (Avalon Knob?) instead of `avalon:`? Is it required to be of a very short form. Why not be more explicit?

avalon/nuke/lib.py
```diff
@@ -3,82 +3,283 @@
...
+                        label=label,
+                        value=value
+                    ))
+                    node.addKnob(knob)
+                else:
+                    log.info("Updating: `{0}` to `{1}`".format(name, value))
+                    node[name].setValue(str(value))
+
+        # adding closing group knob
+        knob = eval(
```
BigRoy: Why is this using `eval()`? That seems very fishy that it's required to do it like that.

I'd think that the pythonic way would be:
```python
fn = getattr(nuke, knob["type"])
knob = fn(name, value, group)
```

Actually, even line 132 to 161 could be replaced with:
```python
# pseudocode
knob = node.knob(name)
if not knob:
    # Create knob
    knob_type = "String_Knob" if label in non_hidden else "Text_Knob"
    fn = getattr(nuke, knob_type)
    knob = fn(name, label, value)
knob.setValue(value)
```

avalon/nuke/lib.py
```diff
@@ -3,82 +3,283 @@
...
+            "nuke.{type}('{name}', '{value}', {group})".format(**knobs[-1]))
+        node.addKnob(knob)
```
BigRoy: Could you elaborate why this closing group knob is being done separately? Doesn't this mean it'll try to add this closing tab every time, even when it already existed?

avalon/nuke/lib.py
```diff
@@ -3,82 +3,283 @@
...
+        prefix (str, optional): filtering prefix
+
+    Returns:
+        data (dict)
+    """
+    try:
+        # check if data available on the node
+        test = node['avalon_data'].value()
+        log.debug("Only testing if data avalable: `{}`".format(test))
+    except Exception as e:
```
BigRoy: Very broad exception, can this be more specific? E.g. `NameError` or what does it raise?

avalon/nuke/pipeline.py
```diff
@@ -171,13 +196,14 @@ def ls():
     See the `container.json` schema for details on how it should look,
     and the Maya equivalent, which is in `avalon.maya.pipeline`
     """
-    all_nodes = nuke.allNodes(recurseGroups=True)
+    all_nodes = nuke.allNodes(recurseGroups=False)
 
     # TODO: add readgeo, readcamera, readimage
```
BigRoy: This todo seems to have been resolved since it now includes literally `all_nodes`?
Which basically means the `nodes` list comprehension does nothing except making a copy of the full list. As such, it's redundant?

avalon/nuke/pipeline.py
```diff
@@ -239,17 +265,16 @@ def uninstall(config):
...
+    """Installing Avalon menu to Nuke
+    """
     from ..tools import (
         creator,
-        # publish,
+        publish,
         workfiles,
-        loader,
-        sceneinventory,
+        cbloader,
```
BigRoy: This will break things, as current master has them refactored to `loader` and `sceneinventory` already.

avalon/nuke/pipeline.py
```diff
@@ -367,12 +458,11 @@ def _on_task_changed(*args):
...
     try:
-        # nuke = getattr(sys.modules["__main__"], "nuke", None)
-        # lock all connections between nodes
-        # nuke.Root().knob('lock_connections').setValue(1)
-
         # stop active viewer to update any change
-        nuke.activeViewer().stop()
+        try:
+            nuke.activeViewer().stop()
+        except Exception as e:
```
BigRoy: Can we make this error more explicit?
```
try:
    viewer = nuke.activeViewer()
except WhatEverErrorNukeRaises as e:
    log.warning("No available active Viewer: `{}`".format(e))
```

Furthermore I'd expect Nuke to return None or raise a specific error when there is no active viewer.
Because currently you might be failing on actually stopping the viewer! Should be something like this maybe?
```python
viewer = nuke.activeViewer()
if viewer:
    viewer.stop()
```

avalon/nuke/pipeline.py
```diff
@@ -261,68 +286,134 @@ def _install_menu():
...
 
     nuke.root()["fps"].setValue(fps)
-    asset = api.Session["AVALON_ASSET"]
-    asset = io.find_one({"name": asset, "type": "asset"})
+    name = api.Session["AVALON_ASSET"]
+    asset = io.find_one({"name": name, "type": "asset"})
+    asset_data = asset.get("data")
+
+    if not asset_data:
+        msg = "Asset {} don't have set any 'data'".format(name)
```
BigRoy: If this ever happens there's something else that's wrong because the Asset schema requires `data` to exist, see [here](https://github.com/getavalon/core/blob/60da0a0494fb3fd930680d5eba55e441a9e999f7/avalon/schema/asset-2.0.json#L16). So if you face(d) this problem make sure wherever you are creating assets that you validate the schema before inserting into the database... because this check should be entirely redundant. :)

avalon/nuke/lib.py
```diff
@@ -3,82 +3,283 @@
...
-    """Adding a tab and a knob into a node
+def select_nodes(nodes):
+    """Selects all inputed nodes
+
+    Arguments:
+        nodes (list): nuke nodes to be selected
     """
+    assert isinstance(nodes, (list, tuple)), "nodes has to be list or tuple"
+
+    [n['selected'].setValue(True) for n in nodes]
```
BigRoy: This is better written as:
```python
for node in nodes:
    node['selected'].setValue(True)
```

A list comprehension should not be used to perform actions or have those side effects, see [here](https://docs.python-guide.org/writing/style/#short-ways-to-manipulate-lists):

> Never use a list comprehension just for its side effects.

avalon/nuke/lib.py
```diff
@@ -3,82 +3,283 @@
...
+            [n['selected'].setValue(True)
+             for n in nodes
+             if n.name() in previous_selection]
+
+
+def reset_selection():
+    """Deselect all selected nodes
+    """
+    nodes = nuke.allNodes()
+    [n['selected'].setValue(False) for n in nodes]
```
BigRoy: See [this comment](https://github.com/getavalon/core/pull/438/files#r319584885) about not using list comprehensions to perform actions, it's not pythonic.

Plus, for deselecting you wouldn't need to iterate over all nodes. Instead just do:
```python
for node in nuke.selectedNodes():
    node['selected'] = False
```

avalon/nuke/lib.py
```diff
@@ -3,82 +3,283 @@
...
+    """
+    nodes = nuke.allNodes()
     previous_selection = nuke.selectedNodes()
     try:
         yield
     finally:
         # unselect all selection in case there is some
-        current_seletion = nuke.selectedNodes()
-        [n['selected'].setValue(False) for n in current_seletion]
+        [n['selected'].setValue(False) for n in nodes]
```
BigRoy: Here you can use the `reset_selection()` that you conveniently wrote below. If the `reset_selection` is really to simple to even be it's own function (which I think it actually is) I'd just change this to:

```python
for node in nuke.selectedNodes():
    node['selected'].setValue(False)
```

BigRoy:
Good to see this update! :) I went ahead and made some comments and remarks as there seem to be some code style quirks. They are all listed above, but just wanted to point one thing out in particular which is in some other locations too that I did not comment on and that is [Don't use list comprehension just for its side effects.](https://github.com/getavalon/core/pull/438#discussion_r319584885)

Also, I had a question regarding the selections in Nuke. What's the reason you're selecting and deselecting the nodes with `node['selected'].setValue()` as opposed to using [`node.setSelected()`](https://learn.foundry.com/nuke/developers/70/pythonreference/nuke.Node-class.html#setSelected). I'd imagine the latter is shorter, faster and clearer. Plus, it's easier to refactor with most IDEs.

avalon/nuke/lib.py
BigRoy: If there's another reason to use `allNodes()` then note that Nuke also seems to expose [`clear_selection_recursive` in nukescripts](https://learn.foundry.com/nuke/developers/70/pythonreference/nukescripts.misc-pysrc.html#clear_selection_recursive). Not sure if that helps.

avalon/nuke/lib.py
```diff
@@ -3,82 +3,283 @@
...
+        {"name": '__divider__'},
+        {"name": 'avalon_data',
+         "value": 'Warning! Do not change following data!',
+         "type": "Text_Knob"},
+        {"name": '__divider__'},
+        {"name": 'begin',
+         "value": 'Avalon data group',
+         "type": "Tab_Knob", "group": -1}
+    ]
+    non_hiden = ["asset", "subset", "name", "namespace"]
```
BigRoy: Typo `non_hiden` -> `non_hidden`

mottosso: > This could also be turned into visible if that makes more sense.

+1 to that. Avoid double-negatives.

mottosso:
Great review Roy; what he said!

avalon/nuke/lib.py
jezscha: no particular reason. It could be `avalon` or it could be whatever

avalon/nuke/lib.py
```diff
@@ -3,82 +3,264 @@
...
+        return False
+
+
+def fix_data_for_node_create(data):
+    """Fixing data to be used for nuke knobs
+    """
+    for k, v in data.items():
+        if isinstance(v, six.text_type):
+            data[k] = str(v)
+        elif "0x" in str(v):
```
BigRoy: Where would the `0x` value come from that it's a string? 
Anyway, should this be `str(v).startswith("0x")` instead? 

Seems like this can be more strict as opposed to it somehow messing up with a potential subset `hero_var0xtra` (worst naming convention ever). You never know what someone might name their stuff.

avalon/nuke/pipeline.py
```diff
@@ -132,31 +140,48 @@ def update_container(node, keys=dict()):
...
-
         if (self.options or {}).get("useSelection"):
             nodes = nuke.selectedNodes()
 
-        instance = [n for n in nodes
-                    if n["name"].value() in self.name] or None
+        if len(nodes) > 0:
+            node = nodes[0]
+        elif len(nodes) > 1:
+            nuke.message("Please select only one node")
```
BigRoy: Doesn't this message still allow the code to continue? Or is it like raising an error? If this still continues then it seems odd. Any reason it's done like this?

mottosso:
What's the status of this PR? There are a number of comments by Roy that hasn't been addressed, but I see commits still being made. Any plans to make this mergeable soon?

davidlatwe:
I think the first step for picking this up again for everyone is to update the change note, the previous change note seems outdated.

A bit hard to review now. :worried:

A list of changes would be a good second start and bring much help for us @jezscha :)

jezscha:
> Good to see this update! :) I went ahead and made some comments and remarks as there seem to be some code style quirks. They are all listed above, but just wanted to point one thing out in particular which is in some other locations too that I did not comment on and that is [Don't use list comprehension just for its side effects.](https://github.com/getavalon/core/pull/438#discussion_r319584885)
> 
> Also, I had a question regarding the selections in Nuke. What's the reason you're selecting and deselecting the nodes with `node['selected'].setValue()` as opposed to using [`node.setSelected()`](https://learn.foundry.com/nuke/developers/70/pythonreference/nuke.Node-class.html#setSelected). I'd imagine the latter is shorter, faster and clearer. Plus, it's easier to refactor with most IDEs. Additionally it also seems to be what Nuke itself seems to use in nukescripts, so it must be the default Nuke way of doing things too.

Thank you @BigRoy for the detailed feedback. Yes I agree it is better to iterate in regular way but there is so many ways to iterate so I am glad you had pointed it out for me. Regarding your second point of selecting nodes, you are right, this could be used but I believe it is function which is doing the same inside. This is just because I am used to write it this way for years so haven't been pushed to research other methods :) 

Anyway, thank you all for your time and hope this will find its user in the community!

jezscha:
BTW, sorry for the delay. I was having lot to do on other integrations. Nukestudio is almost done! :)

avalon/nuke/pipeline.py
```diff
@@ -132,31 +140,48 @@ def update_container(node, keys=dict()):
...
+
+    # If not all required data return the empty container
+    required = ['schema', 'id', 'name',
+                'namespace', 'objectName', 'representation', 'version']
+    container = dict()
+
+    # check if key in data
+    for k in required:
+        if not data.get(k, 0):
+            # if not then create the key in container and data
```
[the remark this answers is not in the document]
jezscha: You are right, this code is actually obsolete so I had removed it. It was there as older version of container data in node were missing `version` so this was just a patch to add it it there. I had removed it now.

avalon/nuke/pipeline.py
jezscha: well, this all is fundamentally wrong. so I will rewrite it in the Maya's fashion.

davidlatwe:
I am about to adopt Nuke into our pipeline here, so I'll try to have a deeper look into this PR in next few days and sort things out. :)

avalon/nuke/pipeline.py
jezscha: Default `avalon.nuke.pipeline.Creator` will now create Backdrop node with imprinted publishable metadata

tokejepsen:
@jezscha I see a lot of minor Hound issues. Are you using a linter?

Seems like these issues you could easily spot if you have a linter like flake8 or something.

jezscha:
> @jezscha I see a lot of minor Hound issues. Are you using a linter?
> 
> Seems like these issues you could easily spot if you have a linter like flake8 or something.

Yes I do, but the length of row is quite strict.

jezscha:
> I am about to adopt Nuke into our pipeline here, so I'll try to have a deeper look into this PR in next few days and sort things out. :)

Are you about to adopt even Nukestudio in your studio?

tokejepsen:
> Hi @tokejepsen how are you doing with testing, will you find some time for it?

No, sorry. We wont be updating our pipeline until the current project is done. Wont have any free time to test this either.

BigRoy:
@davidlatwe any status update on your end on testing this?

davidlatwe:
Just started an hour ago ! Still trying to get things to work, Avalon + Nuke day 1.

jezscha:
> @jezscha according to Github there seems to be a conflict on `avalon/nuke/pipeline.py` - could you take a look?

Hi @BigRoy I cannot see any other conflict. Where is it?

davidlatwe:
Haha, I solved the conflict yesterday for testing on my end. :grimacing:

jezscha:
> Haha, I solved the conflict yesterday for testing on my end. 😬

great! ;) thanks

davidlatwe:
After take a deeper look into this PR and have a go on those new functions, although they could work but I think some of them are misplaced at the very first beginning.

I think `get_handles` and `reset_frame_range_handles` should **NOT** be in `pipeline.py`, instead they should be in `command.py` like the one in Maya, or in `lib.py`.

Because to me, `pipeline.py` should be containing something that responsible to build Avalon framework on top of Nuke, not something like component which can be used in multiple cases or opt-out.

Next, `lib.imprint` shoud **NOT** be wrapping `add_publish_knob` and `set_avalon_knob_data`, should be in other way around instead.

Since the name *imprint*, did not give a hint that it's working on a specific goal but to imprint *any kind of data* into node. But what has been implemented was only for writing Avalon related data at the end.

Now, I am not sure that we should merge this at the moment (sorry). I could, because at least I can confirm that nothing is broken here. :relaxed:

But we should not, because there are things we need to refactor in prior (those mentioned above). :construction: 

What I have in mind for now is to start a couple of PRs for shipping concepts from this PR piece by piece. What do you guys think ?

jezscha:
Hi David, thank you very much for your time to look into it! Please read bellow my comments.

> I think `get_handles` and `reset_frame_range_handles` should **NOT** be in `pipeline.py`, instead they should be in `command.py` like the one in Maya, or in `lib.py`.
> 
Good point, will move it into `command.py`

> Because to me, `pipeline.py` should be containing something that responsible to build Avalon framework on top of Nuke, not something like component which can be used in multiple cases or able to be opt-out in production.
>
We will look into it with @mkolar and make it more consistent with Maya 

> Next, `lib.imprint` shoud **NOT** be wrapping `add_publish_knob` and `set_avalon_knob_data`, should be in other way around instead.
> 
Yes. Also I am thinking to even move `add_publish_knob` out of _lib.imprint_, perhaps the function could be used from outside of imprint

> Since the name _"imprint"_, did not give a hint that it's working on a specific goal but to imprint _any kind of data_ into node. But what has been implemented was only for writing Avalon related data at the end.
> 
Not sure exactly what do you mean with: __ut what has been implemented was only for writing Avalon related data at the end.__

> Now, I am not sure that we should merge this at the moment (sorry). I could, because at least I can confirm that nothing is broken here. ☺️
> What I have in mind for now is to start a couple of PRs for shipping concepts from this PR piece by piece. What do you guys think ?
> 
Again, we will look into it with @mkolar next week and for sure will split it into several PR and push them separately - this PR has got too big already. 
If you have done something already let us know so we don't duplicate work to each other ;)

davidlatwe:
> If you have done something already let us know so we don't duplicate work to each other ;)

Thanks @jezscha !

I have pushed a PR for new `imprint` #473 , and submitted an issue for discussing Nuke containerizing #474 . Hope you guys could have a look. :grimacing:

jezscha:
I guess this could be closed.. All of changes here were rewritten by @davidlatwe.

PR #440, opened 2019-09-04
Refactor avalon.tools to avoid cross-referencing imports between tools

BigRoy:
_Note:_ This is a big PR as it refactors locations of models, views, delegates and functions. It's this big because unfortunately it needs to be.

**Problems**

1. Different tools cross-referenced each other with imports from the other tools.
2. The tree model's item is called `Node`, which is wrong terminology as Qt refers to `QAbstractItemModel` where it is an `Item`
3. The family and group configs that are "refreshed" in `loader.lib` are actually caches reused across multiple tools. This is unclear from the variable and function naming. (This "global state" is a problem of its own, but this PR will not try to solve that as it would require _even_ more changes.) It only solves the cross-referencing across tools.
4. Some variables are mixed case like `lowerCamelCase` where they should be `snake_case` as per #437 
5. An `AssetView` class exists which purely exists for one-time use and only sets 3 small view tweaks - it's redundant complexity.
6. The `TreeModel` defines an attribute called `COLUMNS` which should be lowercase `columns`.
7. There is an unused model `ExactMatchesFilterProxyModel` which is redundant lines of code.

**What's changed?**

1. All shared tool models, views, delegates, widgets and functions have been respectively refactored to `avalon.tools.models`, `avalon.tools.views`, `avalon.tools.delegates`, `avalon.tools.widgets` and `avalon.tools.lib`
2. The `Node` is refactored to `Item`, including variables `node` and alike that I could find are refactored to `item`.
3. This is _still_ a global state, however the variables and functions now are clearly referring to it as a cache and simplified to the only way they are used. It's up to a following PR to move the caching to inside a model as opposed to a global state.
4. Refactored to `snake_case`.
5. The `AssetView` class has been removed and the [related lines of code can now directly be found in the `AssetWidget`](https://github.com/BigRoy/core/blob/c350d18476173a8ae4acdbb722edd34a3d6553f3/avalon/tools/widgets.py#L63). 
6. `TreeModel.COLUMNS` refactored to `TreeModel.Columns`
7. Remove `ExactMatchesFilterProxyModel`.

---

This is a work in progress PR and has not been tested yet. However I'm opening this to kickstart the discussion about any faulty choices I'm making. Plus, since it's a lengthy PR it gives others time to go through what is being changed.

I'll clearly state in a comment when I have been able to thoroughly test changes and feel it's in a good state for others to test too.

---

This PR acts as a replacement to #414 trying to do some similar things yet now along the new Contribution guidelines. Additionally it tries to describe the problems that we faced before and how they have been resolved.


avalon/tools/widgets.py
```diff
@@ -0,0 +1,327 @@
+import logging
+
+from .lib import (
+    preserve_expanded_rows,
+    preserve_selection,
+    iter_model_rows
+)
```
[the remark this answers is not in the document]
BigRoy: Thanks David. Is that based on the [e3 example in `CONTRIBUTING.md`](https://github.com/getavalon/core/blob/master/CONTRIBUTING.md#e3)? I interpreted it as *not* importing as a shorter name and not necessarily removing all functions imports.

But probably @mottosso can best describe his intentions. I'm fine with both, do note however that if this needs to be adapted that it accounts for more things, e.g. [see these lines](https://github.com/getavalon/core/blob/27449a5d809f1d325ed474d50d2388c93723a96a/avalon/tools/widgets.py#L8).

I did now see there's an import there that states:
```python
from ..vendor import qtawesome as awesome
```

Which I believe should be:
```python
from ..vendor import qtawesome
```

BigRoy:
With this [last commit](https://github.com/getavalon/core/pull/440/commits/a4756233958a58ee3e3dac1a6bc376adaacd1b1c) I checked out the pipeline features in Maya and ran through:

- [x] Switching contexts with ContextManager
- [x] Publishing with Pyblish QML
- [x] Loading some published assets.
- [x] Managing some published assets with scene inventory (removing + updating to specific versions)
- [x] Creating some publish instances with the Creator.
- [x] Opening the Project Manager, creating an asset and adding tasks.
- [x] Work files opening a file and saving a file.

This now all worked so I consider this testable for others.

avalon/tools/widgets.py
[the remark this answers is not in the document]
iLLiCiTiT: That irritates me for a long time I think it should be `from ..vendor import qtawesome`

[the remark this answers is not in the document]
BigRoy: > `from ..vendor import qtawesome as awesome`

This has now been resolved with [this commit](https://github.com/getavalon/core/pull/440/commits/9bc098edc701f9856d1b9ba0e016ddf22577fcb2).

avalon/tools/delegates.py
```diff
@@ -0,0 +1,71 @@
...
+
+        # Current value of the index
+        value = index.data(QtCore.Qt.DisplayRole)
+        assert isinstance(value, int), "Version is not `int`"
+
+        # Add all available versions to the editor
+        item = index.data(TreeModel.ItemRole)
+        parent_id = item['version_document']['parent']
+        versions = io.find({"type": "version", "parent": parent_id},
+                           sort=[("name", 1)])
```
davidlatwe: Maybe we could also clean up all the single quote `'` into double quote `"` in this PR ?
Avalon uses double quote, right ?

avalon/tools/widgets.py
```diff
@@ -0,0 +1,327 @@
...
+        silos = list(set(silos))  # ensure unique
+        self.set_silos(silos)
+        self.silo_added.emit(silo)
+
+        self.set_current_silo(silo)
+
+
+def _list_project_silos():
+    """List the silos from the project's configuration"""
+    silos = io.distinct("silo")
```
davidlatwe: This function and entire `avalon/tools/projectmanager/lib.py` looks like they could be put together into `avalon/tools/database.py` ? A place for database io functions. What do you think ?

avalon/tools/widgets.py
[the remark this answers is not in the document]
davidlatwe: > Is that based on the e3 example in CONTRIBUTING.md?

I was actually referencing from [A8 in `CONTRIBUTING.md`](https://github.com/getavalon/core/blob/master/CONTRIBUTING.md#code-quality---architecture), which aim to preserve the namespace :P

> Especially now that there's a lib at multiple levels (avalon-wide + tools-wide + per tool) seeing lib. wouldn't tell much.

Yeah.. maybe this will work ? Say we are in bad luck, used all three levels' `lib.py` in one module:
```python
from ... import lib as avalon_lib
from .. import lib as tool_lib
from . import lib
```

avalon/tools/widgets.py
BigRoy: I'm not necessarily for *more separation* where it's currently not required. Unless it's solving a particular problem. This function is solely there because one of the widgets uses it, as described in the contribution manual it should be as close as possible to the classes/functions using it.

If we would move it into a `database.py` or something alike we could make it a separate PR.

avalon/tools/delegates.py
BigRoy: Fine with me. Will take a peek at doing that tomorrow. :)

BigRoy: Should be resolved with [this commit](https://github.com/getavalon/core/pull/440/commits/4c30973ee482fd749ebf041aa9223a32ef9d5905). Note however that I only went through everything in `avalon.tools` to keep this PR contained to the tools. :)

davidlatwe:
Ah, backward compatibility, but do we need to do that for this kind of *low level refactoring* ?

I have some script/tools that, for example, used `projectmanager.model.TreeModel`. But since these doesn't count as part of `api` nor other higher level things, so maybe we won't need to support this ?

BigRoy:
> I have some script/tools that, for example, used projectmanager.model.TreeModel. But since these doesn't count as part of api nor other higher level things, so maybe we won't need to support this ?

Exactly, as those have never been exposed to any public facing API these changes are in the realm of "allowed to change". So, those don't need to be backwards compatible and you'll have to refactor in-house code that used it to change along. We had some internal tools referencing things too. ;)

Anyway, *anything not exposed in the API* is allowed to change without being fully backwards compatible. (e.g. a function name may behave differently, be renamed, moved, etc.). And the recommendation is not to reference those methods, but of course... you can if you really need to at your own risk, like now.

avalon/tools/widgets.py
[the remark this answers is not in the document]
mottosso: This one's a hard problem.

On the one hand, word and line length obscure code. One the other, namespaces give the reader a hint at what a function or class is coming from.

For some modules, it's obvious.

```py
qtawesome.some_function()  # I know you're third-party
```

For others, not so much.

```py
model = SomeModel()  # Third-party, local module, separate root-level module?
```

Let's try this as a general guideline; preserve the leaf-level module namespace.

- `from .vendor import qtawesome` is Good
- `from .vendor.qtawesome import icon` is Bad

> maybe this will work ?

This is another hard problem; on the one hand, we should use `lib` at various levels, as they are context sensitive. A `lib` under `tools` is tools libraries, whereas a `lib` under `maya/` is Maya libraries. And on the other they complicate the above general guideline.

I think this is as elegant of a solution as we'll get, if we are to have at least some guideline for how to deal with it. Preserve the *minimum* number of namespaces; in this case, we need both `tool` and `lib`, if `avalon_lib` is also present.

avalon/tools/delegates.py
mottosso: > Maybe we could also clean up all the single quote ' into double quote " in this PR ?
Avalon uses double quote, right ?

Yes, thank you David for being vigilant. This is one of those "pebbles in a shoe" that's innocent on its own but distracts the reader (at least me) from the meaning of code.

avalon/tools/loader/model.py
```diff
@@ -1,17 +1,23 @@
...
+    (NOTE) Recursive filtering was introduced in Qt 5.10.
+
+    """
+    return hasattr(QtCore.QSortFilterProxyModel,
+                   "setRecursiveFilteringEnabled")
 
 
 class SubsetsModel(TreeModel):
-    COLUMNS = ["subset",
+    columns = ["subset",
```
mottosso: Could we make this `Columns`? It's to separate it from variables and methods.

avalon/tools/widgets.py
mottosso: Yes, let's keep it contained until it absolutely must be separated. Less indirection the better.

mottosso:
I suppose that "Approved" should have been "Requesting changed", as I think we should address my comments before merging. I'll let @davidlatwe hit the Merge button for this PR on my behalf. Great work guys.

avalon/tools/widgets.py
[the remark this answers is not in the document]
BigRoy: Thanks @mottosso - just to be complete, if I understand you correctly the above example to import as `tool_lib`, etc. when required is considered the correctly approved change. Yes? 👍

avalon/tools/loader/model.py
BigRoy: Sure. Fixed with [this commit](https://github.com/getavalon/core/pull/440/commits/8f798ca2699897b51fd9d0fe65eb158eed40a8f5).

BigRoy:
This should be ready to go if I understood @mottosso correctly [here](https://github.com/getavalon/core/pull/440#discussion_r321654086) then there are now no open comments anymore as I refactored the `columns` to `Columns` which was the last remark.

I'd say, give it one other test run and feel free to merge @davidlatwe 🚀 

Thanks all!

avalon/tools/loader/app.py
```diff
@@ -1,20 +1,18 @@
...
 
 from ...vendor.Qt import QtWidgets, QtCore
 from ... import api, io, style
 from .. import lib
 
-from .lib import (
-    refresh_family_config,
-    refresh_group_config,
-    get_active_group_config,
+from ..lib import (
```
mottosso: No, this should be `from .. import lib` which you're already doing above, but then not using?

mottosso: It's being imported three times (!) in three different ways, let's clean this up.

BigRoy: Thanks @mottosso - well spotted! I tried to simplify and clean things up with [this commit](https://github.com/getavalon/core/pull/440/commits/5de72f38cf1ac5a3b39f7f5e6e2729fbc1a95255). Let me know if that suffices.

davidlatwe:
Ah, I believe we should bump the version to `5.4.0`, :wave: @BigRoy :)
I will merge this PR and draft a *pre-release* after verison bumped !

BigRoy:
@davidlatwe this hasn't been merged yet. Do you want me to quickly bump version? :)

davidlatwe:
Yes, please do ! It seems that I could not bump it myself into this PR :P

PR #442, opened 2019-09-10
Merge Context Manager + Work Files (Fix #412)

### LINKED ISSUE #412: Merge Context Manager + Workfiles tool into single UI.
## Issue

Currently the context (current asset + task) is managed with the Context Manager and the Work Files can be saved/opened through the Work Files app.

I'd argue that when combining these tools into a single unified interface that it would become much faster to switch tasks and load the right work file.

Here's a quick mockup of what I envision it would look like when merged:

merged_contextmanager_workfiles

---

Feel free to start a discussion about pros/cons and potential other redesigns.

BigRoy:
This is a work in progress implementation for issue #412 that describes to merge the Context Manager and Work Files tool into one single tool that handles both, as usually the actions performed with these happen in relation to each other.

**What's changed?**

- The refactoring code is now written into the work files tool. I haven't removed the Context Manager yet (WIP)
- I've also started to refactor the code for the work files `app.py` according to the contribution guidelines, however I think this will introduce quite some discussion and pointers on how to improve - as such I'm opening this PR In an early stage to ensure as much as possible this can be discussed and I can improve it as needed.
- The current implementation works roughly similar to what was described [here](https://github.com/getavalon/core/issues/412#issuecomment-529398602).

---

**Note:** This is a work-in-progress and I'm opening a PR to kickstart a discussion on code style, but I'm fully aware this is not as clean as it can be. However, even at this stage, code style ideas or notes according to the contribution guidelines are very welcome.

**Note 2:** I've added some todos that are more notes to self as I'm developing and changing some minor things purely for testing/development sake. Feel free to be captain obvious and point out if you have questions - just wanted to highlight that many are reminders to myself.


avalon/tools/workfiles/app.py
```diff
@@ -183,173 +235,336 @@ def refresh(self):
...
         # Select last modified file
-        if self.list.count():
-            item = self.list.item(modified.index(max(modified)))
+        if list.count():
+            item = list.item(modified.index(max(modified)))
             item.setSelected(True)
 
             # Scroll list so item is visible
-            QtCore.QTimer.singleShot(100, lambda: self.list.scrollToItem(item))
+            callback = lambda: self.widgets["fileList"].scrollToItem(item)
```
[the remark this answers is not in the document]
BigRoy: Resolved with https://github.com/getavalon/core/pull/442/commits/eb5f152c96e9ebebead241709287f6a5084be487

avalon/tools/workfiles/app.py
```diff
@@ -183,173 +235,339 @@ def refresh(self):
...
+
+        # Refresh current scene label
+        current = os.path.basename(self.host.current_file()) or "<unsaved>"
+        self.widgets["fileCurrent"].setText("Current File: %s" % current)
+
+        # Refresh breadcrumbs
+        self.widgets["header"].refresh()
+
+        # Refresh files list
+        list = self.widgets["fileList"]
```
davidlatwe: Wouldn't it a bit danger that overriding the class `list` ? Since it's often used :o

avalon/tools/workfiles/app.py
```diff
@@ -3,83 +3,123 @@
...
 
         # Force default focus to comment, some hosts didn't automatically
         # apply focus to this line edit (e.g. Houdini)
-        self.comment_lineedit.setFocus()
+        comment.setFocus()
+
+        self.widgets = {
+            "preview": preview,
+            "comment": comment,
+            "version": version,
```
davidlatwe: Not sure about this one, the `self.widgets` here is collecting widgets from defined vars, but in `ContextBreadcrumb` and the `Window`, the `self.widgets` is init classes first, then making changes.

I wonder if this should be unified, too ?

I ***imaging*** that I may firstly look up the `self.widgets` to have a quick overview on this class, so if there were all just variables would not be very helpful. Again, just imaging :P

avalon/tools/workfiles/app.py
BigRoy: Well spotted, I guess I went a bit to quickly with my refactoring - will fix.

avalon/tools/workfiles/app.py
BigRoy: Ah, yes. I see how they differ. Admittedly I have no preference as I have to get used to either. So I guess this is a good one to take a pick and see what @mottosso brings to the table about how he intended it.

The convention that is [laid out currently seems to first declare solely the widgets at the top](https://github.com/getavalon/core/blob/e40eac945bdeae91dedc38418863624db1bcf613/avalon/tools/convention/view.py#L26) so I guess we should move into that direction.

mkolar:
Just realised that I commented on the issue rather than here. So for completeness here's clarification on my early comments https://github.com/getavalon/core/issues/412#issuecomment-530479801

tokejepsen:
Have recently been using the Context Switcher more, and I immediately go to the Workfiles app afterwards, so merging the two would make it very streamlined.

BigRoy:
Tested it again, and now the tools open fine on Houdini without errors and they are nicely parented. The Avalon stylesheet still doesn't apply perfectly, it still seems to inherit quite a bit from the Houdini styling.

However it seems the [styling of Houdini itself might change in the near future](https://www.sidefx.com/forum/topic/66718/?page=1#post-284959) that might automatically resolve it. Maybe it's best to leave it as is since it isn't a big issue and then maybe with an upcoming Houdini version less of the Houdini styling pushes through in the Avalon tools.

Nice work @davidlatwe ! 🚀

BigRoy:
I've been recently fiddling with the Work Files API to allow the tool to list the files even without requiring to switch the context prior to it. Admittedly it's a bit hacky, especially with the root dir in Maya being dependent of the `workspace.mel`.

Anyway, in the end it seemed to be fast and work quite well. I think from there it's a matter of cleaning up and simplifying again.

I hope to share some bits of it in about two or three days as my schedule this week is quite busy. Just wanted to let you know I am back on this PR.

jasperges:
@BigRoy That's awesome! This would also help in adopting 'Allzparkalon'. And like @tokejepsen is saying: most of the time when you switch context you open the Workfiles app anyways.

BigRoy:
Here's a preview of the latest functionality:
avalon_workfiles_merged_simple_gif

Download zipped `.mp4`: [avalon_workfiles_merged_simple.zip](https://github.com/getavalon/core/files/3968218/avalon_workfiles_merged_simple.zip)

_The little log window at the bottom is Maya's script editor and is **not** part of the Work Files UI._

To-do:
- Prior to save/open the Session should first be set to ensure it's switched prior to open/save automatically.
- **Set Context functionality**: Double clicking the Task sets the context manually (forcing it) - I want to make that clear by having a "Breadcrumb" of sorts at the top of the UI showing the selected Asset/Task and color coding that e.g. orange when it is *not* the currently active Session context, if that's the case... behind that I'll add two buttons, one for "Setting the context" (like double clicking the task does now) and one for "Go to current context" that will browse back to your current active `api.Session` context. The buttons will be disabled when they are redundant (e.g. go to current context will be disabled when you're already there) and "set context" is disabled when you're already in the context. Then browsing should make this self-explanatory and the tooltips of the buttons can aid in that as well.
- **Initialize workdir**: In the case when an Host has never been initialized to a specific asset+task the Workdir does not exist. What should happen then is to perform the [App initialization](https://github.com/getavalon/core/blob/master/avalon/pipeline.py#L376) to make sure it's setup like the studio config intended. When browsing to a non-initialized Work Directory I want the most right-widget to be swapped with a message saying "No work directory exist" with a "Initialize Work directory" button underneath. Currently [performing this "App Initialize" is non-trivial yet a hacky method for it can be found here](https://github.com/Colorbleed/colorbleed-config/blob/master/colorbleed/maya/__init__.py#L193). What would be the best way to improve this?

Any first thoughts are more than welcome.

tokejepsen:
Looks good!

`Initialize Work directory` might sound a bit too techy/cody. Maybe `Create work area`

davidlatwe:
I like the ideas in #489, will they get implemented in this PR, or in another one ?

How about adding a "New" button, so there should be pretty much what artists will do when they open up context manager. Then I think we won't need double clicking to set context and the Breadcrumb.

BigRoy:
> How about adding a "New" button, so there should be pretty much what artists will do when they open up context manager. Then I think we won't need double clicking to set context and the Breadcrumb.

@davidlatwe Sorry I'm not sure I get what you mean. 

Whenever you launch the host it should have already initialized the current application. Or at least that is what Avalon currently does as it launches the host. So for the *current context* you would never have to "create work area" since it already exists. It would by default just show an empty folder on the right hand side. Where I'd want to add a single non-selectable greyed out entry in the list stating: "No files. Use Save As to save your first file."

The tricky thing is that as soon as you click on another Asset or Task you're operating in an area that is not related to you active Context/Session. Previously it came up that there could be reasons to force yourself into the workspace context, even without Saving/Opening a file there which @mkolar mentioned before. It would basically be, setting that context and doing some work *prior* to saving there if I understood correctly.  Exactly that is what the "buttons" would allow, to _if you want_ set the context. However, in most scenarios I assume an Artist doesn't care and will just open the scene (which will always first ensure the context is set appropriately).

davidlatwe:
Ah, I meant "New" for set context + open new scene.

I assume that one would open up the tool only for following cases :
1. Changing work space to start something new.
2. Changing work space to open up old scene.
2. Save the file, but change work space if other asset/task selected on save.

So if there's a "New" button for opening new scene, I think there wouldn't need double clicking to set context and the Breadcrumb.

And in the end, I think artist will no need to be aware of the concept of "context" when using this tool as long as there's only "New", "Open", "Save" these there buttons being presented.

avalon/pipeline.py
```diff
@@ -926,61 +926,103 @@ def get_representation_context(representation):
...
+                                          "type": "asset"})
+            assert asset_document, "Asset must exist"
+
+        changes["AVALON_ASSET"] = asset
+
+        # Update silo
+        changes["AVALON_SILO"] = asset_document["silo"]
+
+        # Update hierarchy
+        parents = asset_document['data'].get('parents', [])
```
BigRoy: This code has moved and fixes a BUG where previously if asset had not changed with the call then `asset_document` would not have been a set variable.

avalon/pipeline.py
```diff
@@ -926,61 +926,103 @@ def get_representation_context(representation):
     return context
 
 
-def update_current_task(task=None, asset=None, app=None):
-    """Update active Session to a new task work area.
+def compute_session_changes(session, task=None, asset=None, app=None):
```
BigRoy: I've made this a new function to compute the changes to the current session required to become a new valid session. Before we only had `update_current_task` which did the same but would *always* change the active session. That computing of the required changes is now moved to this separate function.

We might want to compute up with a better name and maybe expose it in the API?

avalon/pipeline.py
```diff
@@ -926,61 +926,103 @@ def get_representation_context(representation):
...
+
+    if task:
+        changes["AVALON_TASK"] = task
+
+    if app:
+        changes["AVALON_APP"] = app
+
+    # Update silo and hierarchy when asset changed
+    if asset:
+        if isinstance(asset, dict):
```
BigRoy: This has now additionally been allowed to be passed a database document for an asset as opposed to only working by its name. This allows to avoid the additional query to the database.

avalon/pipeline.py
```diff
@@ -926,61 +926,103 @@ def get_representation_context(representation):
...
+        parents = asset_document['data'].get('parents', [])
+        hierarchy = ""
+        if len(parents) > 0:
+            hierarchy = os.path.sep.join(parents)
+        changes['AVALON_HIERARCHY'] = hierarchy
 
     # Compute work directory (with the temporary changed session so far)
     project = io.find_one({"type": "project"},
                           projection={"config.template.work": True})
     template = project["config"]["template"]["work"]
```
BigRoy: It could be worth checking whether we might want to cache this call to project because usually as you're working in a project the Work files template should be a static concept (not change during the project runs) and since we're unable to change project in-app this means this template would never change. Maybe we can cache it on the first ever call. (Optimization only)

avalon/tools/widgets.py
```diff
@@ -120,6 +120,11 @@ def get_active_asset(self):
         current = self.view.currentIndex()
         return current.data(self.model.ObjectIdRole)
 
+    def get_active_asset_document(self):
+        """Return the asset id the current asset."""
+        current = self.view.currentIndex()
+        return current.data(self.model.DocumentRole)
```
BigRoy: Solely added this to make it very trivial to retrieve the currently active asset's document as opposed to its name.

avalon/tools/workfiles/app.py
```diff
@@ -3,86 +3,120 @@
...
-        self.label = QtWidgets.QLabel("File name")
-        grid_layout.addWidget(self.label, 2, 1)
+        self.result = None
+        self.host = api.registered_host()
+        self.root = root
+        self.work_file = None
+
+        if session is None:
+            # Fallback to active session
+            session = api.Session
```
BigRoy: Fallback could be removed and `session` made a required argument for the Window since it's only used by the Work Files tool and it will always explicitly pass a session, or at least it should.

avalon/tools/workfiles/app.py
```diff
@@ -166,10 +212,18 @@ def refresh(self):
             template = self.get_work_file(template)
             template = "^" + template + "$"           # match beginning to end
 
+            # Match with ignore case on Windows due to the Windows
+            # OS not being case-sensitive. This avoids later running
+            # into the error that the file did exist if it existed
+            # with a different upper/lower-case.
+            kwargs = {}
+            if platform.system() == "Windows":
+                kwargs["flags"] = re.IGNORECASE
```
BigRoy: This is a fix for a bug I found when there were matching files with the template however they had a different case (uppercase vs lowercase) - previously it would ignore the match and it could reach the "This a bug, file exists.." assertion below. That is patched by (on Windows) matching without case sensitivity because `os.path.exists` would detect is as existing.

avalon/tools/workfiles/app.py
```diff
@@ -186,201 +240,261 @@ def refresh(self):
...
+        self._task = None
+        self.root = None
+        self.host = api.registered_host()
+
+        widgets = {
+            "list": QtWidgets.QListWidget(),
+            "duplicate": QtWidgets.QPushButton("Duplicate"),
+            "open": QtWidgets.QPushButton("Open"),
+            "browse": QtWidgets.QPushButton("Browse"),
+            #"currentFile": QtWidgets.QLabel(),
```
BigRoy: Currently unused and hidden to simplify the UI. Showing current file name seems redundant since usually the Host would already show it. Or maybe the current file can be more easily show elsewhere as a standard, e.g. like in the window title. Would love to hear thoughts about this.

avalon/tools/workfiles/app.py
```diff
@@ -186,201 +240,261 @@ def refresh(self):
...
+            "save": QtWidgets.QPushButton("Save As")
+        }
+
+        # Build buttons widget for files widget
+        buttons = QtWidgets.QWidget()
+        layout = QtWidgets.QHBoxLayout(buttons)
+        layout.setContentsMargins(0, 0, 0, 0)
+        #layout.addWidget(widgets["duplicate"])
```
BigRoy: Removed "duplicate" button in favor of adding it to the Right Mouse context menu in the Files Widget _TODO_ also see #489

avalon/tools/workfiles/app.py
```diff
@@ -186,201 +240,261 @@ def refresh(self):
...
+        self._task = task
+
+    def _get_session(self):
+        """Return a modified session for the current asset and task"""
+
+        session = api.Session.copy()
+        # todo: expose this in the API?
+        changes = pipeline.compute_session_changes(session,
+                                                   asset=self._asset,
+                                                   task=self._task)
```
BigRoy: Should we expose this in API? Or is call to `pipeline.xx` fine inside the tools? I suppose it's fine?

avalon/tools/workfiles/app.py
```diff
@@ -439,27 +588,225 @@ def on_browse_pressed(self):
...
+        self.refresh()
 
-        self.close()
+    def refresh(self):
+        """Refresh listed files for current selection in the interface"""
+
+        # Refresh current scene label
+        #filepath = self.host.current_file()
+        #current = os.path.basename(filepath) if filepath else "<unsaved>"
+        #self.widgets["currentFile"].setText("Current File: %s" % current)
```
BigRoy: Disabled widget, see comment above.

avalon/tools/workfiles/app.py
```diff
@@ -439,27 +588,225 @@ def on_browse_pressed(self):
...
+        if "task" in context:
+            self.widgets["tasks"].select_task(context["task"])
+
+    def refresh(self):
+
+        # Refresh asset widget
+        self.widgets["assets"].refresh()
+
+        # Refresh breadcrumbs
+        #self.widgets["header"].set_session(session)
```
BigRoy: Disabled widget, see comments above.

avalon/tools/workfiles/app.py
```diff
@@ -439,27 +588,225 @@ def on_browse_pressed(self):
...
+        self.widgets["files"].setEnabled(all([bool(task), bool(asset)]))
+
+        files = self.widgets["files"]
+        files.set_asset_task(asset, task)
+        files.refresh()
+
+
+def show(root=None, debug=False, parent=None, use_context=True):
     """Show Work Files GUI"""
+    # todo: remove `root` argument to show()
```
BigRoy: Since now the Work Files tool can browse to different assets and tasks setting an explicit root makes no sense, since the root is computed based on asset/task work root.

As such, we can remove the `root` argument. Should we preserve this for API compatibility?

BigRoy:
With https://github.com/getavalon/core/pull/442/commits/7ad657a075856b38dd95c21fcfbcbe8fa0060a4c I've pushed the new prototype functionality. It's not flawless yet but it works to some extent. 

I've tried to comment on most of the important/big changes or odd choices above. Hope that helps reviewing what's there so far.

There's a lot to potentially check:

- #465 this crash might have been reintroduced with https://github.com/getavalon/core/pull/442/commits/cef28803bbd604507470f34539f54e84d6aaa4fd since I "patched" it different in the conflict merge, I set the parents explicitly as opposed to storing the variables. This should result in the same behavior in the end. So if someone is able to check potential Blender crashes that would be helpful. @jasperges 
- Does all the functionality work across different hosts, e.g. Maya, Nuke, Fusion, Houdini, etc.
- Please check and comment on other odd code changes.
- The [failed Travis check](https://travis-ci.org/getavalon/core/builds/626365702?utm_source=github_status&utm_medium=notification) is due to #492 -> PR #494

jasperges:
@BigRoy This new version almost works fine in Blender. If you open a workfile the Qt interface is locked and the only way to get it back or close it to open it again from within Blender.
There used to be an explicit `self.close()`. Did you remove that on purpose? Is the UI supposed to stay open or should it close, like it used to?

BigRoy:
Regarding the issue @jasperges mentioned for Blender. I checked it briefly with him. This isn't actually a problem with this particular interface but with the way the Qt interfaces are running there in general. Whenever any Qt interface is open with the Modal Operator and you open a file then it'll hang the Qt application. As such, on scene open with this Work Files tool it shows that issue.

> Did you remove that on purpose? Is the UI supposed to stay open or should it close, like it used to?

In this case I did leave it open on purpose and maybe wanted to add a toggle "autoclose" or some setting. But I'm happy to also force either of the one. It's just much easier to debug without the auto-closing, and additionally some Artists mumbled it a couple of times whether it could be worth looking into just having it open as "explorer" when they need to check some scene from different assets, so I removed the automatic closing on purpose for now.

BigRoy:
I've been working on tweaking the interface for the Work Files some more:
afbeelding

This is based on some of @mkolar 's [comments](https://github.com/getavalon/core/issues/489#issuecomment-567540981) on #489 and in particular implementing point 3-5. Specifically having date modified there too, allow sorting by it and having a filename filter at the top.

I've also moved the Save As button to be next to the other ones, this allows for more screen space for listing files and makes the button sizes more consistent. @jasperges do you think this looks better?

I'll need to do some more changes to the code to have the file open/save functionality working again as I just quickly threw together the UI changes but haven't yet accounted for the change to Model+View as opposed to the previous QListWidget for the files. I think I can push it Monday.

BigRoy:
I just pushed https://github.com/getavalon/core/pull/442/commits/d7130433f2afd2c7f32643d35f25267cf5e551ed which implements the UI as shown in [this comment](https://github.com/getavalon/core/pull/442#issuecomment-568186619).

- It also has Duplicate in a right mouse context menu now.
- It also selects the work file with highest date modified, like with the original work files tool.

Artists on our end also requested to add a Filesize column. Any remarks regarding that? Might it get too detailed or cluttered? Any reasons not to add it?

[Like before](https://github.com/getavalon/core/pull/442#issuecomment-566733107) the open TODOs:
- **Set Context functionality**: TODO
- **Initialize workdir**: TODO

avalon/maya/workio.py
```diff
@@ -30,10 +30,27 @@ def current_file():
...
+    workspace_mel = os.path.join(work_dir, "workspace.mel")
+    if os.path.exists(workspace_mel):
+        scene_rule = 'workspace -fr "scene" '
+        # We need to use builtins as `open` is overridden by the workio API
+        open_file = __builtins__["open"]
+        with open_file(workspace_mel, "r") as f:
+            for line in f:
+                if line.strip().startswith(scene_rule):
+                    remainder = line[len(scene_rule):]      # == "rule";
+                    scene_dir = remainder.split('"')[1]     # == rule
```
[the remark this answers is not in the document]
davidlatwe: > But somehow when scrolling through the asset list and it doing that many times after each other often crashed Maya.

Can we add this as note comment in code ? Was afraid that I might forgot why doing this way instead of using Maya command. :baby:

avalon/nuke/workio.py
```diff
@@ -42,6 +42,11 @@ def current_file():
     return os.path.normpath(current_file).replace("\\", "/")
 
 
-def work_root():
-    from avalon import Session
-    return os.path.normpath(Session["AVALON_WORKDIR"]).replace("\\", "/")
+def work_root(session):
+
+    if scene_dir:
+        path = os.path.join(work_dir, scene_dir)
```
davidlatwe: I believe that `scene_dir` and `work_dir` are undefined. :wink:

avalon/pipeline.py
davidlatwe: Or not expose `compute_session_changes` in API, but add a new arg called `dry_run=False` to `update_current_task` ?

If `dry_run` set to `True`, then return early without update.

avalon/pipeline.py
```diff
@@ -926,61 +926,111 @@ def get_representation_context(representation):
...
     mapping = {
         "AVALON_ASSET": asset,
         "AVALON_TASK": task,
         "AVALON_APP": app,
     }
-    changed = {key: value for key, value in mapping.items() if value}
-    if not changed:
-        return
+    changes = {key: value for key, value in mapping.items() if value
+               and value != session.get(key)}
```
davidlatwe: This is trivial, but my IDE is highlighting this line because of PEP8 (line break before binary operator).

avalon/pipeline.py
davidlatwe: +1, maybe cache it on App launch.

avalon/pipeline.py
davidlatwe: And maybe need to add one more *internal* arg called `_changes=None` to `update_current_task`, so we can pass the `changes` that returned from `compute_session_changes` into `update_current_task` to avoid double work.

avalon/tools/workfiles/app.py
davidlatwe: Just pop a warning if `root` in `kwargs` ?

avalon/pipeline.py
BigRoy: So would it then be like this? 
```python
changes = {key: value for key, value in mapping.items() 
           if value and value != session.get(key)}
```

avalon/nuke/workio.py
BigRoy: Thanks, well spotted. Where is the Hound? 🐶

avalon/pipeline.py
davidlatwe: Yes, that would work !

avalon/pipeline.py
BigRoy: The `dry_run` argument does sound like a decent alternative to having this in a separate function. @mkolar @mottosso any preference on this one? Separate functions like this or make one and add `dry_run` argument?

I'm not too big a fan of adding the `_changes` internal argument. It seems like a premature optimization that for now seemingly adds redundant complexity.

avalon/pipeline.py
BigRoy: Fixed with https://github.com/getavalon/core/pull/442/commits/3d7c65e1c3aecc92fa4a0a08a174af2c940df8de

avalon/nuke/workio.py
BigRoy: Fixed with: https://github.com/getavalon/core/pull/442/commits/83617b63544ccb0336a8afc2726fd2e8dfacb69f

avalon/maya/workio.py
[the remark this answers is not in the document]
BigRoy: Added a comment with: https://github.com/getavalon/core/pull/442/commits/2f4aa80808010be8f8dfc10bdac54c7a05ad0a85

jasperges:
@BigRoy Yes, I think this is better. Every button the same size and a bit more room for the files list.

About adding a filesize column: I don't mind it that much, although it will get a bit cluttered. Maybe you can add the functionality so users can choose which columns to show (right mouse menu), just like in a regular file browser. Then everybody can choose for themselves what to show.

BigRoy:
> About adding a filesize column: I don't mind it that much, although it will get a bit cluttered. Maybe you can add the functionality so users can choose which columns to show (right mouse menu), just like in a regular file browser. Then everybody can choose for themselves what to show.

That could definitely be nice, but it does add some more complexity which I currently think we don't need yet. Because it'd mean we'd need to store the preferences of the user too (otherwise they always need to set their settings again). Easiest would be then to store per machine, but then we might want to have it per user - where there is no User-specific _yet_ in Avalon. I think for now it's best if we can manage to have these tools as simple as possible as it'll be easier to maintain.

jasperges:
> That could definitely be nice, but it does add some more complexity which I currently think we don't need yet. Because it'd mean we'd need to store the preferences of the user too (otherwise they always need to set their settings again). Easiest would be then to store per machine, but then we might want to have it per user - where there is no User-specific _yet_ in Avalon. I think for now it's best if we can manage to have these tools as simple as possible as it'll be easier to maintain.

I completely agree with keeping it simple.
The problem for me with more columns is, that it is already about preference. Not everybody has the same preference. So how do you determine what is a good baseline for everybody and when it's really a user preference?
Apart from adding the functionality to view extra columns, can't you easily store the preferences in `QSettings`?

Again: for me it's fine to add the filesize column, but I would definitely draw the line there. If an artist wants to see more, they can always use the 'Browse' button and get more info there...

BigRoy:
> can't you easily store the preferences in QSettings?

Yes, you can. They would however be per machine and not per 'avalon user' (which is not a thing yet I suppose).

> The problem for me with more columns is, that it is already about preference. Not everybody has the same preference. So how do you determine what is a good baseline for everybody and when it's really a user preference?

Agreed. I'll leave it as is without the filesize column. If it keeps coming up with the team then we could also look into our options for a future issue.

jasperges:
> @BigRoy This new version almost works fine in Blender. If you open a workfile the Qt interface is locked and the only way to get it back or close it to open it again from within Blender.
> There used to be an explicit `self.close()`. Did you remove that on purpose? Is the UI supposed to stay open or should it close, like it used to?
> 
> To keep it working, I quickly added `module.window.close()`.

To come back to this: I changed the way Blender calls `QApplication.processEvents()`. This is not an issue anymore.

BigRoy:
### Initialize Work Area

Here's the **Initialize Work Area** in action. Does this look like the way it should be? It doesn't feel as smooth as I'd like but I can't think of anything that's more clear than this to the end user.

@jasperges @mkolar @davidlatwe any pointers?

workfiles_init_workarea

This resolves the **Initialize workdir: TODO**

### Remove Set Context functionality

I have removed the "double click on task" functionality to explicitly set the current `api.Session` to another context than the current work file. Instead, just saving or opening in another context will first switch the context.

That way accidentally switching contexts is near impossible, and also avoids the complexity for the end user. @mkolar would that still allow all your use cases to happen?

This current avoids the need for **Set Context functionality: TODO**

---

This is now in a usable state. 

- Please remark any open comments on code style. 
- If you have time, please test it. Try to break it too please.
- Confirm the App work directory initialization works on your end.

tokejepsen:
Any reason why work area creation cant happen on `Save As`?

BigRoy:
> Any reason why work area creation cant happen on Save As?

@tokejepsen Good question. Technically it could. Then before work area initialization it would just show no files (and maybe the model would just have a non-selectable placeholder "Work Area does not exist" and *open* and *browse* get disabled. It might be somewhat "confusing" though if you find out that the folder doesn't exist at all yet. 

Would you want that to happen instead? What do others think? It's not a bad idea.

Aside of clarity the only reason I'm thinking of is that I'm not too sure how well App Work Directory initialization will work for others who launch with a different means than the Launcher. Specifically because it requires the "Application" name to be available and allow to trigger a `.toml` file for it.

tokejepsen:
> It might be somewhat "confusing" though if you find out that the folder doesn't exist at all yet.

Have noticed that artists hardly navigate the folder structure anymore when `Workfiles` is working.

> Would you want that to happen instead?

Personally I would like to see necessary actions like `work area` creation hidden from artists, and simplify/speed up their workflow > Less button presses.

BigRoy:
> Personally I would like to see necessary actions like work area creation hidden from artists, and simplify/speed up their workflow > Less button presses.

Agreed. Here's the updated version.

workfiles_init_workarea_v2

davidlatwe:
Will have a good look and test later today !

davidlatwe:
So I have tested, and great work !

But, I would prefer having an extra button for initializing work area as the one in this [comment](https://github.com/getavalon/core/pull/442#issuecomment-571177579). :relaxed: 
Because if I am starting from an empty scene, it's a bit weird to save anything for getting a workspace.

Or, how about just hide that *"work area does not exists..."* item, and auto create work area when *Save As* is pressed. If the *Browse* button is pressed, pop a warning dialog.

Is there any reason that one need to explicitly initialize work area ?

Oh and, would be better if the task view can display "Task" in header instead of "name". :)

BigRoy:
@davidlatwe @tokejepsen what if we just swap the bottom buttons Open, Browse, Save with one big button "Create Work Directory" at the bottom. That way the change to UI is not as odd as in my [first draft](https://github.com/getavalon/core/pull/442#issuecomment-571177579)?

@mkolar what's your thought on this subject?

tokejepsen:
I would avoid any unnecessary mouse clicks from the user. The main point of making the Workfiles app, was to make it faster to load/save your workfiles so the user's preference wasn't the DCCs native load/save workflow.

> Because if I am starting from an empty scene, it's a bit weird to save anything for getting a workspace.

I'm not sure I follow this @davidlatwe ?
If you have an empty scene and need to create a workspace, then wouldnt you have to save your file?

davidlatwe:
> If you have an empty scene and need to create a workspace, then wouldnt you have to save your file?

Yeah, that's true. I think I kept mixing the usage of context manager app when I imaging the workflow of workfile app...

> The main point of making the Workfiles app, was to make it faster to load/save your workfiles so the user's preference wasn't the DCCs native load/save workflow.

I think you are right. Yeah, I could go with "Save As" to init work area if not exists. :)

mkolar:
I actually think that it's very clear in this implementation. If the folder doesn't exists, `save as` will create it alongside the actual file. quick and clean. No extra clicks.

BigRoy:
> Yeah, I could go with "Save As" to init work area if not exists. :)

@davidlatwe does that mean the current implementation is good to go?

---

@mkolar @tokejepsen did you try the code on your end too? Or was it just a comment based on the GIF?

---

If everyone approves I'd still like to do the following before merging
- Delete `avalon.tools.contextmanager`
- Remove context manager from all host implementations.
- Rename column header `name` to `Tasks` in the second list seen [here](https://github.com/getavalon/core/pull/442#issuecomment-571208616)

tokejepsen:
> did you try the code on your end too? Or was it just a comment based on the GIF?

Based on the GIF.

> If everyone approves I'd still like to do the following before merging

Sounds like a good clean. Nice!

davidlatwe:
One little thing, the "Work area does not exists.." menu item is able to show context ment on right click, which shouldn't be. If clicked the menu's "Duplicate" action, error raised.

BigRoy:
> "Work area does not exists.." menu item is able to show context ment on right click, which shouldn't be. If clicked the menu's "Duplicate" action, error raised.

Well caught. Will need to patch that up, thanks!

I also noticed that, because it's a regular Item, whenever you would have a name filter set top right that you could be filtering this placeholder item too. Not a major issue but something to watch out for. Probably will need to see if we can easily disable the filter then.

jasperges:
@BigRoy Any update on this? Can I help in any way?

BigRoy:
> Any update on this? Can I help in any way?

Thanks for tagging me! :) I'm a bit swamped with other stuff so haven't been able to put more time into this. You should be able to push into this PR (or if not set up a PR to my branch) to fix any outstanding issues that you think you can easily adapt for.

It's likely I won't have the time until end of next week to continue myself.

jasperges:
No rush. I was just wondering. Just merge this to our production branch so we can finally start using Avalon! :rocket: 

Personally I'm not aware of issues (apart from the _excepthook_, but that's part of another [discussion](../issues/512).

I'll be patient or if I encounter issues, I will see what I can do. :)

iLLiCiTiT:
I wonder, what is missing and needs to be done?

mottosso:
If I'm not mistaken, it looks like what's missing is someone hitting that `Merge pull request` button.

Any additional fixes or features seem well suited for an additional PR. First one to press it wins 10 units of Avalon currency! Oh and you must also make a release for the new version it makes.

BigRoy:
[This would be on the to do (or a merge and posted as an Issue)](https://github.com/getavalon/core/pull/442#issuecomment-573508628). And the [bottom bullet points of this comment](https://github.com/getavalon/core/pull/442#issuecomment-573411893).

Most seem like small notes. Feel free to do a quick fix or otherwise put them in a separate PR if you consider the current state save enough.

Go get them Avalon currency y'all.

davidlatwe:
Context label has been removed in commit be07276, was there a replacement ? 0.0

iLLiCiTiT:
@davidlatwe True that's my fault. But where to put context label when `Edit Context..` action was removed?

davidlatwe:
Can we change them into normal menu item instead ? Or keep the context manager for now, until artists get familiar with new workfile app ?

iLLiCiTiT:
I'm not user so my word is not much valuable at this point, but I would rather add empty action with context label and without callback instead of keep context manager, because me as user wouldn't understand why he is there if I can do it through workfiles.

jasperges:
I'm also not a user, but I think it's a good thing it has been removed, because I don't think it's ever needed to switch the context without loading or saving a work file. (Although we probably have the need to publish several character rigs from one base rig. So inside the file we have to run the Creator for different assets... but that is an exception, not the rule.)
I think it's a good idea to show the current context somewhere, but inside a menu feels like the wrong place for that. Not sure where it should go, though...

tokejepsen:
Just tested this out myself and it works nicely!

One question did pop up. If there arent any task assigned to the asset, you wont be able to save a workfile, so you would have to open the project manager to create one.
Maybe its taking too much responsibility on, but should the user be able to create tasks on assets through this UI?

jasperges:
Personally I don't think so. I think most of the time this is not something an artist should do (a production manager/coordinator should do this or automated from Shotgun/FTrack/Kitsu) and definitely not in the workfiles app.

Just my two cents.

BigRoy:
So, does this mean it's ready to merge? 🔥🚀

tokejepsen:
> So, does this mean it's ready to merge?

Except for the hound violations?

BigRoy:
> Except for the hound violations?

Was just editing my comment to state that, yes. I can't seem to find the open Hound violations...

tokejepsen:
Ahh, fair enough. Ready to merge!

davidlatwe:
But what about the context label ? :O

I can agree that context label is not good to be in side the menu, but it's a place that work for most of the DCC hosts for now to indicate current context, right ?

tokejepsen:
How about changing the window title?

davidlatwe:
> How about changing the window title?

That did cross my mind but should be in a separate issue/PR to discuss.

Anyway, if t's just me, then maybe it's okay to be merged. Either add back context label as a  normal menu item, or merge it as is, both work for me now. ☺️ 

Still, IMHO, deprecating context manager with new workfile App is one thing, removing context indicator altogether is another, so I think this kind of UI change should be brought up earlier. 💡

iLLiCiTiT:
I've tried to add not enabled action item to menu PR: https://github.com/BigRoy/core/pull/5

BigRoy:
> Still, IMHO, deprecating context manager with new workfile App is one thing, removing context indicator altogether is another, so I think this kind of UI change should be brought up earlier. 

Totally agree that these shouldn't be taken lightly. I just envisioned that to be part of it and assumed the discussion was already debating that too.

Anyway, shall we include https://github.com/BigRoy/core/pull/5 and then open an additional issue to track any better ways to approach that @davidlatwe ?

davidlatwe:
> Anyway, shall we include https://github.com/BigRoy/core/pull/5 and then open an additional issue to track any better ways to approach that @davidlatwe ?

That would be great :D

Just pulled and tested that branch, and only require [one tiny fix](https://github.com/BigRoy/core/pull/5/files#r378741747) and good to go for me !

iLLiCiTiT:
> Just pulled and tested that branch, and only require one tiny fix and good to go for me !

Just pulled fix. Can you check please?

davidlatwe:
Perfect !! 🚀

BigRoy:
Ready for take-off!

jasperges:
I've set `AVALON_SCENEDIR` to `'scenes'` for Maya. But using this new workfiles app, I get an error when initializing the work directory (that doesn't exist yet). I've add `default_dirs = ["scenes"]` to the `maya.toml`, but still no luck. Am I missing something?

davidlatwe:
If you are using Avalon Launcher, maybe you need to head back to first page or restart the launcher so the environment vars get update from those `.toml` change.

iLLiCiTiT:
Just realized small detail... Will that context label change on context change?

davidlatwe:
Yeah, it did change in my test, and still does.

davidlatwe:
@jasperges  is everything working there now ?

jasperges:
@davidlatwe On Monday I can check again. I don't have access to Maya at home. Will let you know.

jasperges:
@davidlatwe All is good! I still have some small issues, but I'm pretty sure that's because of the Avalon configuration. So, go, go, go! :rocket:

davidlatwe:
Merging this !

PR #443, opened 2019-09-10
Fix#246 : Store and lock the "family" on a "subset"

### LINKED ISSUE #246: Store on and lock the of "family" for a "subset"

davidlatwe:
This PR aim to resolve #246, the implementation I am proposing is to involve new schema for subset and version, so that we could have a standard to maintain backwards compatibility.

**What's changed?**
* New schema files, `subset-3.0` and `version-3.0`
* In Loader App and Scene Inventory, get `data.families` from **subset** document if the `schema` is `subset-3.0`

### DIFF
--- a/avalon/tools/sceneinventory/model.py
+++ b/avalon/tools/sceneinventory/model.py
@@ -228,7 +228,11 @@ class InventoryModel(TreeModel):
             # Get the primary family
             family = version["data"].get("family", "")
             if not family:
-                families = version["data"].get("families", [])
+                if subset["schema"] == "avalon-core:subset-3.0":
+                    families = subset["data"]["families"]
+                else:
+                    families = version["data"].get("families", [])
+
                 if families:
                     family = families[0]
 
--- a/avalon/pipeline.py
+++ b/avalon/pipeline.py
@@ -1294,7 +1294,11 @@ def is_compatible_loader(Loader, context):
         bool
 
     """
-    families = context["version"]["data"]["families"]
+    if context["subset"]["schema"] == "avalon-core:subset-3.0":
+        families = context["subset"]["data"]["families"]
+    else:
+        families = context["version"]["data"]["families"]
+
     representation = context["representation"]
     has_family = ("*" in Loader.families or
                   any(family in Loader.families for family in families))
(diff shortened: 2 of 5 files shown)


davidlatwe:
What I have tested:
[x] Publish with new subset schema (Modified config's database integrator plugin)
[x] Load in both new and old schema subsets to one scene
[x] Check Scene Inventory App display them correctly

avalon/tools/sceneinventory/model.py
```diff
@@ -228,7 +228,11 @@ def add_items(self, items, parent=None):
             # Get the primary family
             family = version["data"].get("family", "")
```
BigRoy: @davidlatwe how about we push this into the schema if/else too. Because in this case the `version.data["family"]` will still overwrite the `subset["data"]["families"]` for the primary family although it should actually be the other way around.

davidlatwe: Ah, so if the schema is `subset-3.0`, then just take from `subset["data"]["families"]`, and consider `version["data"]["family"]` if not.
Yeah, this is better !

davidlatwe: Okay, I have this fixed in commit 0c548f8, and the readability should be suffice I think :)

Issue #179, opened 2017-06-12
Highlight new or extended subsets


mottosso:
### Goal

Let the user know when an asset is being updated or created anew. For example, if `rigDefault` has been created elsewhere, color the line edit orange. Otherwise, color it green.

### Motivation

It can be difficult to know whether you are accidentally updating something that already exists, or creating something new when you meant to update. Especially if working on the same subset in more than a single scene.

### Implementation

Apply color to represent existence.

![untitled](https://user-images.githubusercontent.com/2152766/27023472-97e93c92-4f4a-11e7-95ec-21ea6bbca622.gif)

PR #444, opened 2019-09-11
Implement#179

davidlatwe:
This PR implement #179. But instead of changing subset `QLineEdit` widget's background color, I chose to use `border-color` for better UX on text editing.

**What's changed?**
* Avalon Creator tool's UX on subset naming

#### Demo GIF
creator

### DIFF
--- a/avalon/tools/creator/app.py
+++ b/avalon/tools/creator/app.py
@@ -227,6 +227,7 @@ class Window(QtWidgets.QDialog):
                 subset_name = subset_name[0].upper() + subset_name[1:]
             result.setText("{}{}".format(family, subset_name))
 
+            # Indicate subset existence
             if not subset_name:
                 subset.setStyleSheet("border-color: #3A3939;")
             elif subset_name in existed_subsets:
--- a/avalon/tools/creator/app.py
+++ b/avalon/tools/creator/app.py
@@ -227,6 +227,13 @@ class Window(QtWidgets.QDialog):
                 subset_name = subset_name[0].upper() + subset_name[1:]
             result.setText("{}{}".format(family, subset_name))
 
+            if not subset_name:
+                subset.setStyleSheet("border-color: #3A3939;")
+            elif subset_name in existed_subsets:
+                subset.setStyleSheet("border-color: #BB6F4E;")
+            else:  # New subset
+                subset.setStyleSheet("border-color: #7AAB8F;")
+
             item.setData(ExistsRole, True)
             self.echo("Ready ..")
         else:


BigRoy:
Just wondering whether we can improve on this slightly further. I checked with an artist and asked "what do you think this orange/green highlighting means"? His reply was that he thought he orange line meant he must've typed symbols or characters that are invalid which unfortunately isn't what we're trying to state with the outline. As such, we should somehow clarify what the coloring means or show it differently to the user.

Some ideas:
1. Add a tooltip so that when the Artists hovers over it at least he understands what's up. However, not everyone will do this and it might still be missed.
2. Maybe adding icons could help clarify what is up, e.g. a [refresh](https://fontawesome.com/v4.7.0/icon/refresh) kind of symbol when it overwrites and a [plus](https://fontawesome.com/v4.7.0/icon/plus) when it's a completely new one. Again, this could still be misinterpreted.
3. Maybe instead of orange we can pick a color that indicates slightly less danger?
4. The artist mentioned it would be clearer to him if the Creator would show a message on the moment when you're trying to create it. E.g. "This subset has already been published for this asset before. Are you sure you want to generate publishes that will publish new versions for it? Yes/no?" -> Most clear, but a bit interuptive. Not the biggest fan of it, but it describes how clear it hopefully would be in the end as to what it's doing.

avalon/tools/creator/app.py
```diff
@@ -227,6 +227,13 @@ def _on_data_changed(self):
                 subset_name = subset_name[0].upper() + subset_name[1:]
             result.setText("{}{}".format(family, subset_name))
 
+            if not subset_name:
+                subset.setStyleSheet("border-color: #3A3939;")
```
BigRoy: Is this the case of having no border-color whatsoever? If so, then wouldn't `subset.setStyleSheet("")` do the trick to reset any overrides?

tokejepsen:
> Just wondering whether we can improve on this slightly further. I checked with an artist and asked "what do you think this orange/green highlighting means"? His reply was that he thought he orange line meant he must've typed symbols or characters that are invalid which unfortunately isn't what we're trying to state with the outline. As such, we should somehow clarify what the coloring means or show it differently to the user.

Hadn't read the issue and just saw the GIF and had the same thoughts as the artist.

> The artist mentioned it would be clearer to him if the Creator would show a message on the moment when you're trying to create it. E.g. "This subset has already been published for this asset before. Are you sure you want to generate publishes that will publish new versions for it? Yes/no?" -> Most clear, but a bit interuptive. Not the biggest fan of it, but it describes how clear it hopefully would be in the end as to what it's doing.

Maybe inform the user in the status bar, where it currently says "Ready"?

avalon/tools/creator/app.py
davidlatwe: I did not see that ! Fixed in commit 4726466

davidlatwe:
Here's the changed demo :)
I though a bluish color may be a good fit, since both scenario are acceptable actions, so just picking another color that does not have the impression of danger. And improve the indication by the echo message.

creator

avalon/tools/creator/app.py
```diff
@@ -227,8 +227,19 @@ def _on_data_changed(self):
                 subset_name = subset_name[0].upper() + subset_name[1:]
             result.setText("{}{}".format(family, subset_name))
 
+            # Indicate subset existence
+            if not subset_name:
+                subset.setStyleSheet("")
+                message = "Empty subset name .."
+            elif subset_name in existed_subsets:
+                subset.setStyleSheet("border-color: #4E76BB;")
+                message = "Existed subset, appending next version."
```
tokejepsen: This should probably be "Existing subset, appending next version".

davidlatwe: Solved !

PR #447, opened 2019-09-12
Optimize AssetModel refresh time

BigRoy:
### Issue

There have been reports of the Asset Widget being slow to refresh on bigger projects. This resulted in slow launch times of the **Loader** and **Project Manager** for listing the assets.

*Where large project is defined as for example 2 seasons of 20 episodes, with each 30 shots. Resulting in 2 * 20 * 30 = 1200 assets and [there have been reports of very slow load times](https://gitter.im/getavalon/Lobby?at=5d67cd66f94e7d10a16125dd)

### What has changed?

This is an optimization of the method that builds the hierarchy in the asset widget that is much faster for large projects. The way the assets are queried is now optimized to do less queries against the database, allowing it do more caching (as it will hit the same query more often) and have less problems with any delay in the connection with the database as there are much less calls to be made.

Funnily enough, it's also much faster for small projects. Win-win. 🚀 

#### Indication of speed changes

- Tested myself: 0.2790s -> 0.0055s
- Reported by @iLLiCiTiT: 4.5s -> 0.01s
- Reported by @tokejepsen: 10x speed improvement

---

_Thanks to the team at [pypeclub](https://github.com/pypeclub) and @iLLiCiTiT to quickly separate out some commits related to the optimization._

### DIFF
--- a/avalon/tools/models.py
+++ b/avalon/tools/models.py
@@ -319,13 +319,16 @@ class AssetModel(TreeModel):
     def _add_hierarchy(self, assets, parent=None):
         """Add the assets that are related to the parent as children items.
 
+        This method does *not* query the database. These instead are queried
+        in a single batch upfront as an optimization to reduce database
+        queries. Resulting in up to 10x speed increase.
+
         Args:
             assets (dict): All assets in the currently active silo stored
                 by key/value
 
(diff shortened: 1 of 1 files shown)


BigRoy:
Merging this. 👍

PR #448, opened 2019-09-12
Fix#289: Validate subset name for database

### LINKED ISSUE #289: Creator does not ensure the subset has only alphabetical characters
Currently the instance creator does not ensure the subset has a name consisting only of alphabetical characters, any symbol, space or special characters seems to be allowed in the Creator.

Is this ok to be generic and abstract? Or should it never allow so to begin with? 

For example, should a space even be allowed? Or more specific, what characters _should_ be?

davidlatwe:
This PR resolve #289, Creator will ensure the subset name is valid for MongoDB.

**What's changed?**
* Validate subset name in Creator, and the GUI will indicate the validation result

### DIFF
--- a/avalon/tools/creator/app.py
+++ b/avalon/tools/creator/app.py
@@ -21,6 +21,91 @@ PluginRole = QtCore.Qt.UserRole + 5
 Separator = "---separator---"
 
 
+class SubsetNameValidator(QtGui.QRegExpValidator):
+
+    invalid = QtCore.Signal(set)
+    pattern = "^[a-zA-Z0-9_.]*$"
+
+    def __init__(self):
+        reg = QtCore.QRegExp(self.pattern)
+        super(SubsetNameValidator, self).__init__(reg)
+
+    def validate(self, input, pos):
+        results = super(SubsetNameValidator, self).validate(input, pos)
+        if results[0] == self.Invalid:
+            self.invalid.emit(self.invalid_chars(input))
+        return results
+
+    def invalid_chars(self, input):
+        invalid = set()
+        re_valid = re.compile(self.pattern)
+        for char in input:
+            if char == " ":
+                invalid.add("' '")
+                continue
+            if not re_valid.match(char):
+                invalid.add(char)
+        return invalid
+
+
+class SubsetNameLineEdit(QtWidgets.QLineEdit):
+
+    report = QtCore.Signal(str)
+    colors = {
+        "empty": (QtGui.QColor("#78879b"), ""),
+        "exists": (QtGui.QColor("#4E76BB"), "border-color: #4E76BB;"),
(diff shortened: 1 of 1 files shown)


davidlatwe:
I have change to use `QValidator` to block invalid inputs, also add a tool tip.
Here the result :

creator

Note that the invalid characters will not be accepted at all, so I add a border color transition animation for better feedback.

Please let me know what you think about this change :)

tokejepsen:
Looks great!

avalon/tools/creator/app.py
```diff
@@ -20,6 +21,91 @@
 Separator = "---separator---"
 
 
+class SubsetNameValidator(QtGui.QRegExpValidator):
```
BigRoy: To be sure, the only reason we require to define our own class is so we can report to the user which characters were invalidated in particular. Correct? If so, then this class is fine. Otherwise, this seems redundant.

avalon/tools/creator/app.py
```diff
@@ -20,6 +21,91 @@
...
+
+    def __init__(self, *args, **kwargs):
+        super(SubsetNameLineEdit, self).__init__(*args, **kwargs)
+
+        validator = SubsetNameValidator()
+        self.setValidator(validator)
+        self.setToolTip("Only alphanumeric characters (A-Z a-z 0-9), "
+                        "'_' and '.' are allowed.")
+
+        self._cstatus = None
```
BigRoy: Not sure if there's enough reason to abbrevaite `_color_status` to `_cstatus`. I don't mind it. Just wondering.

avalon/tools/creator/app.py
davidlatwe: > Correct?

Not entirely, the main reason I subclass the `QtGui.QRegExpValidator` was to override the function `validate`, so that I could emit a signal when the input has invalid character. I could not find any built-in way to implement this notification on `QLineEdit`, it just works silently.

Reporting which characters exactly were invalid was something I picked up on the way :)

avalon/tools/creator/app.py
davidlatwe: Hmm, this should be improved for reading.

avalon/tools/creator/app.py
BigRoy: Ah. If you only wanted to have the invalid signal, it's [`QLineEdit.inputRejected`](https://doc.qt.io/qt-5/qlineedit.html#inputRejected) but it won't signal what characters were invalid I believe + it's only been [introduced since Qt 5.12](https://code.qt.io/cgit/qt/qtbase.git/tree/dist/changes-5.12.0/?h=v5.12.0#n349) so I guess we can't rely on it.

mottosso:
I love it!

davidlatwe:
Well merge this tomorrow !

PR #450, opened 2019-09-17
Fix QPushButton::menu-indicator being cut off

BigRoy:
## Issue

The Instance Creator's subset dropdown button had it's "triangle" menu indicator cut off on the right hand side. This made it look very confusing and hard to understand. It was clearly wrong, see:

triangle

#### Menu Indicator cut off on right-hand side

This was due to the [`QPushButton::menu-indicator` styling in Avalon's stylesheet](https://github.com/getavalon/core/blob/c000ac5f07a8f849dd0ec0e9a9b81ff00fe521e4/avalon/style/style.qss#L1067).

This was reproducable on just a QPushButton example:
```python
# Assumes QApplication already runs
from avalon.vendor.Qt import QtWidgets
from avalon import style

box = QtWidgets.QPushButton("Test")
box.setStyleSheet(style.load_stylesheet())
menu = QtWidgets.QMenu(box)
menu.addAction("test")
box.setMenu(menu)
box.show()
```
Output:

#### Menu Indicator vertical alignment

Aside of that the vertical alignment of the triangle was also slightly odd, as it was [forced to the bottom right](https://github.com/getavalon/core/blob/c000ac5f07a8f849dd0ec0e9a9b81ff00fe521e4/avalon/style/style.qss#L1066). It's more expected to have this vertically centered on the button on the right-hand side.

Comparison: 

- Before with `bottom right`:

- After with `right`:

## What's changed?
The stylesheet is updated to not have the indicator cut off and have it vertically centered on right hand side.

---

Reference:

- [Discussion on Avalon Gitter](https://gitter.im/getavalon/Lobby?at=5d8103dea6f4896449204223)

### DIFF
--- a/avalon/style/style.qss
+++ b/avalon/style/style.qss
@@ -1064,13 +1064,16 @@ QToolButton::menu-arrow:open {
 QPushButton::menu-indicator  {
     subcontrol-origin: padding;
     subcontrol-position: right;
-    width: 7px;
-    height: 7px;
+    width: 8px;
+    height: 8px;
     padding-right: 5px;
 }
 
 QPushButton::menu-indicator:disabled {
     image: url(:/qss_icons/rc/down_arrow_disabled.png);
+    width: 9px;
+    height: 9px;
+    padding-right: 3px;
 }
 
 QTableView
--- a/avalon/style/style.qss
+++ b/avalon/style/style.qss
@@ -1064,7 +1064,6 @@ QToolButton::menu-arrow:open {
 QPushButton::menu-indicator  {
     subcontrol-origin: padding;
     subcontrol-position: bottom right;
-    left: 8px;
 }
 
 QTableView
--- a/avalon/style/style.qss
+++ b/avalon/style/style.qss
@@ -1063,7 +1063,7 @@ QToolButton::menu-arrow:open {
 
 QPushButton::menu-indicator  {
     subcontrol-origin: padding;
-    subcontrol-position: bottom right;
+    subcontrol-position: right;
 }
 
 QTableView
--- a/avalon/style/style.qss
+++ b/avalon/style/style.qss
@@ -1064,6 +1064,13 @@ QToolButton::menu-arrow:open {
 QPushButton::menu-indicator  {
     subcontrol-origin: padding;
     subcontrol-position: right;
+    width: 7px;
+    height: 7px;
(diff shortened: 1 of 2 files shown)


iLLiCiTiT:
that's huge triangle compared to text

BigRoy:
> that's huge triangle compared to text

Interesting. It didn't feel so off to me. But now comparing with other applications it does seem bigger than what's usually the case. Should we try and fix that up too? Anyone cares to have an opinion? :)

iLLiCiTiT:
you can try to add width and height to stylesheet (10px is random value):
```
QPushButton::menu-indicator  {
    width: 10px;
    height: 10px;
    subcontrol-position: right;
}
````

**Comment by davidlatwe** [comment]
I'd say we moving on for now :rocket: 
Since that sounds like an issue that can differ in each host/environment.

**Comment by BigRoy** [comment]
> I'd say we moving on for now 

Unfortunately it does look quite off in the Creator too. See:

not_centered

Not as bad as before, but still not super clean.
Also, the disabled state is still a problem:

disabled

Will look into it now.

**Comment by iLLiCiTiT** [comment]
override `QPushButton::menu-indicator:disabled` should work I think

**Comment by BigRoy** [comment]
This PR came to existence to it feeling off in the Creator. So I made sure it looks good there now, which includes a small tweak to the tool to improve styling even more.

Anyway, here is the result

**Creator**
_Enabled_
enabled


_Disabled_
disabled2


**QPushButton big**
_Enabled_
enabled_big

_Disabled_
disabled2_big

Somehow the disabled state icon looks slightly blurry, I'm not sure why that is.

**Comment by davidlatwe** [comment]
Changing to 
```css
QPushButton::menu-indicator  {
    subcontrol-origin: padding;
    subcontrol-position: right;
    width: 8px;
    height: 8px;
    padding-right: 5px;
}

QPushButton::menu-indicator:disabled {
    image: url(:/qss_icons/rc/down_arrow_disabled.png);
    padding-right: 3px;
    width: 9px;
    height: 9px;
}
```
Seems better
creator

BigRoy:
Thanks @davidlatwe that indeed works much better, also much neater when toggling the state because now it doesn't seem to "jump around". 👍  Fixed with https://github.com/getavalon/core/pull/450/commits/9288d0b9266f2086efa92168acd037752e55c792

PR #451, opened 2019-09-18
Fix context manager allowing to switch to 'No Task'

BigRoy:
### Issue

It's possible to have the Context Manager switch the `AVALON_TASK` to "No Task" whereas it should disallow that entry to be active task.

Reported [here](https://gitter.im/getavalon/Lobby?at=5d8246c45ab9361694142ad2)

**What's changed?**

Previously the selected task was retrieved from the view as the `view.currentIndex()` however that also allows non-selectable items to be active and returned. Now this is updated to solely using the actually selected rows from the view to ensure the Task is selected and not just the current.

This way you cannot switch to the Task, because the "No Task" item is unselectable in the model.

### DIFF
--- a/avalon/tools/contextmanager/app.py
+++ b/avalon/tools/contextmanager/app.py
@@ -103,7 +103,16 @@ class App(QtWidgets.QDialog):
         self._context_task.setText("Task: {}".format(task))
 
     def _get_selected_task_name(self):
-        task_index = self._task_view.currentIndex()
+
+        # Make sure we actually get the selected entry as opposed to the
+        # active index. This way we know the task is actually selected and the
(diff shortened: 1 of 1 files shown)


BigRoy:
This could also be adapted to disabling the "set context" button altogether when there's no selected asset or task. Will adapt this PR.

BigRoy:
Added the enable/disabled state toggle for the *Apply* button with https://github.com/getavalon/core/pull/451/commits/2620c2909a4943381e0a3f1c9aa99dc7f96be16b

**Example:**

contextmanager_enable_disable_apply_btn

mottosso:
Great work. Nice and concise. Merge when happy. 👍

PR #452, opened 2019-09-18
Make the "No Task" entry clearly distinct from actual tasks by its icon

BigRoy:
The "No Task" entry icon is the same default as tasks in an asset which makes it harder to distinguish on first sight than it needs to be.

**What's changed?**

The "No Task" icon is changed into `exclamation-circle` and now drawn in a greyish color to stand out from the actual tasks which are by default orange.

_Example:_

afbeelding

### DIFF
--- a/avalon/tools/models.py
+++ b/avalon/tools/models.py
@@ -240,10 +240,11 @@ class TasksModel(TreeModel):
         default_icon = self._icons["__default__"]
 
         if not tasks:
+            no_task_icon = self._icons["__no_task__"]
             item = Item({
                 "name": "No task",
                 "count": 0,
-                "icon": self._icons["__no_task__"],
+                "icon": no_task_icon,
                 "enabled": False,
             })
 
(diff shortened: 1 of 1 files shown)


avalon/tools/models.py
```diff
@@ -241,7 +243,7 @@ def set_assets(self, asset_ids):
             item = Item({
                 "name": "No task",
                 "count": 0,
-                "icon": default_icon,
+                "icon": self._icons["__no_task__"],
                 "enabled": False,
```
davidlatwe: Could we make the `self._icons["__no_task__"]` be unified to `no_task_icon` ?
Like the `default_icon` above, for better reading :relaxed:

BigRoy: Are you saying you want me to solely declare a variable above it? :)
```python
            no_task_icon = self._icons["__no_task__"]
            item = Item({
                "name": "No task",
                "count": 0,
                "icon": no_task_icon,
                "enabled": False,
            })
```

Like that?

davidlatwe: Yes ! And if that make sense to you :relaxed:

BigRoy: Fine with me. Pushed it https://github.com/getavalon/core/pull/452/commits/a13433b3424d1f628af3127ae5a92c6682467849 :)

Issue #454, opened 2019-09-19
Scene Inventory does not distinguish what Loader is used


BigRoy:
### Issue

The **Scene Inventory** does not list what Loader was used to load a specific representation. It solely groups by the exact representation that was loaded, however the exact same published Alembic file could be loaded by a `gpuCacheLoader` *and* a `referenceLoader` for example.

### Proposal

It would be good if the artist could identify which entry was loaded in which way. We could do this in two ways:

1. We include the "Loader" solely as a new column. As such, expanding the hierarchy of a loaded representation shows which loaders were used. This is a trivial light-weight implementation.

2. We include the "Loader" column and additionally also change the grouping to "group by" both representation and loader. That way, without expanding the hierarchies, one can easily identify the Loader used too. This change would be slightly more involved.

I've checked with the artist here who reported it to me and he stated that he'd be fine with 1) where just a column is visible to identify the Loader as one expands it.

davidlatwe:
This should be consider as fixed, right ?

BigRoy:
Sure is. It's implemented as option 1) through #455

PR #455, opened 2019-09-19
Expose the "loader" column in Scene Inventory

BigRoy:
This implements a fix for #454 so that it's visible to the artist which Loader is being used for specific loaded content directly within the Scene Inventory.

This implements solution option 1) as presented in the issue. It's a very minor change in code. :)

**BEFORE**

before

**AFTER**

after

### DIFF
--- a/avalon/tools/sceneinventory/model.py
+++ b/avalon/tools/sceneinventory/model.py
@@ -15,7 +15,7 @@ from . import lib
 class InventoryModel(TreeModel):
     """The model for the inventory"""
 
-    Columns = ["Name", "version", "count", "family", "objectName"]
+    Columns = ["Name", "version", "count", "family", "loader", "objectName"]
 
     OUTDATED_COLOR = QtGui.QColor(235, 30, 30)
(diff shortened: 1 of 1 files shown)


mottosso:
Oo, that is nice. 👍

davidlatwe:
What about putting the loader name inside the family column with margin ? Looks like they could be merged together.

BigRoy:
> What about putting the loader name inside the family column with margin ? Looks like they could be merged together.

It would complicate the code more than it currently does. Additionally it disallows sorting either on Family or on Loader type. I think it offers less features with more complexity and I personally don't have an issue with it being a column of its own. With that it even allows to hide it if you wouldn't care about it and don't want to show it.

As such, I personally wouldn't merge them.

davidlatwe:
> It would complicate the code more than it currently does.

Yeah, good point, just found out that this PR only modified one line. Now I could not bare to complicate it for that, too.

BigRoy:
> just found out that this PR only modified one line. 

It was definitely *more* trivial than I expected. 😄

mkolar:
great!

PR #458, opened 2019-10-04
Fix #457: Remove schema validation when parsing containers

### LINKED ISSUE #457: Optimisation: Improve host ls() speed by removing schema validation
### Issue

In #456 the performance checks brought forward that on very large scenes with many containers the **schema validation of the container** can [degrade overall performance to an unwanted level](https://github.com/getavalon/core/pull/456#issuecomment-536956313) for `host.ls()`. [Another performance measurement can be found at the bottom of this comment](https://github.com/getavalon/core/pull/456#issuecomment-536940356).

The comparisons:

```
Scene with roughly 24 containers:
- 0.309000015259 seconds without validation
- 0.477999925613 seconds with validation

Scene with 1279 container (by @davidlatwe) <- this is solely `parse_container()`
- 2.621 seconds with validation
- 0.155 seconds without validation
```

Note how for *many* containers the speed difference is very noticeable.

_This was tested with host Maya, but it's the same `schema.validate` check in other hosts._

### Discussion

The `schema.validate(container)` is a check to ensure the container behaves as expected for a container as it is generated. Without it nothing ensures that whatever the function returns actually behaves exactly like a container. 

**Should we remove the validation altogether?**

One could argue that Python tests should ensure the API is stable to return correct containers (e.g. tests that run on example scenes with the validation) as opposed to *always* needing to run this on every container on `host.ls()` in production. That way we can still ensure the API always ends up working as expected. This would then result in the same expected safety yet without the speed degradation in production.

---

_I'm opening this issue because of some comments were made in #456 about whether to allow [toggling the validation through an argument to `host.ls()`](https://github.com/getavalon/core/pull/456#issuecomment-536993209) or whether to [remove it alltogether so that UIs would also allow the speed improvement](https://github.com/getavalon/core/pull/456#issuecomment-537001689)._

BigRoy:
This implements #457 by removing the `schema.validate(container)` for `parse_container` which is used by most `ls()` methods of the hosts (only `fusion` didn't have it).

### DIFF
--- a/avalon/houdini/pipeline.py
+++ b/avalon/houdini/pipeline.py
@@ -201,12 +201,11 @@ def containerise(name,
     return container
 
 
-def parse_container(container, validate=True):
+def parse_container(container):
     """Return the container node's full container data.
 
     Args:
         container (hou.Node): A container node name.
-        validate(bool): turn the validation for the container on or off
 
     Returns:
         dict: The container schema data for this container node.
@@ -220,9 +219,6 @@ def parse_container(container, validate=True):
     # Append transient data
     data["objectName"] = container.path()
     data["node"] = container
-
-    if validate:
-        schema.validate(data)
 
     return data
 
--- a/avalon/maya/pipeline.py
+++ b/avalon/maya/pipeline.py
@@ -423,13 +423,11 @@ def containerise(name,
     return container
 
 
-def parse_container(container, validate=True):
+def parse_container(container):
     """Return the container node's full container data.
 
     Args:
         container (str): A container node name.
-        validate (bool, optional): Whether to validate the container schema.
-            Defaults to True.
 
     Returns:
         dict: The container schema data for this container node.
@@ -442,9 +440,6 @@ def parse_container(container, validate=True):
 
     # Append transient data
     data["objectName"] = container
-
-    if validate:
-        schema.validate(data)
 
     return data
 
--- a/avalon/nuke/pipeline.py
+++ b/avalon/nuke/pipeline.py
@@ -99,7 +99,7 @@ def containerise(node,
     return node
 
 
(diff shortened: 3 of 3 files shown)


avalon/houdini/pipeline.py
```diff
@@ -201,12 +201,11 @@ def containerise(name,
     return container
 
 
-def parse_container(container, validate=True):
+def parse_container(container):
     """Return the container node's full container data.
 
     Args:
         container (hou.Node): A container node name.
-        validate(bool): turn the validation for the container on or off
```
davidlatwe: Should we keep that argument `validate` and just adding deprecation note in doc string for not breaking old stuff for now (at least before version `6.0.0`) ?

Although `parse_container` is not in host API, but I guess some people might using it in their custom code since it's kind of API snubbed useful function in my opinion.

Having that said, I didn't used that argument so it's okay to just remove it on my end.

BigRoy: The `parse_container` is not exposed in the API, however it is somehow exposed in the host API for Nuke where it shouldn't be actually. Which now complicates things... 

I'd refrain from the backwards compatibility for something that should not be a part of the API. And maybe even *asap* remove it from `__init__.py` for the Nuke host. I think the only ones using Nuke currently are @tokejepsen and @mkolar (and some of their clients). They also initialized the Nuke host setup so probably can tell us whether it should or should not be exposed for Nuke (and whether they ended up using it - if not, I'd vote for removing it there too.)

*If* in the end we decide it's not part of the API then I would skip the backwards compatibility.

tokejepsen: Personally I say get rid of it.

BigRoy:
Can we merge this PR? :)

PR #459, opened 2019-10-10
creationFlags are not supported on linux/osx

antirotor:
Avalon is using `creationFlags` to launch subprocess via its `launch()` function. `CreationFlags` are supported only on Windows in Python. It is therefor failing on linux and macos.

**What's changed?**
I've just added condition to use it only on Windows.

### DIFF
--- a/avalon/lib.py
+++ b/avalon/lib.py
@@ -219,13 +219,11 @@ def launch(executable, args=None, environment=None, cwd=None):
 
     # this won't do anything on linux/macos as `creationFlags` are
     # only windows specific.
-    if env.get("CREATE_NEW_CONSOLE"):
-        if IS_WIN32:
-            kwargs["creationflags"] = CREATE_NEW_CONSOLE
+    if IS_WIN32 and env.get("CREATE_NEW_CONSOLE"):
+        kwargs["creationflags"] = CREATE_NEW_CONSOLE
         kwargs.pop("stdout")
         kwargs.pop("stderr")
     else:
-
         if IS_WIN32:
             kwargs["creationflags"] = CREATE_NO_WINDOW
 
--- a/avalon/lib.py
+++ b/avalon/lib.py
@@ -218,7 +218,8 @@ def launch(executable, args=None, environment=None, cwd=None):
     )
 
     if env.get("CREATE_NEW_CONSOLE"):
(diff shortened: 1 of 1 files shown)


BigRoy:
Looking good. Does it behave as expected in that condition for Linux/Mac. For example, as `CREATE_NEW_CONSOLE` is defined does it launch in a separate console/process (e.g. detaching itself).

antirotor:
well no, there is no easy way to do it. If you are running from terminal, there is really no concept of *detaching*. Running under desktop environment, there you would need to specify what terminal emulator you are using, what shell you are using and then spawn it and run subprocess. This is by my opinion too much hustle to do. Now it doesn't detach iself but I don't think it affect usability very much - I've ran into it when trying to launch Python action from Launcher. Now the Python interpreter will start in the same terminal we used to start avalon.

BigRoy:
Ah ok, the only caveat then is that the `CREATE_NEW_CONSOLE` environment variable might be misleading on Mac OS / Linux as it wouldn't actually do so. Not sure if a problem, we just need to make sure anyone using it is aware of it.

Should we maybe mention it in a comment in the code too?

antirotor:
True, I've added comment there explaining behaviour. We could probably later add such functionality for linux, but it would require lots of code I don't think is worth to invest there. Those flags were until now show-stopper on linux so this little change will stop crashes there albeit without full functionality as on Windows.

antirotor:
This is not problem with `CREATE_NEW_CONSOLE` but with setting *creationgflags* at all, they are not recongized at all on Linux and lead to crash and traceback stating so.

avalon/lib.py
```diff
@@ -217,8 +217,11 @@ def launch(executable, args=None, environment=None, cwd=None):
         universal_newlines=True,
     )
 
+    # this won't do anything on linux/macos as `creationFlags` are
+    # only windows specific.
     if env.get("CREATE_NEW_CONSOLE"):
```
mottosso: To improve readability of this, could we make it:

```py
if IS_WIN32 and env.get("CREATE_NEW_CONSOLE"):
  ...
```

mottosso:
Looks good, thanks!

davidlatwe:
Add another note for future referencing :)
https://stackoverflow.com/a/25124003/4145300

PR #462, opened 2019-10-28
Fix #461: use `getfullargspec` for Python 3

### LINKED ISSUE #461: `inspect.getargspec` is deprecated since Python 3.0
As stated in the [docs](https://docs.python.org/3/library/inspect.html#inspect.getargspec) the `inspect.getargspec` function is deprecated since Python 3.0 and `inspect-getfullargspec` is a drop-in replacement in most cases.

In my Blender integration I use function annotations and also have keyword-only parameters, which is handled correctly by `getfullargspec` (but not `getargspec`).

~I will also make a pull request with a simple fix.~
Pull request is [here](https://github.com/getavalon/core/pull/462).

jasperges:
This resolves #461 

**What's changed?**
Use `inspect.getfullargspec` for Python 3. Also see [issue #461](../issues/461).

### DIFF
--- a/avalon/pipeline.py
+++ b/avalon/pipeline.py
@@ -708,7 +708,10 @@ def _validate_signature(module, signatures):
 
         else:
             attr = getattr(module, member)
-            signature = inspect.getargspec(attr)[0]
+            if sys.version_info.major >= 3:
+                signature = inspect.getfullargspec(attr)[0]
+            else:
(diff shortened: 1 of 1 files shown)


jasperges:
Just to be sure: yes, I *did* test it :). Because I work mainly with Blender, everything I do is tested in Python 3.7.

mottosso:
Looks good, short and sweet!

PR #464, opened 2019-10-28
Resolve #463: RuntimeError doesn't have message attribute

### LINKED ISSUE #463: RuntimeError doesn't have `message` attribute in Python 3
In the `loader` the message of an exception is checked to see if it contains 'already deleted' (see [here](../blob/master/avalon/tools/loader/app.py#L407)).
To do this the `message` atrribute is checked. In Python 3 this doesn't exist.
A simple, straightforward fix would be to use `str(e)` instead of `e.message`. This works in both Python 2 and 3.

~Pull request follows.~
Pull request: #464

jasperges:
**What's changed?**
Very simple and basic fix for #463 . This makes it also work in Python 3 which doesn't have the `message` attribute on a `RuntimeError`.

### DIFF
--- a/avalon/tools/loader/app.py
+++ b/avalon/tools/loader/app.py
@@ -404,7 +404,7 @@ def show(debug=False, parent=None, use_context=False):
             module.window.refresh()
             return
         except RuntimeError as e:
-            if not e.message.rstrip().endswith("already deleted."):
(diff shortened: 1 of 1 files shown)


mottosso:
Nice one, thanks!

PR #465, opened 2019-10-30
Make the workfiles app work in Blender

jasperges:
For this I didn't make an issue, because it seems easier to discuss the changes when looking at the code itself. Questions, concerns or suggestions are more then welcome!

There are basically 2 things needed to make it work:

1. Keep a reference to the Qt window (and children) to avoid garbage collection. To keep track of the window `module.window` is now used. This also has the advantage of making it consistent with the other tools which do this already.
2. Avoid running `exec_()` for the main window, because this blocks Blender and creates conflicts between the Qt event loop and Blender event loop (if I'm not mistaken). Instead just `show()` the window and keep it responsive by triggering `processEvents()` all the time from within Blender. This last part is not shown here, because it's part of my Blender integration. This also makes it consistent with the other tools which also just run `show()`. For the `messagebox` and `name_window` it seems fine to run `exec_()`, they are not working if you run `show()`. To be honest I'm not completely sure why.

I hope this doesn't cause any problems with other DCC's, but I suspect it's fine because the other tools also do this.

### DIFF
--- a/avalon/tools/workfiles/app.py
+++ b/avalon/tools/workfiles/app.py
@@ -9,6 +9,8 @@ from ... import style, io, api
 
 from .. import lib as tools_lib
 
+module = sys.modules[__name__]
+module.window = None
 
 class NameWindow(QtWidgets.QDialog):
     """Name Window"""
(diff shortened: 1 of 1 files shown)


tokejepsen:
Nice one @jasperges !

> because it's part of my Blender integration

What is your Blender integration? Is Avalon missing something, or are you talking about your Avalon config?

jasperges:
> What is your Blender integration? Is Avalon missing something, or are you talking about your Avalon config?

I'm adding Blender 2.80 as a [supported host to Avalon](https://github.com/jasperges/avalon-core/tree/blender28).

tokejepsen:
Ahh, I see. This PR is part of a future bigger PR.

jasperges:
Yes, this is in preperation of that one. But I felt a seperate PR is in place here, because apart from giving problems with Blender it also is not consistent with the other tools. But I also wouldn't mind to incorporate it into the the 'blender PR'.

mottosso:
Nice and consistent, thanks!

BigRoy:
@jasperges Could this have been fixed by explicitly setting the parent on those widgets to `self` so it automatically keeps a reference to it? I'm working on the Work Files tool and merging it with Context Manager. I'll try and do the 'parent' widget correctly and see if that helps. Then you can see if it crashes on you, if so we can always revert.

jasperges:
@BigRoy I have no idea... Let's find out! :nerd_face:

PR #473, opened 2019-11-08
Re-implement Nuke `imprint`

davidlatwe:
This PR is the first step of refactoring #438, for resolving `imprint` which functionality was a bit rigid.

### What's changed

* Re-implemented `lib.imprint`
* Changed following functions to adopt new `lib.imprint`:
    - `lib.set_avalon_knob_data`
    - `lib.add_publish_knob`
    - `pipeline.Creator`
* Add back `lib.get_avalon_knob_data` from #438

### About `lib.imprint`

Nuke has [three categories of knobs](https://learn.foundry.com/nuke/developers/90/ndkdevguide/knobs-and-handles/intro.html#knob-classifications), *data-stored knob*, *layout knob* and *complex data knob*, but let's put *complex data knob* aside for now.

On *data-stored knob*, `imprint` will create knob which fits to regular data types, like `bool`, `int`, `float`, `str`, `list`. For `dict`, it will become a group of knobs, or a group of tabs.

For example:

```python
# Creating two group of knobs
node = nuke.createNode("NoOp")
data = {
    "group1": {
        "group1.var1": 5,
        "group1.var2": "hello",
        "group1.var3": ["a", "b"],
    },
    "group2": {
        "group2.var1": 6,
        "group2.var2": "world",
        "group2.var3": ["o", "p"],
    },
}
lib.imprint(node, data, tab="User")
```
Which you will get

image

And this:

```python
data = {
    "tabGroup": {
        "tab1": {"count": 5},
        "tab2": {"isGood": True},
        "tab3": {"direction": ["Left", "Right"]},
    },
}
node = nuke.createNode("NoOp")
lib.imprint(node, data, tab="User")
```

Will looks like this

image

And if you wish to use other kind of knobs which is not mapped in `lib.create_knob`, you could use `lib.Knobby`, for example:

```python
data = {
    # Using tuple as key for manual nice naming
    ("my_knob", "Nice Knob"): lib.Knobby("Text_Knob", "Non-editable text"),
    ("divider", ""): lib.Knobby("Text_Knob", ""),
    "MyFilePath": lib.Knobby("File_Knob", "/file/path"),
}
node = nuke.createNode("NoOp")
lib.imprint(node, data, tab="User")
```

And you have a file path knob

image

---

Hope this PR make sense to you, please let me know what you think. :relaxed:

### DIFF
--- a/avalon/nuke/lib.py
+++ b/avalon/nuke/lib.py
@@ -244,7 +244,7 @@ def set_avalon_knob_data(node, data=None, prefix="avalon:"):
     body[("divd", "")] = Knobby("Text_Knob", "")
 
     for key, value in data.items():
-        name = prefix + key
+        name = (prefix + key, key)  # Hide prefix on GUI
         if key in editable:
             body[name] = value
         else:
--- a/avalon/nuke/lib.py
+++ b/avalon/nuke/lib.py
@@ -193,11 +193,20 @@ def create_knobs(data, tab=None):
 
 
 def add_publish_knob(node):
+    """Add Publish knob to node
+
+    Arguments:
+        node (obj): nuke node to be processed
+
+    Returns:
+        node (obj): processed nuke node
+
+    """
     if "publish" not in node.knobs():
-        knob = nuke.Boolean_Knob("publish", "Publish")
-        knob.setFlag(0x1000)
-        knob.setValue(False)
-        node.addKnob(knob)
+        body = OrderedDict()
+        body[("divd", "")] = Knobby("Text_Knob", "")
+        body["publish"] = True
+        imprint(node, body)
     return node
 
 
--- a/avalon/nuke/lib.py
+++ b/avalon/nuke/lib.py
@@ -153,6 +153,7 @@ def create_knobs(data, tab=None):
         elif isinstance(value, bool):
             knob = nuke.Boolean_Knob(name, nice)
             knob.setValue(value)
+            knob.setFlag(nuke.STARTLINE)
 
         elif isinstance(value, int):
             knob = nuke.Int_Knob(name, nice)
--- a/avalon/nuke/lib.py
+++ b/avalon/nuke/lib.py
@@ -180,10 +180,14 @@ def create_knobs(data, tab=None):
         elif isinstance(value, dict):
             if all(isinstance(v, dict) for v in value.values()):
                 # Create a group of tabs
-                knobs.append(nuke.BeginTabGroup_Knob())
+                begain = nuke.BeginTabGroup_Knob()
+                end = nuke.EndTabGroup_Knob()
+                begain.setName(name)
+                end.setName(name + "_End")
+                knobs.append(begain)
                 for k, v in value.items():
                     knobs += create_knobs(v, tab=k)
-                knobs.append(nuke.EndTabGroup_Knob())
+                knobs.append(end)
             else:
                 # Create a group of knobs
                 knobs.append(nuke.Tab_Knob(name, nice, nuke.TABBEGINGROUP))
--- a/avalon/nuke/lib.py
+++ b/avalon/nuke/lib.py
@@ -3,6 +3,7 @@ import contextlib
 import nuke
 import re
 import logging
+from collections import OrderedDict
 
 from ..vendor import six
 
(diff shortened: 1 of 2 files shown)


avalon/nuke/lib.py
```diff
@@ -23,48 +24,255 @@ def maintained_selection():
             [n['selected'].setValue(True) for n in previous_selection]
 
 
-def add_avalon_tab_knob(node):
-    """Adding a tab and a knob into a node
+def imprint(node, data, tab=None):
+    """Store attributes with value on node
+
+    Parse user data into Node konbs.
```
BigRoy: Typo `konbs` -> `knobs`

avalon/nuke/pipeline.py
```diff
@@ -155,10 +155,13 @@ def process(self):
         if (self.options or {}).get("useSelection"):
             nodes = nuke.selectedNodes()
 
-        instance = [n for n in nodes
-                    if n["name"].value() in self.name] or None
+        nodes = [n for n in nodes
```
BigRoy: Could this be turned into

```python
node = next((n for n in nodes if n["name"].value() in self.name), None)`
```

Also, the fallback to `nuke.allNodes()` at the top, would you _ever_ want that to accidentally happen? Seems like that could cause crazy things at some point. You might, I'm just not sure what the intended workflow is.

And, line 161 will cause error on nothing selected `None[0]`. Maybe better to report a decent error message like "Select X or Y"?

davidlatwe: I think this has been taken cared by some of the commits in #483, a bit hard to tell which's which, but from the [full changes](https://github.com/getavalon/core/pull/438/files#diff-c5bffa46ac90c806a167aa2ca1b96509R167-R194) could see that's been improved.

Should be shipped in next PR :P

avalon/nuke/lib.py
davidlatwe: Fixed in b4a556c :wrench:

mkolar:
I like it. Certainly cleaner

davidlatwe:
### New minor changes

First two commits (7079c14, 9229fec) were the fixes to complete this PR. The rest were minor improvements.

It's good to be merged now, but I'll be out for travel starting tomorrow till next Thursday so if no other objections, merging this next weekend !

davidlatwe:
Sorry, just spotted that updating container will create duplicated knobs if existed, commit dc82024 fixed that. And one more change is if given an invalid container node, `TypeError` will raised. (7715ebd)

That's it !!

avalon/nuke/lib.py
```diff
@@ -23,48 +24,255 @@ def maintained_selection():
...
+            body[name] = value
+        else:
+            body[name] = Knobby("Text_Knob", str(value))
+
+    imprint(node, group, tab="AvalonTab")
+
     return node
 
 
+def get_avalon_knob_data(node, prefix="avalon:"):
```
jezscha: Would it be possible to turn this into `prefix=["avalon:",  "ak:"]` for backward compatibility with already running productions? it was previously `prefix="ak:"`

avalon/nuke/pipeline.py
jezscha: yes this has been fixed additionally in #438

jezscha:
@davidlatwe this PR is awesome! Thanks for help ;)

avalon/nuke/lib.py
davidlatwe: Ah, okay, but I won't put mutable value as arg's default value, I might hard code it in the for loop below :)

Speaking of which, was there any reason for applying prefix on knobs ?
Would it be okay if we not using prefix in future (with backward compatibility of course) ?

jezscha: > Would it be okay if we not using prefix in future (with backward compatibility of course) ?

If you found the efficient way to get efficient way of filtering only avalon attributes from all knobs then yes. The reason was to separate only avalon metadata attributes, but didm't see other way to do it ;)

davidlatwe: I have found a way to list out all user-defined knobs in any given node, just like other host implementations (`lib.read`). But noted that user-defined layout knobs like tab or group will be listed too.

Here's the gist for trying:
https://gist.github.com/davidlatwe/519528bb01b4ff816389464ec2709e97

Maybe we could work without prefixing knob name if this works ?
Anyway, I'll try resolving this in another PR, this PR will be merged as is for now. :relaxed:

tokejepsen: Interesting solution!
Listing user defined knobs would definitely be useful.

davidlatwe: @jezscha  I end up implementing the backward compatibility in a new method `lib.read`, instead of changing `lib.get_avalon_knob_data`. Please let me know how you think of this, it's in PR #476 :smiley:

PR #475, opened 2019-11-12
change fps mapping in maya

antirotor:
**What's changed?**
Fps in Maya is set by mapping actual fps number to string constants used by Maya to set time units.  You can have fps set as integer - 25 or as float 25.0 They are numerically same but mapping recognize only integer. If you pull fps from system like ftrack, where this attribute is set as decimal to allow values like 23.976 it will also returns 25 as 25.0. To handle it, this change cast numbers like 25.0 to integer so mapping still works and adds few other Maya string constants.

### DIFF
--- a/avalon/maya/commands.py
+++ b/avalon/maya/commands.py
@@ -8,6 +8,7 @@ For interactive use, see :mod:`interactive.py`
 """
 
 from maya import cmds
+import math
 
 from .. import io, api
 
--- a/avalon/maya/commands.py
+++ b/avalon/maya/commands.py
@@ -8,7 +8,6 @@ For interactive use, see :mod:`interactive.py`
 """
 
 from maya import cmds
-import math
 
 from .. import io, api
 
--- a/avalon/maya/commands.py
+++ b/avalon/maya/commands.py
@@ -34,29 +34,22 @@ def reset_frame_range():
         cmds.warning("No edit information found for %s" % shot["name"])
         return
 
-    # this handles fps set as decimal number. For example
-    # fps 25.0 is corrected to int so mapping will still work
-    avalon_fps = api.Session.get("AVALON_FPS")
-    d, i = math.modf(float(api.Session.get("AVALON_FPS", 25)))
-    if d == 0.0:
-        avalon_fps = int(i)
-
-    fps = {'15': 'game',
-           '24': 'film',
-           '25': 'pal',
-           '30': 'ntsc',
-           '48': 'show',
-           '50': 'palf',
-           '60': 'ntscf',
-           '23.98': '23.976fps',
-           '23.976': '23.976fps',
-           '29.97': '29.97fps',
-           '47.952': '47.952fps',
-           '47.95': '47.952fps',
-           '59.94': '59.94fps',
-           '44100': '44100fps',
-           '48000': '48000fps'
-           }.get(str(avalon_fps), "pal")  # Default to "pal"
+    fps = {15: 'game',
+           24: 'film',
+           25: 'pal',
+           30: 'ntsc',
+           48: 'show',
+           50: 'palf',
+           60: 'ntscf',
+           23.98: '23.976fps',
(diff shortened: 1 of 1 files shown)


avalon/maya/commands.py
```diff
@@ -33,17 +34,29 @@ def reset_frame_range():
...
+           '25': 'pal',
+           '30': 'ntsc',
+           '48': 'show',
+           '50': 'palf',
+           '60': 'ntscf',
+           '23.98': '23.976fps',
+           '23.976': '23.976fps',
+           '29.97': '29.97fps',
+           '47.952': '47.952fps',
+           '47.95': '47.952fps',
```
BigRoy: Just wondering, should both of these values result in the same final FPS? Or is one intended to be the `df` (Drop Frame) variant? If the latter then it seems the resulting value is wrong?

avalon/maya/commands.py
```diff
@@ -33,17 +33,29 @@ def reset_frame_range():
...
-        "25": "pal",
-        "30": "ntsc",
-        "48": "show",
-        "50": "palf",
-        "60": "ntscf"
-    }.get(api.Session.get("AVALON_FPS"), "pal")  # Default to "pal"
+    # this handles fps set as decimal number. For example
+    # fps 25.0 is corrected to int so mapping will still work
+    avalon_fps = api.Session.get("AVALON_FPS")
+    d, i = math.modf(float(api.Session.get("AVALON_FPS", 25)))
```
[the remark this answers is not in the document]
BigRoy: Resolved with https://github.com/getavalon/core/pull/475/commits/52574eb4eddf0d1cccc2f898a2709a28c7b87ce7

avalon/maya/commands.py
antirotor: Well those are not drop frames, just way to handle rounding. I've sometime noticed that different productions specify framerates with different precision. This handle two and three decimal places as fewer or more doesn't make much sense. This could be handled in more elegant way I suppose but way one can see whats mapped to what.

avalon/maya/commands.py
```diff
@@ -33,17 +34,29 @@ def reset_frame_range():
...
+           '60': 'ntscf',
+           '23.98': '23.976fps',
+           '23.976': '23.976fps',
+           '29.97': '29.97fps',
+           '47.952': '47.952fps',
+           '47.95': '47.952fps',
+           '59.94': '59.94fps',
+           '44100': '44100fps',
+           '48000': '48000fps'
+           }.get(str(avalon_fps), "pal")  # Default to "pal"
```
davidlatwe: I think we could just use float or int as key instead of strings, that way, you won't need `math.modf` to ensure the string representation of FPS value.

For example:

```python
FPS_MAP = {
    24: "film",
}
assert FPS_MAP[24] == "film"
assert FPS_MAP[24.0] == "film"
```

So we could just convert the value of "AVALON_FPS" into float and directly map to the Maya FPS.

BigRoy: The danger would I guess be a potential floating point precision problem? But if that doesn't happen then maybe yes, this could be simplified. @antirotor could you try David's approach out for your cases and see if it results in issues?

antirotor: This is good approach, more clean I would say. But it would require changes in config or other places where when using AVALON_FPS we would need to convert from str to float, as environment variable should have its value in str anyway. So I think we can either go with cleaner solution breaking dependent code slightly or hack it like I did but probably without need to change anything else.

davidlatwe: > But it would require changes in config or other places where when using AVALON_FPS we would need to convert from str to float, as environment variable should have its value in str anyway.

I think neither of us would need to change anything in any config since the *str-to-float* conversion already happened [here] in this PR, right ? :thinking:

antirotor: You are absolutely right. Fixed it.

PR #479, opened 2019-12-02
Fix loader in Houdini 18 as it returns 'long' instead of 'int'

BigRoy:
**What's changed?**

It seems somehow the returned version number is not an `int` in Houdini 18 but a `long`. As such, the assertion in the `VersionDelegate` would catch it incorrectly. 

This is now resolved by checking it against `numbers.Integral` to ensure it's also Python 3+ proof where `long` no longer exists.

### DIFF
--- a/avalon/tools/delegates.py
+++ b/avalon/tools/delegates.py
@@ -1,3 +1,5 @@
+import numbers
+
 from ..vendor.Qt import QtWidgets, QtCore
 from .. import io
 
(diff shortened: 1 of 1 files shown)


iLLiCiTiT:
Hi, we solved this issue few day ago with try except under import part. Like:
```
try:
    long
except NameError:
    long = in
```
and in displayText we add this check:
```
isinstance(value, long):
    value = int(value)
```

BigRoy:
Seems like a similar fix. I'll merge this in a few days if there are no objections.

PR #471, opened 2019-11-05
Fix Avalon-Nuke implementation

davidlatwe:
### Problem

Nuke launched from Avalon will instantly crashed due to some missing function and module-not-found error. 

### What's changed

* Cleanup a bit so I can launch Nuke without any error.
    - Import `toml` from `avalon.vendor` (c78255b)
    - Remove `ls_img_sequence` (fb4cde5)
* Instead of hard-coded to use `pyblish-lite`, revert back to use `avalon.tools.publish` so one can register any pyblish GUI on demand via using `pyblish.register_gui` in config module. (Also, did not encounter pyblish/pyblish-qml#301, even so, no reason to hard code for fix :relieved: ) (b668318)

PR #484, opened 2019-12-13
Porting#438

davidlatwe:
This PR is aim to resolve #438.

PR #438 was huge, and a few parts of it was build on top of old tech-debt. But after PR #471, #473, #476, I think this is the final piece to finish up the rest.

### What's changed

* `nuke.ls` now listing containers from all types of nodes, not just `Read` type node.
* Changed `nuke.Creator` to allow instance to have multiple nodes with backdrop.
* Added `nuke.command` and move util functions from `nuke.pipeline` to there. (Backwards compatible)
* Added a few helper functions to `nuke.lib`

Please let me know if there's anything missing or broken. @jezscha

### DIFF
--- a/avalon/nuke/lib.py
+++ b/avalon/nuke/lib.py
@@ -387,13 +387,9 @@ def fix_data_for_node_create(data):
     """Fixing data to be used for nuke knobs
     """
     for k, v in data.items():
-        try:
-            if isinstance(v, unicode):
-                data[k] = str(v)
-        except Exception:
+        if isinstance(v, six.text_type):
             data[k] = str(v)
-
-        if "0x" in str(v):
+        if str(v).startswith("0x"):
             data[k] = int(v, 16)
     return data
 
--- a/avalon/nuke/lib.py
+++ b/avalon/nuke/lib.py
@@ -5,7 +5,7 @@ import re
 import logging
 from collections import OrderedDict
 
-from ..vendor import six
+from ..vendor import six, clique
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


avalon/nuke/lib.py
```diff
@@ -24,6 +32,25 @@ def maintained_selection():
             [n['selected'].setValue(True) for n in previous_selection]
 
 
+def reset_selection():
```
tokejepsen: Maybe `deselect_all` is a better name?

avalon/nuke/lib.py
```diff
@@ -24,6 +32,25 @@ def maintained_selection():
...
+    """Deselect all selected nodes
+    """
+    for node in nuke.selectedNodes():
+        node["selected"] = False
+
+
+def select_nodes(nodes):
```
tokejepsen: Maybe `select_input_nodes` is more descriptive?

davidlatwe: I think adding `input` will give an impression that was used to select input (upstream) nodes of given nodes.

avalon/nuke/lib.py
davidlatwe: Yeah, @jezscha  what do you think ?

avalon/nuke/lib.py
tokejepsen: Ahh, I only read the doc string. Name is good then.

avalon/nuke/lib.py
davidlatwe: Hmm, no wait. Since this isn't a new function that introduced here and already been used, so we better not to change the name. :thinking:

jezscha: Hi @davidlatwe sorry for the slow response.. Yes please this is used many times.. anyway you could rename it but link this for backward compatibility

davidlatwe: I can live with current name :)

davidlatwe:
Merging this :)

PR #488, opened 2019-12-14
Fix #487: Preserve loader styling in Maya 2020

### LINKED ISSUE #487: Maya 2020 - Avalon Loader stylesheet is wrong
**What happened?**

The Loader in Maya 2020 seems to have a very different styling, as if Maya now overrides it. I'm not noticing similar differences in other Maya tools, only on the loader.

afbeelding

**What did you expect?**

This should resemble more the original styling as when run Standalone or like in Maya 2019:

afbeelding

**To Reproduce**

Open the Loader in Maya 2020.

---

Reference:

- [As mentioned on Avalon Gitter](https://gitter.im/getavalon/Lobby?at=5df35df9ac14cc652c76c537)
- Maya 2020 updates to Qt for Python 5.12.5 from PySide2 2.0.0~alpha0.

BigRoy:
**What's changed?**

This fixes #487 by avoiding the need to set a stylesheet to the `QSplitter` and instead have it be applied in the global avalon stylesheet.

I felt this was the best approach from the proposed solutions as it required the least amount of code, plus allows anyone to still override any styling themselves if they'd want.

Note: to test this with Maya 2020 you'll need #486 if you want to have Maya set up the Avalon menu correctly. You could still test it without that code, but you'd need to show the Avalon Loader through code manually.

### DIFF
--- a/avalon/style/style.qss
+++ b/avalon/style/style.qss
@@ -1220,3 +1220,7 @@ QSplitter::handle:horizontal:hover {
     background-color: #787876;
     border: 1px solid #3A3939;
 }
+
+QSplitter {
+    border: 0px;
+}
--- a/avalon/tools/loader/app.py
+++ b/avalon/tools/loader/app.py
@@ -58,9 +58,6 @@ class Window(QtWidgets.QDialog):
         split.addWidget(version)
         split.setSizes([180, 950, 200])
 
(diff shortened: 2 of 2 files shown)


davidlatwe:
Can confirm this fixed the issue.

Issue #499, opened 2019-12-29
Avoid requiring host in config to avoid "No module named xxx" errors


BigRoy:
### Issue

Say you're a newcomer to Avalon and you'd want to try the setup in Nuke with a config that does not have `{config}/nuke/__init__.py` yet you will [run into an `ImportError` error saying "No module name nuke"](https://github.com/getavalon/core/blob/d380d88861764a1a53e49de5a7066f1f6de313e5/avalon/nuke/pipeline.py#L243).

The same is true for other hosts like:
- [Maya](https://github.com/getavalon/core/blob/master/avalon/maya/pipeline.py#L91) same error.
- [Houdini](https://github.com/getavalon/core/blob/d380d88861764a1a53e49de5a7066f1f6de313e5/avalon/houdini/pipeline.py#L71) same error.
-  [Fusion](https://github.com/getavalon/core/blob/d380d88861764a1a53e49de5a7066f1f6de313e5/avalon/fusion/pipeline.py#L46) which looking at the code will actually pass but instead return the config itself as opposed to the `fusion` package inside of it. Which is also very wrong. It should actually set `config = None`.

#### Clarify what is going on, and allow it to pass.

It should either:

1. Show a descriptive error message on how this could be solved. 
2. Or, maybe even more preferred, allow it to actually run fine without _requiring_ `{config}/nuke` module to exist. Maybe just logging a warning of something like: 
      - _Config "polly" has no dedicated "nuke" module. Skipping config initialization for "nuke"_

I'd opt for two and make it very clear where and how one could set something like that up. Maybe even pointing to a part in the documentation on how to set up custom installations in a studio config for an integration.

#### Avoid code duplication

Additionally, note how the `find_host_config` code is duplicated among all integrations. I'd recommend instead making a single function in `avalon.lib` like:
```python
# pseudocode
def find_host_config(config, host):

    config_name = config.__name__
    module_name = "%s.%s" % (config_name, host)
    try:
        config = importlib.import_module(module_name)
    except ImportError as exc:
        if str(exc) != "No module name {}".format(module_name):
            log.warning("Config has no '%s' module." % module_name)
        config = None

    return config
```

And then e.g. use it in the [`avalon.nuke.pipeline`](https://github.com/getavalon/core/blob/d380d88861764a1a53e49de5a7066f1f6de313e5/avalon/nuke/pipeline.py#L231) as simply as `lib.find_host_config(config, "nuke")`.

### To Reproduce

[...]

PR #501, opened 2019-12-31
Patch #499: Refactor find_host_config to lib.find_module_in_config

BigRoy:
This PR is related to #499 

This now makes sure the way the `config.{host}` module is retrieved per host matches across each integration using `avalon.lib.find_module_in_config` as a replacement for the `find_host_config` function that each `avalon.{host}` module implemented.

Aside of that, I've made the variable names consistently `config_host` as opposed to replacing the `config` variable that could previously had left some confusion thinking it was the config module, but was actually the `config.{host}` mdoule.

And of course, as per #499 this should now allow `config.{host}` module to not exist at all for each of the integrations with only a logged warning along the lines of: `Config has no '{config}.{host}' module.`

Additionally this makes these changes:

- Add `uninstall()` function to `avalon.fusion` as that was missing.
- Match Houdini and Fusion metadata collecting to Maya implementation, which had an optimization.

### Discussion

**NOTE:** This code should still be tested. However I wanted to set up this PR to discuss the code choices.

1. Is the code clear enough?

2. This imports the `find_module_in_host` function from `avalon.lib` as it became more readable than doing the following. However I believe this was frowned upon by the code style, but I couldn't find that mentioned anywhere.
```python
# in this PR
from ..lib import find_module_in_host
find_module_in_host(config, "maya")

# the other option
import ..lib as avalon_lib
avalon_lib.find_module_in_host(config, "maya")
```
    
3. Can we simplify further? Could the function be simplified to: `get_submodule`? This would shorten the function name and make it more generic. Then however, it logging the warning message feels to "specific" to the use cases. Maybe removing the warning is actually fine?

### DIFF
--- a/avalon/houdini/pipeline.py
+++ b/avalon/houdini/pipeline.py
@@ -219,7 +219,7 @@ def ls():
     if hasattr(config_host, "collect_container_metadata"):
         has_metadata_collector = True
 
-    for container in sorted(container_names):
+    for container in sorted(containers):
         data = parse_container(container)
 
         # Collect custom data if attribute is present
--- a/avalon/houdini/pipeline.py
+++ b/avalon/houdini/pipeline.py
@@ -38,10 +38,6 @@ def install(config):
     pyblish.api.register_host("hpython")
 
     self._has_been_setup = True
-
-    config_host = find_module_in_config(config, "houdini")
-    if hasattr(config_host, "install"):
-        config_host.install()
         
 
 def uninstall(config):
@@ -53,10 +49,6 @@ def uninstall(config):
         config: configuration module
 
     """
-
-    config_host = find_module_in_config(config, "houdini")
-    if hasattr(config_host, "uninstall"):
-        config_host.uninstall()
 
     pyblish.api.deregister_host("hython")
     pyblish.api.deregister_host("hpython")
--- a/avalon/houdini/pipeline.py
+++ b/avalon/houdini/pipeline.py
@@ -10,7 +10,7 @@ import hou
 
 # Local libraries
 from . import lib
-from ..lib import logger
+from ..lib import logger, find_module_in_config
 from avalon import api, schema
 
 from ..pipeline import AVALON_CONTAINER_ID
@@ -39,10 +39,10 @@ def install(config):
 
     self._has_been_setup = True
 
-    config = find_host_config(config)
-    if hasattr(config, "install"):
-        config.install()
-
+    config_host = find_module_in_config(config, "houdini")
+    if hasattr(config_host, "install"):
+        config_host.install()
+        
 
 def uninstall(config):
     """Uninstall Houdini-specific functionality of avalon-core.
@@ -54,25 +54,13 @@ def uninstall(config):
 
     """
 
-    config = find_host_config(config)
-    if hasattr(config, "uninstall"):
-        config.uninstall()
+    config_host = find_module_in_config(config, "houdini")
+    if hasattr(config_host, "uninstall"):
+        config_host.uninstall()
 
     pyblish.api.deregister_host("hython")
     pyblish.api.deregister_host("hpython")
     pyblish.api.deregister_host("houdini")
-
-
-def find_host_config(config):
-    config_name = config.__name__
-    try:
-        config = importlib.import_module(config_name + ".houdini")
-    except ImportError as exc:
-        if str(exc) != "No module name {}".format(config_name + ".houdini"):
-            raise
-        config = None
-
-    return config
 
 
 def get_main_window():
@@ -236,13 +224,17 @@ def ls():
                        "pyblish.mindbender.container"):
         containers += lib.lsattr("id", identifier)
 
-    for container in sorted(containers):
+    has_metadata_collector = False
+    config_host = find_module_in_config(api.registered_config(), "houdini")
+    if hasattr(config_host, "collect_container_metadata"):
+        has_metadata_collector = True
+
+    for container in sorted(container_names):
         data = parse_container(container)
 
         # Collect custom data if attribute is present
-        config = find_host_config(api.registered_config())
-        if hasattr(config, "collect_container_metadata"):
-            metadata = config.collect_container_metadata(container)
+        if has_metadata_collector:
+            metadata = config_host.collect_container_metadata(container)
             data.update(metadata)
 
         yield data
--- a/avalon/houdini/pipeline.py
+++ b/avalon/houdini/pipeline.py
@@ -11,7 +11,7 @@ import hou
 # Local libraries
 from . import lib
 from ..lib import logger, find_submodule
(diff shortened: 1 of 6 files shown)


BigRoy:
Actually, this could be done much easier. 

We should remove the code from the individual host integrations and move it to the global install function. Will update the PR.

BigRoy:
With https://github.com/getavalon/core/pull/501/commits/3f1aa64bc8e53cde982f09f67a213a13c2e03408 the Host Integrations now don't need to manually call `install()` and `uninstall()` on the configuration.

The `{config}.{host}.install()` and `{config}.{host}.uninstall()` are called from respectively `avalon.pipeline.install()` and `avalon.pipeline.uninstall()` now when the studio configuration has these methods exposed.

avalon/lib.py
```diff
@@ -281,3 +281,23 @@ def modules_from_path(path):
...
+        module (str): The config's submodule to search.
+    
+    Returns:
+        types.ModuleType or None: The module, if found.
+        
+    """
+    config_name = config.__name__
+    module_name = "%s.%s" % (config_name, module)
+    try:
+        return importlib.import_module(module_name)
```
davidlatwe: Houdini and Nuke raised `NameError: global name 'importlib' is not defined` at this line.

avalon/lib.py
```diff
@@ -281,3 +281,23 @@ def modules_from_path(path):
         modules.append(module)
 
     return modules
+    
+
```
davidlatwe: A few places pop PEP8 warning due to blank line containing white space.

davidlatwe:
> However I believe this was frowned upon by the code style, but I couldn't find that mentioned anywhere.

I think it's here, in [CONTRIBUTING # Code Quality - Architecture](https://github.com/getavalon/core/blob/master/CONTRIBUTING.md#code-quality---architecture) at row A8.

> Could the function be simplified to: get_submodule?

Yeah, we can simplify that, but I lean to name it `find_submodule`, since I think the word `get` sounds like "You will get what you want for sure", and `find` is a bit uncertain which fit the scenario "You may not found the submodule you want, be careful".

> Then however, it logging the warning message feels to "specific" to the use cases. Maybe removing the warning is actually fine?

I more like to keep the message, and rephrase it to like:
```python
log_.warning("Could not find '%s' in module '%s'." % (submodule, module))
```
And the variables need to be renamed as well.

BigRoy:
Thanks @davidlatwe for the thorough check. Admittedly I wrote the previous code at home and it was untested (as mentioned in the original post) because I was on holiday leave. Back in the office Today, cleaned up the code, listened to the lovely hound that is back to his regular woof-woof (thanks Marcus!) and tested the integrations in Maya, Houdini and Fusion to be sure. All good now.

Can you double check the code style and let me know if there's still anything out of the ordinary?

avalon/lib.py
```diff
@@ -281,3 +282,23 @@ def modules_from_path(path):
...
+        types.ModuleType or None: The module, if found.
+
+    """
+    name = "%s.%s" % (module.__name__, submodule)
+    try:
+        return importlib.import_module(name)
+    except ImportError as exc:
+        if str(exc) != "No module name {}".format(name):
+            log_.warning("Could not find '%s' in module: %s" % (submodule,
+                                                                module))
```
jasperges: Just a minor thing: the preferred method for string formatting for logging functions is to pass arguments (see [here](https://docs.python.org/2/library/logging.html?#logging.Logger.debug) for more info). So I thinks it's better to write:

```python
log_.warning(
    "Could not find '%s' in module: %s",
    submodule,
    module,
)

File: avalon/houdini/pipeline.py [hunk 4]
```diff
@@ -10,8 +10,8 @@
 
 # Local libraries
 from . import lib
-from ..lib import logger
-from avalon import api, schema
+from ..lib import logger, find_submodule
+from avalon import api
```
jasperges: Shouldn't this be `from .. import api`?

BigRoy: It sure should. I just removed the `schema` part but will fix it right up! Well spotted.

avalon/lib.py
BigRoy: That's entirely true. I'll reformat hit. However I believe in many places the code is written with actual string formatting... which we should actually update too. Usually the difference is really only worth it for code that needs to be heavily optimized (runs very frequently) because the `logging` formatting in the function will only trigger when an actual log handler is connected to it and the message actually needs to be formatted. Which in our case usually is the case anyway.

avalon/lib.py
```diff
@@ -281,3 +282,23 @@ def modules_from_path(path):
...
+
+    Args:
+        module (types.ModuleType): The module to search in.
+        submodule (str): The submodule name to find.
+
+    Returns:
+        types.ModuleType or None: The module, if found.
+
+    """
+    name = "%s.%s" % (module.__name__, submodule)
```
jasperges: Maybe also use `.format()` here to be consistent?

avalon/houdini/pipeline.py
BigRoy: Resolved with https://github.com/getavalon/core/pull/501/commits/986157351ac5bd6ea8dd5d22cb1ec4ac892952c3

avalon/lib.py
BigRoy: Resolved with https://github.com/getavalon/core/pull/501/commits/6d894d13133e7726d9e76bf64fec1c11985732a9

avalon/lib.py
BigRoy: Resolved with: https://github.com/getavalon/core/pull/501/commits/5aea369f9d3818d89c81297a1e32764d34180c5f

BigRoy:
Just found [this odd exception printing](https://github.com/getavalon/core/pull/501/files#diff-95efa4b4467abab66473e907fd11d122R278-R280) which is not a change by this PR but has been there for quite some time. Is there any Python version where that actually prints as a formatted string? I don't believe so. 

**Python 2**
```python
err = "bar"
print("Skipped: \"%s\" (%s)", mod_name, err)
# ('Skipped: "%s" (%s)', 'foo', 'bar')
```

**Python 3**
```python
err = "bar"
print("Skipped: \"%s\" (%s)", mod_name, err)
# Skipped: "%s" (%s) foo bar
```

I'll patch that up too.

BigRoy:
[This](https://github.com/getavalon/core/pull/501#issuecomment-570183899) is fixed with: https://github.com/getavalon/core/pull/501/commits/241e05d83e53ad128dbb3038dab493122cb3fea9

Retested in Maya + Houdini to be sure. All seems to work as expected. 👍

jasperges:
@BigRoy Did a quick check with Blender with the changes you mentioned and everything seems to work fine.

BigRoy:
Merging this.

Issue #474, opened 2019-11-08
Re-Implement Nuke containerizing


davidlatwe:
# Re-Implement Nuke containerizing

### Issue/Problem

* Loaded nodes were not actually containerized, only data imprinted.
* Due to the above, current method for listing container is scaning all nodes and using try-except to find out which node is able to be parsed. I think this could be improved.
* Does `avalon-core` need container data updating method to be implemented here (`pipeline.update_container`) ? Might need to be disscussed.

#### Containerizing

Using a `Group` node as an Avalon main container, and after Loader loads subset into root space or anyother sub-space, *clone* the subset into the container with data imprinted.

The goal of grouping and cloning loaded resources into one place, was to provides a certain block for maintaining.

Example code:

```python
AVALON_CONTAINERS = "AVALON_CONTAINERS"
def containerise(name,
                 namespace,
                 node,
                 context,
                 loader=None,
                 data=None):
    main_container = nuke.toNode(AVALON_CONTAINERS)
    if main_container is None:
        main_container = nuke.createNode("Group")
        main_container.setName(AVALON_CONTAINERS)
        # Maybe tweaking the tile_color and font_size
        ...

    # Format a container name
    container_name = "{}_{}".format(name, suffix or "CON")
    node.setName(container_name)

    # Clone it into main container
    main_container.begin()
    nuke.clone(node)
    main_container.end()

    # Imprint data and other stuff
    data = {
        "schema": "avalon-core:container-2.0",
        "id": AVALON_CONTAINER_ID,
        "name": name,
        "namespace": namespace,
        "loader": str(loader),
        "representation": str(context["representation"]["_id"]),
    }
    ...
```

Note that the third arg `node` is one `nuke.Node` object, not like in other hosts were a list of nodes.

#### Listing containers

If above make sense, now we have a place to look up all containers by the `id` knob.

Like..

```python
def ls():
    main_container = nuke.toNode("AVALON_CONTAINERS")

    for node in nuke.allNodes(group=main_container):
        knob = "root.AVALON_CONTAINERS.%s.id" % node.name()
        if nuke.exists(knob) and nuke.knob(knob) == AVALON_CONTAINER_ID:
            data = parse_container(node)
            # Collect custom data via `config.collect_container_metadata`
            ...
            yield data
```

Or, impelement `lib.lsattr` just like other hosts and `ls` from there

```python
# In lib.py
def lsattr(attr, value=None):
    if value is None:
[...]

PR #504, opened 2020-01-07
Implement #474

davidlatwe:
This PR is aimed to implement #474, I'll try to be as clear as possible.

## What's changed
1. Fixed a bug in `lib.reset_selection` (919f1ca)
2. Set logging message level to `INFO` (56f55fc)
3. Implemented `lsattr` and `lsattrs` for finding nodes by knob and value (1ff01fa)
4. Re-implemented subset containerizing (9df1197, 75e82c8, 3c685d0, 134b4cd, 6a11ea7)
5. Implemented node update/override system (c84473b)
6. Changed `models.TreeModel` and `sceneinventory.app` for Qt4 compat (98d3562)
7. Keep old style `containerize` and `ls` by env var `AVALON_NUKE_OLD_CONTAINER` (bd88c1a, a19b066)
8. Replaced `lib.imprint`, `lib.read` with vendorized `knobby` (05b5086, f217fa4, 875f83f)
9. Removed *avalonDataGroup* knob from Avalon Tab (f11e595)

### lsattr, lsattrs
These two functions have three extra args compares to other hosts, `type=None`, `group=None` and `recursive=False`. These three args will be passed into `nuke.allNodes` for defining the node search range.

### Subset containerizing
The new workflow of subset containerizing, will keep a copy (not clone) of loaded subset inside a `AVALON_CONTAINERS` group node.

image

image

Since the new `containerise` and `ls` of this workflow is not backwards compatible, you have to set `AVALON_NUKE_OLD_CONTAINER` in environment so the old style will be enabled.

But I am not sure about the name of that environment variable, so any suggestion is welcome.

### Node update/override system
Since the subset will be containerized, we have at least one copy of subset being presented outside of main container for artist to operate on. So we need a way to update all copies when updating subset.

This feature was mentioned [here](https://github.com/getavalon/core/issues/474#issuecomment-557462698), the implementation is a context manager `lib.sync_copies`.

Every nodes that being containerized will have `avalonId` and `containerId` imprinted, the `containerId` is the container's `avalonId`. And we look for copies by matching node type and `avalonId`.

To use it in Loader plugin, like `containerise` in `Loader.load`, you would call it in `Loader.update`, for example:
```python
class Loader(avalon.api.Loader):

    def update(self, container, representation):

        with lib.sync_copies(nodes):
            # Update subset
            for node in nodes:
                self.update_file_path(node)
                ...
        # All copies of `nodes` are auto updated by matching `avalonId`

        with lib.sync_copies([container_node], force=True):
            # Update container data
            ...
        # All copies of `container_node` updated

```

### Vendoring [`knobby`](https://github.com/davidlatwe/nuke-knobby)
There were some issues in previous implementation of knob imprinting, like:

1. Creating duplicate named knob without raising any error from Nuke
2. Knob name prefix being an option on `imprint` might not good for `read`

I have tried to improve and resolved, but the code was a bit complicate for being in `lib`. Then I thought maybe to move it out as a standalone module would be better. If you don't like it to be in `nuke.vendor`, I would love to move it back.

So what `knobby` will do for us ? TL;DR

1. It will merge new and old knobs then re-build all user knobs if the imprinting `tab` existed.
2. Knob will be auto prefixed with it's tab name.
3. New method `mold` is like `read`, but will preserve the tab hierarchy.

I think this is much stable and should be a bit more straight forward when writing/reading data to/from node than previous implementation.

### Drop `avalonDataGroup` knob
I though the Avalon data tab could be simplified, so I have removed it from `set_avalon_knob_data`. If you don't mind.

### Qt4 support
I was tested in Nuke 9 which still using `PySide`, so I had encountered some errors while using scene inventory app. Fixed them all together.

---

Please have a test if you could. :)


tokejepsen:
Looks very good!

> Since the new containerise and ls of this workflow is not backwards compatible, you have to set AVALON_NUKE_CONTAINERS_AT_LARGE in environment so the old style will be enabled.

I was wondering whether this is necessary?
I know backward compatibility is a must, but the Nuke implementation for Avalon seems somewhat in a beta stage where we are still ironing things out, so the question is whether enough people are using it to warrant the added complexity of this.

davidlatwe:
Yeah, it's true that Nuke's implementation in Avalon official release should be considered as beta stage.

The main reason I add that environment variable was actually for *pypeclub* implementation, so they could pick up without to much trouble. (I hope !)

Would be great if we don't need that. :relaxed: 
Pinning @jezscha :telephone_receiver: , is that okay if we drop the old containerization implementation ?

avalon/nuke/vendor/knobby/util.py
```diff
@@ -0,0 +1,320 @@
...
+            knob_name = knob.name()
+            if not knob_name:
+                # Ignore unnamed knob
+                continue
+
+            knob_type = nuke.knob(knob.fullyQualifiedName(), type=True)
+            value = knob.value()
+
+            if (
+                knob_type not in EXCLUDED_KNOB_TYPE_ON_READ or
```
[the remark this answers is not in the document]
davidlatwe: Should add `*/vendor/*` into flake8 exclusion list.

davidlatwe:
> I was wondering whether this is necessary?

I think I'll keep the backward comapt for now, untill everyone who use Nuke can confirm that we don't need it.

And I've changed the name from `AVALON_NUKE_CONTAINERS_AT_LARGE ` into `AVALON_NUKE_OLD_CONTAINER`, the new name should be much easier to know what's it for.

Currently I am still working on my Nuke config, but what's inside this PR is working fine here, so hope we can merge this soon if no objections.

Would be great if anyone can have a try out. :)

avalon/nuke/vendor/knobby/util.py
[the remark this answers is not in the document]
jasperges: Actually hitting the same violation in my Blender pull request. If you check [flake8 rules](https://www.flake8rules.com/rules/W503.html) and [pep8](https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator) I can think we can safely ignore this.

It seems breaking around a binary operator is seen as a violation, although it's not.

tokejepsen (approved):

avalon/nuke/vendor/knobby/util.py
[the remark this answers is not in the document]
mkolar: the best practice here is deciding on one and telling hound in the `flake8` config file to ignore the other variant. Line break before and after binary op have different error codes.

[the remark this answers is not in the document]
davidlatwe: I kind of like binary op after line break. ☺️

BigRoy:
What's the status of this long open PR? :) Should this be ready to merge or are there reasons not to?

avalon/tools/models.py
```diff
@@ -62,7 +62,8 @@ def setData(self, index, value, role=QtCore.Qt.EditRole):
                 item[key] = value
 
                 # passing `list()` for PyQt5 (see PYSIDE-462)
-                self.dataChanged.emit(index, index, list())
+                args = () if Qt.IsPySide or Qt.IsPyQt4 else ([],)
```
iLLiCiTiT, line 65: There should be role in the list. `args = () if Qt.IsPySide or Qt.IsPyQt4 else ([role],)`

davidlatwe:
> Should this be ready to merge or are there reasons not to?

Here in my studio, the Avalon-Nuke implementation was just started to run in production, so I am not sure about the implementation of `AVALON_CONTAINER` is good enough or not (exposing it in the node graph). Also, not sure how this PR will compat with the implementation that Pype have.

So I guess we might need to wait a bit. 🤔

avalon/tools/models.py
davidlatwe: Yeah, I think you are right !

mkolar:
To be perfectly honest, we didn't find time to test this properly. Code wise it looks very nice, but we were swamped and couldn't afford the time to test. I'm truly hoping that next week we can give it a day.

tokejepsen:
Testing this PR out atm, and will dump any thoughts and feedback here.

tokejepsen:
I think the backdrop workflow within the `AVALON_CONTAINERS` group could use from documentation and code. For example as a developer how do I get the nodes for a container in the group?
I would support the documentation with an exposed method like this;
```
def get_container_nodes(container_name):
    # Enter container group.
    groupNode = nuke.toNode("AVALON_CONTAINERS")
    groupNode.begin()
    backdrop = nuke.toNode(container_name)
    
    # Clear selection.
    for node in nuke.allNodes():
        node["selected"].setValue(False)
    backdrop.selectNodes()
    
    # Get backdrop nodes.
    backdrop_nodes = []
    for node in nuke.allNodes():
        if node["selected"].getValue():
            backdrop_nodes.append(node)
    groupNode.end()

    return backdrop_nodes
```

tokejepsen:
Finding working with a context manager somewhat confusing. Take this example:
```python
container_nodes = get_container_nodes(container["objectName"])
with avalon.nuke.sync_copies(container_nodes):
    pass
```
Would it not be easier read if we had just a method to call:
```python
container_nodes = get_container_nodes(container["objectName"])
avalon.nuke.sync_copies(container_nodes)
```

Also it might be better to have the `sync_copies` mirror `find_copies` when it comes to arguments. `find_copies` takes a single node as source, but `sync_copies` takes a list as source.

davidlatwe:
Just found out that `sync_copies` may failed if currrent Nuke session has multiple views, since the knob name returned from `knob.fullyQualifiedName()` will append view name at the end.

7374745 resolved this.

davidlatwe:
@tokejepsen 
> Finding working with a context manager somewhat confusing.

The reason why I made it a context was because, it need to firstly compare which copied knob hasn't been changed against source knob before updating them. Or it doesn't need to be that way ? 😮

avalon/nuke/pipeline.py
```diff
@@ -105,20 +171,18 @@ def parse_container(node):
...
     # Store the node's name
     data["objectName"] = node["name"].value()
     # Store reference to the node object
     data["_node"] = node
+    # Get containerized nodes
+    if node.fullName() == "%s.%s" % (AVALON_CONTAINERS, node.name()):
+        data["_members"] = lib.lsattr("avalon:containerId",
+                                      value=data["avalonId"],
+                                      group=nuke.toNode(AVALON_CONTAINERS),
+                                      recursive=True)
```
davidlatwe, line 185: @tokejepsen 
> I think the backdrop workflow within the `AVALON_CONTAINERS` group could use some documentation and code. For example as a developer how do I get the nodes for a container in the group?

Yeah, you are right, that seems very implicit. Currently it looks like you can only get them from the entry `"_member"` in the parsed container.

Can't remember why I implement it like this. :/
Any suggestion ?

tokejepsen:
> The reason why I made it a context was because, it need to firstly compare which copied knob hasn't been changed against source knob before updating them. Or it doesn't need to be that way ?

I just cant see why we are using a context, since nothing gets reverted after the context finishes?
We get the same behaviour from just calling a method.

avalon/nuke/pipeline.py
tokejepsen: Hmm, long time since I posted this, but I think generally I found it confusing about how to fetch nodes to pass to `sync_copies`.
Ideally we would work with similar arguments to get the nodes in the container group and what to pass to `sync_copies`.

davidlatwe:
Here I made an example to show the sync workflow:

```python
from avalon.nuke import lib

def copy_one(node):
    node["selected"].setValue(True)
    nuke.nodeCopy("_copy_")
    nuke.nodePaste("_copy_")
    copied = nuke.selectedNodes()[0]
    copied["selected"].setValue(False)
    return copied

# Make a dummy that will be containerized in real production
source = nuke.createNode("NoOp", inpanel=False)

# Tag an id so we could find it's copies later
lib.set_id(source)

# Make copies
copied_1 = copy_one(source)
copied_2 = copy_one(source)

# Change the second copy to simulate artist input
copied_2["label"].setValue("hi")

# Update source and sync
with lib.sync_copies([source]):
    source["label"].setValue("hello")

assert source["label"].value() == copied_1["label"].value()
assert source["label"].value() != copied_2["label"].value()

```

As you may see, the second copy's label that was touched by artist is stay unchaged since it has been identified as an artist's demand while entering the context, but the first copy is synced at the end of the context.

davidlatwe:
> As you may see, the second copy's label that was touched by artist is stay unchaged since it has been identified as an artist's demand while entering the context, but the first copy is synced at the end of the context.

This was meant to behave just like Maya applying reference edits. 🤔

davidlatwe:
Sorry, one additional note. Without context manager to compare current state of source and copies, we will not able to know which knob that has been changed by artist, and ends up like a forced update.

tokejepsen:
> Sorry, one additional note. Without context manager to compare current state of source and copies just right before updating, we will not able to know which knob that has been changed by artist, and ends up like a forced update.

Good point! I see the need for the context now.

davidlatwe:
Looks like I did not carefully test on forced sync !
Just found a bug that node name will be synced when force enabled, which no one will expect that happen !
Resolved with 3129b24.

avalon/nuke/pipeline.py
davidlatwe: Not sure if this helps, currently I only use it inside our loader plugin's update method, like this

https://github.com/MoonShineVFX/reveries-config/blob/a5c56c5d3ee6e87498b024004a41e0b80fc48501/plugins/nuke/load/load_renderlayer.py#L244-L257

Haven't had the needs to find backdrop(container) nodes in other cases.
Maybe we could decouple this
```python
# Get container backdrop child nodes
lib.lsattr("avalon:containerId",
           value=data["avalonId"],  # container's id
           group=nuke.toNode(AVALON_CONTAINERS),
           recursive=True)
```
into another function and expose it ?

tokejepsen: Yeah, sounds like a good idea.

PR #510, opened 2020-01-10
Turn off Hound's "Black would make changes" msg (flake8-black)

### LINKED ISSUE #506: Add config for Black or disable the 'Black check' for Hound CI
It seems Hound CI recently (see for example [this question](https://gitter.im/getavalon/Lobby?at=5e134ffc316c1473272624ef) by @BigRoy) started to check the code with [Black](https://github.com/psf/black). The only output you get from the Hound is "Black would make changes." So it seems it runs `black --check` on your code.

If this check is wanted, I would suggest to add a config for Black in the repository, unless the default options are totally fine. To add specific config parameters a [`pyproject.toml` file](https://github.com/psf/black#pyprojecttoml) has to be added with configuration options for Black.
As @davidlatwe mentioned [here](https://gitter.im/getavalon/Lobby?at=5e15906dc39503490263b4f1) the default values seem to conflict with PEP8. That is something that could probably be fixed in the config.

How do people feel about this?

davidlatwe:
This PR will fix #506 , the solution was found in houndci/hound#1769.
:rocket:

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,2 @@
+[flake8]
+ignore = BLK100
--- a/.hound.yml
+++ b/.hound.yml
@@ -1,4 +1,8 @@
 python:
   enabled: true
 
+flake8:
+  enabled: true
+  config_file: .flake8
+
 fail_on_violations: true


mottosso:
Yes, haha. I was searching through their poor documentation looking for this but couldn't find it. This is clearly a blunder on their part that I expect will disappear on its own, but we should definitely ignore this either way. 👍

davidlatwe:
Merging this !

jasperges:
Awesome! Thanks @davidlatwe.

Issue #512, opened 2020-01-16
Make debug behaviour consistent for all tools


jasperges:
When doing #470 I missed some similar calls in the other tools. I would like to make it consistent across all the tools and avoid confusing behaviour.

Things I would like to do:

- Add the `debug` variable to the `show()` function of **all** the tools.
- Remove anything that changes the context/session, for example the current asset, project, etc. That can be used when running tests, but should never happen when debugging. It actually makes debugging a lot harder!
- Include the following ([see snippet below](#excepthook)) when `debug=True`. That way hosts who don't rely on Qt themselves don't crash when there is an error in one of the tools. As mentioned [here](https://github.com/getavalon/core/pull/466#issuecomment-572183101) this might actually be quite dangerous. So the other option would be to remove this from all tools. The most important thing is to make it **consistent**.

---

<a name="excepthook"></a>
```python
if debug:
    import traceback
    sys.excepthook = lambda typ, val, tb: traceback.print_last()
```

BigRoy:
Consistency sounds good! Perfect.

> sys.excepthook = lambda typ, val, tb: traceback.print_last()

I wouldn't bother actually, but would instead mention it somewhere as a recommendation for manual debugging in case of crashes in e.g. blender. I am a bit wary as including it anywhere by default, even for debug mode. However, I am not fully against it so will leave this up the general consensus of everyone. Just giving my gut feeling.

jasperges:
My gut feeling actually says the same. I also updated this point, to make it clear what I mean. The thing is that some tools already have this and some don't. It's not consistent.

jasperges:
@mottosso Do you have an opinion about this? Especially about the _excepthook_? Should we add it to all tools or remove it from all tools? I am inclined to do the latter.

mottosso:
Oh, no this shouldn't be per tool. It should be per install of Avalon. There's an install per integration, like the one we have [for Maya](https://github.com/getavalon/core/blob/add240f6e29f71322be72c6ad6a6ac46dd2f2b8e/avalon/maya/pipeline.py#L36). That'll happen once, and can take things like this into consideration. Either as an argument, or environment variable.

In addition, we shouldn't overwrite it point-blank, we should try and preserve whatever was there initially. At least then we prevent half of the problems from occurring (i.e. not accounting for someone overwriting it *after* Avalon).

```py
previous_excepthook = sys.excepthook

def avalon_excepthook():
  previous_excepthook()
  # Our business here

[...]

PR #513, opened 2020-01-16
Make debug behaviour consistent

jasperges:
**What's changed?**
This implements #512. In this PR I have included the override for `sys.excepthook`, but that can be removed easily.

### DIFF
--- a/avalon/tools/contextmanager/app.py
+++ b/avalon/tools/contextmanager/app.py
@@ -211,7 +211,7 @@ class App(QtWidgets.QDialog):
         self._assets.select_assets([assetname], expand=True)
 
 
-def show(parent=None):
+def show(parent=None, debug=False):
 
     from avalon import style
     from ...tools import lib
@@ -220,6 +220,10 @@ def show(parent=None):
         del module.window
     except (RuntimeError, AttributeError):
         pass
+
+    if debug:
+        import traceback
+        sys.excepthook = lambda typ, val, tb: traceback.print_last()
 
     with lib.application():
         window = App(parent)
--- a/avalon/tools/creator/app.py
+++ b/avalon/tools/creator/app.py
@@ -569,22 +569,8 @@ def show(debug=False, parent=None):
         del(module.window)
 
     if debug:
-        from avalon import mock
-        for creator in mock.creators:
-            api.register_plugin(api.Creator, creator)
-
         import traceback
         sys.excepthook = lambda typ, val, tb: traceback.print_last()
-
-        io.install()
-
-        any_project = next(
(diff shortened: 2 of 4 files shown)


jasperges:
For people reading this: after this PR there is no tool which calls special code when in debug mode. For now I have a simple if block:

```python
if debug:
    # Add code here that only runs when debugging is enabled.
    pass
```

Should I keep this or remove it completely? The only reason to keep it would be so that it's super clear that code can be run only for debug mode. In case someone wants to add debug code for the tools.

tokejepsen:
Hmm, if none of the tools have any debug code, maybe just remove it all?

jasperges:
As mentioned [here](https://github.com/getavalon/core/pull/523#issuecomment-581920080) the `Loader` tool handles an existing module window in a different way. Although it's not directly related to the debug behaviour, I could change that code too, so it's consistent with the other tools.

@BigRoy it seems you implemented this. Did you have a specific reason for it? Do you think it's fine to change it?

BigRoy:
@jasperges I can't recall. Looking at the code it seems to solve the issue where the Window might have been garbage collected. I can't recall an actual case of when/where that might have happened. Looking at the blame it seems the [original `try..except` statement came from three years](https://github.com/getavalon/core/commit/18a85c604a6b7df5793813b17a6014407da4f61e) ago by @mottosso 

I guess I just ended up preserving that behavior. Thus no need to keep it around for me if it works without it.

jasperges:
@BigRoy Okay, then I will remove it and make it consistent with the other tools.

jasperges:
To be honest I have not tested this properly myself. Also the whole thing got a little bit bigger then initially intended. Do you think this is okay or would it be better to got back to just making the debug behaviour consistent and then open another PR for getting all `show` related code consistent?

BigRoy:
To me the current changes don't seem too large for a single PR. We'll just need to make sure things work as we intended and we're good to go.

BigRoy (approved):
I think the simplified code is good for the code base. 

I do feel that it's a minor loss that clicking "Avalon > Load.." again in for example Maya that after this change it will close any existing loader window if it's open and show a new one (meaning it's reset to default state!). Where previously it would just show and raise the existing window.

I guess if everyone kind of agrees that it's the preferred change then let's get this merged.

davidlatwe (approved):
Looks like try closing `module.window` on `show` was adopting from `sceneinventory` and `projectmanager`.

For `loader` app, it's a minor loss indeed. But in return is simplicity for all DCC App/OS integrations, so +1 for this change :P

tokejepsen (approved):

PR #515, opened 2020-01-16
Remove unneeded `config` variable from host (un)install (closes #505)

### LINKED ISSUE #505: No need to pass 'config' to 'host.(un)install'
When changing the Blender integration so it works with #501, I noticed that the `install()` and `uninstall()` functions still expect the `config` variable. If I'm not mistaken this is not needed anymore and it can be safely removed after changing [this line in avalon.pipeline](https://github.com/getavalon/core/blob/master/avalon/pipeline.py#L79) to `host.install()`.

Is there a reason not to change this? I can make a pull request if needed.

jasperges:
**What's changed?**
Removed the config variable from the `host.install` and `host.uninstall` functions. They were still expected, but not used anymore. After #501 was merged, host integrations don't need to care about the config anymore.

### DIFF
--- a/avalon/fusion/pipeline.py
+++ b/avalon/fusion/pipeline.py
@@ -49,7 +49,7 @@ def ls():
             yield container
 
 
-def install(config):
+def install():
     """Install Fusion-specific functionality of avalon-core.
 
     This function is called automatically on calling `api.install(fusion)`.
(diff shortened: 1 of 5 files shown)


jasperges:
@BigRoy @davidlatwe Any reason not to merge this? Should somebody else also have a look at it?

BigRoy:
Sorry - bit busy. I'm expecting this is all good looking at the code. Since this has seen some signs of approval I'll merge it right away.

PR #520, opened 2020-01-27
Fix #514

### LINKED ISSUE #514: Make sure the QApplication doesn't get garbage collected
Following up on [this 'bug', reported on the Avalon Gitter](https://gitter.im/getavalon/Lobby?at=5e1dd36165badf754d886353).

The `show()` function for the tools creates a `QApplication` if needed (via `tools.lib.application()`). The problem however is that the `QApplication` created in this way will possibly be garbage collected which in turn causes issues with for example `module.window.close()`.
This makes this 'convenience helper' not so convenient anymore. I think we should fix it, so the `QApplication` will never be garbage collected. Else it doesn't serve it's purpose and an integration is better off creating the `QApplication` by itself.

tokejepsen:
**What's changed?**
- Each tool gets destroyed on close.
- Each tool gets raised if there is an existing window.
- All tools' windows gets activated.

### DIFF
--- a/avalon/tools/contextmanager/app.py
+++ b/avalon/tools/contextmanager/app.py
@@ -2,7 +2,8 @@ import sys
 import logging
 
 from ... import api
-
+from avalon import style
+from ...tools import lib
 from ...vendor.Qt import QtWidgets, QtCore
 from ..widgets import AssetWidget
 from ..models import TasksModel
@@ -20,6 +21,7 @@ class App(QtWidgets.QDialog):
     def __init__(self, parent=None):
         QtWidgets.QDialog.__init__(self, parent)
 
+        self.setAttribute(QtCore.Qt.WA_DeleteOnClose, True)
         self.resize(640, 360)
         project = api.Session["AVALON_PROJECT"]
         self.setWindowTitle("Context Manager 1.0 - {}".format(project))
@@ -211,19 +213,21 @@ class App(QtWidgets.QDialog):
         self._assets.select_assets([assetname], expand=True)
 
 
+def create_window(parent):
+    window = App(parent)
+    window.show()
+    window.setStyleSheet(style.load_stylesheet())
+    module.window = window
+
+
 def show(parent=None):
-
-    from avalon import style
-    from ...tools import lib
-    try:
-        module.window.close()
-        del module.window
-    except (RuntimeError, AttributeError):
-        pass
-
     with lib.application():
-        window = App(parent)
-        window.show()
-        window.setStyleSheet(style.load_stylesheet())
+        if lib.existing_app:
+            try:
+                module.window.raise_()
+            except (RuntimeError, AttributeError):
+                create_window(parent)
+        else:
+            create_window(parent)
 
-        module.window = window
+        module.window.activateWindow()
--- a/avalon/tools/creator/app.py
+++ b/avalon/tools/creator/app.py
@@ -118,6 +118,7 @@ class Window(QtWidgets.QDialog):
 
     def __init__(self, parent=None):
         super(Window, self).__init__(parent)
+        self.setAttribute(QtCore.Qt.WA_DeleteOnClose, True)
         self.setWindowTitle("Instance Creator")
         self.setFocusPolicy(QtCore.Qt.StrongFocus)
 
@@ -553,6 +554,13 @@ class FamilyDescriptionWidget(QtWidgets.QWidget):
         self.help.setText(help)
 
 
+def create_window(parent):
+    window = Window(parent)
+    window.show()
+    window.setStyleSheet(style.load_stylesheet())
+    module.window = window
+
+
 def show(debug=False, parent=None):
     """Display asset creator GUI
 
@@ -563,10 +571,6 @@ def show(debug=False, parent=None):
             to this QObject.
 
     """
-
-    if module.window:
-        module.window.close()
-        del(module.window)
 
     if debug:
         from avalon import mock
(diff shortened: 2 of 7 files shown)


mottosso:
Say what? :O How can all tools share a single `list.existing_app`? Is `app` referring to their window instances? Also `existing_app` isn't the name of a boolean, better make it `has_existing_app`. I expected an instance of the `QApplication` there, is that not what it should refer to?

tokejepsen:
> How can all tools share a single lib.existing_app?

They are all sharing the same `QApplication`, but the existing code fails when trying to show the tools in a standalone `QApplication`. See #514 for more details.
It seems like the window instance is kept on the module level, but the `QApplication` gets deleted so we end up with a segfault when showing the tool for the second time.

> Also existing_app isn't the name of a boolean, better make it has_existing_app. 

Fair enough. Can change that.

avalon/tools/contextmanager/app.py
```diff
@@ -211,19 +213,21 @@ def select_asset(self, assetname):
...
-        window = App(parent)
-        window.show()
-        window.setStyleSheet(style.load_stylesheet())
+        if lib.existing_app:
+            try:
+                module.window.raise_()
+            except (RuntimeError, AttributeError):
+                create_window(parent)
+        else:
+            create_window(parent)
```
mottosso: Is this case tested? What do we expect to happen? If there isn't a `QApplication`, then we probably can't create a window.

tokejepsen: I've tested this in Nuke/Maya and standalone (Photoshop).

If there isn't a `QApplication` the `with lib.application()` ensure there is a `QApplication`.

mottosso: But isn't `lib.application()` happening first? :O How can `existing_app` be False, if `application()` ensures it is true?

avalon/tools/contextmanager/app.py
```diff
@@ -20,6 +21,7 @@ class App(QtWidgets.QDialog):
     def __init__(self, parent=None):
         QtWidgets.QDialog.__init__(self, parent)
 
+        self.setAttribute(QtCore.Qt.WA_DeleteOnClose, True)
```
mottosso: If this is true, then the instance we store in `module.window` will be invalidated. It is deliberately *not* deleted such that it can be reused later. The window is preserved to also preserve e.g. user selection and focus within the app. I think we want those things?

mottosso:
This PR seems to conflate a few things.

#514 seems to be about QApplication being garbage collected, then why not just..

**avalon/blender/pipeline.py**

```py
module = sys.modules[__name__]

def install()
  module.qapp = QApplication.instance()
  # Or make one, and store it here.
```

That would preserve it for the lifetime of the inner Python interpreter.

avalon/tools/contextmanager/app.py
tokejepsen: Good point. Will remove that.

tokejepsen: Actually looking closer, it seems like we are deleting the old window anyways: https://github.com/getavalon/core/blob/master/avalon/tools/contextmanager/app.py#L218-L222

mottosso:
> Note how the QApplication instance still exists but it is not executing anymore

Oh wow, that is a good find. I had no idea it would do that.

> If there's a way that we can check whether the QApplication is currently executing at that point in time we can then trigger another exec_() in that else statement if it is not running currently.

The simple solution would be to just call it again, regardless. If it *was* running, then it would block and stop you from running it again, and if it wasn't blocking, then having it block should be normal.

tokejepsen:
> So I've done just that and tried to reproduce it there.

I cant seem to replicate the behaviour you are seeing?

Untitled_ Jan 29, 2020 9_25 AM

BigRoy:
@tokejepsen so you are running all commands at the same time. As such, after closing the first one it will instantly run through the second `exec_()` call.

Try this in a Python process and run them step by step, like the comments somewhat hinted. :)

BigRoy:
@tokejepsen so you are running all commands at the same time. As such, after closing the first one it will instantly run through the second `exec_()` call.

Try this in a Python process and run them step by step, like the comments somewhat hinted. :)

Or to show the hanging UI... remove the last `exec_()`  call.

tokejepsen:
@BigRoy thanks, yeah just realized you probably meant going through the python prompt. Seeing the same behaviour now.

tokejepsen:
Ok, I think I have a simple fix for this:

```python
@contextlib.contextmanager
def application():
    self.app = QtWidgets.QApplication.instance()

    if self.app:
        print("Using existing QApplication..")
        yield self.app
    else:
        print("Starting new QApplication..")
        self.app = QtWidgets.QApplication(sys.argv)
        yield self.app

    self.app.exec_()
```

Before:

Untitled_ Jan 29, 2020 10_23 AM

At the end you can see that the python process exits without any errors, which is what happens when the segfault occurs.

After:

Untitled_ Jan 29, 2020 10_29 AM

At the end you can see that the python process is still running, and I can show another `contextmanager` with the existing `QApplication`.

If we like this fix, I will revert the changes and update with this code change instead.

tokejepsen:
Any takers on this fix suggested above?

avalon/tools/contextmanager/app.py
davidlatwe: Currently only Loader's window get preserved, and I vote for keeping them :P

davidlatwe:
> Ok, I think I have a simple fix for this

I do like that fix. :)

BigRoy:
Sorry, totally missed your comments. 

I somehow expected that wouldn't fix it. But if it fixes all use cased by moving one line of code let's set up a PR and give it a couple of days of testing. Should we close this PR and set a new clean one? Or somehoq squash the commits to solely that change?

Or were there other changes that are required?

mottosso:
Seems appropriate. Does `exec()` not cause issues when e.g. Maya already has a QApplication?

davidlatwe:
> Does exec() not cause issues when e.g. Maya already has a QApplication?

It only return `-1` and nothing else happens.

PR #522, opened 2020-01-31
Activate each tool window.

tokejepsen:
**What's changed?**
Added an `window.activateWindow` to each tool.
When running the tools standalone triggered from a different window, the tools can get hidden behind the triggering window. This will bring them to the front of the user attention.


mottosso:
Roy is right, raising a window is a big mess cross-platform with Qt. Here's [what QML does](https://github.com/pyblish/pyblish-qml/blob/master/pyblish_qml/app.py#L165-L172). Each of those are necessary, in that order, to cover all edge cases and platforms. Though I've still heard some having issues with it appearing behind another window? It's been rare though, but took a while to figure out.

tokejepsen:
> Is this for Photoshop integration ?

Yes, this is in particular for the Photoshop integration where the python process runs in the background, so the windows get hidden behind the Photoshop window.

> For example here it describes using .raise() on Mac OS X.

I cant test on OSX but adding the call does hurt anything.

> Here's what QML does.

Using `WindowStaysOnTopHint` does help with the initial showing of the window as well!
Will need to account for when showing in DCCs, but can just query `parentWidget()`.

Will update the PR in a bit.

BigRoy:
The latest commit made it so that the Windows always stay on top of all Windows when standalone right? So even over other applications. I'm expecting we wouldn't want that. Correct?

tokejepsen:
> The latest commit made it so that the Windows always stay on top of all Windows when standalone right? So even over other applications. I'm expecting we wouldn't want that. Correct?

It only stays ontop of the windows within the QApplication, so not other OS windows.

BigRoy:
> It only stays ontop of the windows within the QApplication, so not other OS windows.

If that's the case then it's fine with me. But are you sure? 

I quickly tried on a random Qt application window that I had running in Houdini and with that flag set it would *never* go behind other applications. As such it was always on top in front of my Browser, Windows Explorer, everything. (except for Windows Task Manager, it would appear behind that one - until I disabled "Options > Always on Top" for Task Manager)

E.g. this always stays on top:
```python
# assumes a running QApplication instance
from PySide2 import QtWidgets, QtCore

class Window(QtWidgets.QDialog):

    def __init__(self, parent=None):
        super(Window, self).__init__(parent=parent)
        
        self.setWindowFlags(
                self.windowFlags() | QtCore.Qt.WindowStaysOnTopHint
        )
        
window = Window()
window.show()
```

tokejepsen:
> I quickly tried on a random Qt application window that I had running in Houdini and with that flag set it would never go behind other applications. As such it was always on top in front of my Browser, Windows Explorer, everything. (except for Windows Task Manager, it would appear behind that one - until I disabled "Options > Always on Top" for Task Manager)

The flag is only added when we use the tools standalone, so Maya/Nuke/Houdini would not be affected.

In standalone (Photoshop) I get this behaviour:

Untitled_ Feb 4, 2020 10_42 AM

BigRoy:
> In standalone (Photoshop) I get this behaviour:

That is odd. That is not at all what I'm seeing in my standalone runs. I wonder where the behavior differs.

I pulled your branch into my repository to give it a go and ran a standalone Loader from the Launcher. It stays on top of everything... then I run it again, suddenly it's allowed to go behind. Then I run it again, and it's the opposite. It seems a bit random.

It seems whenever I run that application and *do nothing* as a user in-between so it comes in with focus... then suddenly it will always stay on top of everything. If however I shift focus and continue in another browser for example during the launch of the application then suddenly the Show on Top behavior does not happen? @tokejepsen can you reproduce this?

window_stay_on_top

tokejepsen:
> It seems whenever I run that application and do nothing as a user in-between so it comes in with focus... then suddenly it will always stay on top of everything. If however I shift focus and continue in another browser for example during the launch of the application then suddenly the Show on Top behavior does not happen? @tokejepsen can you reproduce this?

I cant reproduce the oddness, but can reproduce the always ontop issue. I can just remove the flag setting?

BigRoy:
> I cant reproduce the oddness, but can reproduce the always ontop issue. I can just remove the flag setting?

👍 I think there's reason why @mottosso did [this for Pyblish QML too](https://github.com/pyblish/pyblish-qml/blob/master/pyblish_qml/app.py#L168-L172) so it could be worth actually doing that too if it's there with good reason. Or maybe it's redundant by now, no idea.

jasperges:
Tiling window manager anyone? Who needs floating windows, they are so 90's. :stuck_out_tongue: :grimacing:

tokejepsen:
How do we feel about this PR?

davidlatwe:
Just looked through the commits and comments, I think this is good to merge.
Let's merge this tomorrow if no other objections. 🚀

tokejepsen:
Apart of the errors with coveralls, this is ready to be merged now.

PR #524, opened 2020-02-05
Photoshop Integration

### LINKED ISSUE #516: Photoshop Integration
*Issue*

This is about a [Photoshop](https://www.photoshop.com/) (PS) Integration for Avalon.

*Implementation*

- PS does not support Python, but we can communicate with PS through its DOM (Document Object Model). Best starting resource is; https://github.com/lohriialo/photoshop-scripting-python.
- DOM communication with PS may not provide all the access we need. This needs to be investigated.
- Another communication method would be through socket/port, where PS invokes a server/client relationship with Avalon on demand.

tokejepsen:
**What's changed?**
This is the Photoshop integration which fixes #516

The best introduction to the integration is with the [README](https://github.com/tokejepsen/core/tree/photoshop/avalon/photoshop)


avalon/photoshop/lib.py
```diff
@@ -177,7 +177,7 @@ def _recurse_layers(layers):
     for layer in layers:
         result[layer.id] = layer
         if layer.LayerType == com_objects.constants().psLayerSet:
-            result.update(_recurse_layers([*layer.Layers]))
+            result.update(_recurse_layers([x for x in layer.Layers]))
```
BigRoy: Isn't `[x for x in layer.Layers]` similar to `list(layer.Layers)` @tokejepsen?

tokejepsen: Probably is. Do you want it changed?

tokejepsen:
Can we ignore the failed test on Linux, since this integration only works on Windows?

avalon/photoshop/lib.py
BigRoy: Not necessarily but I believe that's what is officially PEP08 recommended, but couldn't find it just now.

avalon/photoshop/README.md
```diff
@@ -0,0 +1,246 @@
...
+
+This integration requires the third part library `pywin32` which can be installed with:
+
+```
+pip install pywin32
+```
+
+## Setup
+
+The Photoshop integration requires to components to work; `extension` and `server`.
```
BigRoy: `to` -> `two`

avalon/photoshop/lib.py
```diff
@@ -0,0 +1,255 @@
+import json
+import contextlib
+import functools
+
+from win32com.client import Dispatch
+
+from . import com_objects
+from ..tools import html_server
+
+# Convinience variable that could later query whether the platform is Windows
```
BigRoy: Convinience -> Convenience

avalon/photoshop/lib.py
```diff
@@ -0,0 +1,255 @@
...
+# or Mac.
+# This needs to be a partial function we can later call because when calling
+# Dispatch directly from a different thread will result first a
+# "CoInitialize has not been called.", which can be fixed with
+# pythoncom.CoInitialize() but secondly "The application called an interface
+# that was marshalled for a different thread".
```
BigRoy: This comment seems a bit oddly phrased.

> This needs a partial function call because when calling Dispatch directly from a different thread will result in "CoInitialize has not been called" which can be fixed with pythoncom.CoInitialize(). However even then it will still error with "The application called an interface that was marshalled for a different thread"

Does that make it more comprehensible?

avalon/photoshop/lib.py
```diff
@@ -0,0 +1,255 @@
...
+
+    ref = Dispatch("Photoshop.ActionReference")
+    ref.PutClass(app().StringIDToTypeID("layerSection"))
+
+    lref = Dispatch("Photoshop.ActionReference")
+    lref.PutEnumerated(
+        app().CharIDToTypeID("Lyr "),
+        app().CharIDToTypeID("Ordn"),
+        app().CharIDToTypeID("Trgt")
+    )
```
BigRoy: Is it required to initiliaze the Dispatch() each time using `app()`? Or could you do it once and trigger all the commands afterwards?

```python
app = get_app()
app.CharIDToTypeId("Lyr ")
app.CharIDToTypeId("Ordn")
app.CharIDToTypeId("Trgt")
```

No idea how the interface works, but just wondering.

avalon/photoshop/lib.py
davidlatwe: sharp-eyed !

avalon/photoshop/README.md
tokejepsen: Nice catch!

avalon/photoshop/lib.py
tokejepsen: Nice catch!

avalon/photoshop/lib.py
tokejepsen: Yeah, makes more sense. Thanks!

avalon/photoshop/lib.py
tokejepsen: Yeah, we could certainly do that within functions.

avalon/photoshop/lib.py
```diff
@@ -0,0 +1,255 @@
...
+from ..tools import html_server
+
+# Convinience variable that could later query whether the platform is Windows
+# or Mac.
+# This needs to be a partial function we can later call because when calling
+# Dispatch directly from a different thread will result first a
+# "CoInitialize has not been called.", which can be fixed with
+# pythoncom.CoInitialize() but secondly "The application called an interface
+# that was marshalled for a different thread".
+app = functools.partial(Dispatch, "Photoshop.Application")
```
BigRoy: I seemed to have forgotten to actually submit this comment. Would this function also work if it's written as:
```python
def app():
    return Dispatch("Photoshop.Application")
```

In essence that should behave very similar to the `functools.partial` equivalent that you're using.
That would also make it much clearer that it's solely a convenience function, correct?

tokejepsen:
Sorry for all the commit spamming, but got the test to pass now.

davidlatwe:
Nice work !
I see you mentioned about `LayerSets` was not a good fit for being a container in the [comment](https://github.com/getavalon/core/issues/516#issuecomment-579232383) of #516, but what about `group` ? Current container is a single layer object right ? Will it able to load another PSD file while preserving layers ?

tokejepsen:
> but what about group ? 

What do you mean with `group`?
A `LayerSet` is a `group` in the javascript.

> Current container is a single layer object right ?

Yeah, currently a container is a single layer, but you could essentially containerise multiple layers with the same metadata.

> Will it able to load another PSD file while preserving layers ?

When using smart objects, the layers from a PSD gets flattened. Unless I'm missing a workflow?
Smart objects are the only way that I know of to reference media into Photoshop. There might be a method similar to our Nuke nodes referencing, but think that it'll be very complicated.

davidlatwe:
> A `LayerSet` is a `group` in the javascript.

Ah ! Did not know that, thought they were different things. Thanks !

> but you could essentially containerise multiple layers with the same metadata.

If that's the case, could `containerise` accept a list of layers instead of single layer ?

> When using smart objects, the layers from a PSD gets flattened.

So the `COMObject` which should be passed into `containerise` could be a smart object ?

> Unless I'm missing a workflow?

No no, I was just imaging what would I have to do to referencing another PSD file or other kind of subsets that may have multiple images or layers.

davidlatwe:
>> Unless I'm missing a workflow?
>
> No no, I was just imaging what would I have to do to referencing another PSD file or other kind of subsets that may have multiple images or layers.

But if publishing or loading that kind of subset in Photoshop would make things too complicated from actual production needs, then I can ignore that. :)

tokejepsen:
> If that's the case, could containerise accept a list of layers instead of single layer ?

Sure, that is possible. Though I thought for the sake of simplicity of the function, have a single layer as argument would be better? Can always containerise multiple times.
Or is your argument that all other hosts (except Fusion) containerises multiple objects?

> So the COMObject which should be passed into containerise could be a smart object ?

Yeah, could be any layer or `LayerSet` for that matter. In Photoshop layers are `LayerSet` and `ArtLayer` which translates to `group` and `layer` in the UI.
So if you figure out a way of read/loading multiple layers, you could just containerise a `LayerSet` (`group`) with all the layers in.

davidlatwe:
>> If that's the case, could containerise accept a list of layers instead of single layer ?
>
> Sure, that is possible. Though I thought for the sake of simplicity ... Or is your argument that all other hosts (except Fusion) containerises multiple objects?

After a good thought, I think it's not necessary for Photoshop. Plus, artist can only find stuff from that layer stack UI, so yeah, it's a good idea to keep things as simple as possible. 👍 

>> So the COMObject which should be passed into containerise could be a smart object ?
>
> Yeah, could be any layer or LayerSet for that matter.

Nice !

tokejepsen:
This PR is waiting on https://github.com/getavalon/core/pull/523 to be merged.

BigRoy:
The problem with #523 is still [this particular comment](https://github.com/getavalon/core/pull/523/files#r375174256). That should need to be tested and made sure that Blender will continue to work.

tokejepsen:
With https://github.com/getavalon/core/pull/524/commits/98503a61a06472e33142a070b23c4ac61f7aee51, this PR is ready to be merged.
Let me know what you think.

tokejepsen:
The #527 actually has a generic server which we could use for all Adobe products.

Lets leave this PR open for now, and I might rework this PR.

BigRoy:
@tokejepsen You mentioned potentially reworking this PR. Should we merge this as is and open an issue? Should we close this? I noticed Pype merged something like this. So this is usable? If so, is there any documentation? :)

tokejepsen:
The Pype have been using this in production, so there may be a couple of bugs from them. Otherwise, yeah, we could merge this.

The Pype team is also working on a second iteration using websocket and native javascript in PS for cross platform compatibility, so the question is whether we wanna wait for that.

@mkolar and @iLLiCiTiT can update about the production testing and second iteration.

tokejepsen:
This PR has been superseded by the Pype, so it better for the code to come from there.

PR #525, opened 2020-02-06
Optionally initialize Session['AVALON_HIERARCHY']

jasperges:
In [this PR](https://github.com/getavalon/core/pull/371) the concept of an extra hierarchy for assets was introduced. You can reference this in your file templates with `{hierarchy}`. The value is computed when you set the context to a different asset.
However if you start your DCC (only relying on Avalon env variables) and you don't switch to another asset, `Session['AVALON_HIERARCHY']` doesn't exist. That gives an error when you open the workfiles app and try to save your work (at least in the new [combined context/workfiles app](https://github.com/getavalon/core/pull/442).

**What's changed?**
If the environment variable 'AVALON_HIERARCHY' is present Session['AVALON_HIERARCHY'] will be initialized on start up. That way there is no error when saving your work file without switching context to another asset first.

### DIFF
--- a/avalon/io.py
+++ b/avalon/io.py
@@ -139,7 +139,7 @@ def _from_environment():
             ("AVALON_SCENEDIR", None),
 
             # Optional hierarchy for the current Asset. This can be referenced
-            # as `{hierarchy}` is your file templates.
+            # as `{hierarchy}` in your file templates.
             # This will be (re-)computed when you switch the context to another
             # asset. It is computed by checking asset['data']['parents'] and
             # joining those together with `os.path.sep`.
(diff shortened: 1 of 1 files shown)


avalon/io.py
```diff
@@ -138,6 +138,14 @@ def _from_environment():
             # Optional path to scenes directory (see Work Files API)
             ("AVALON_SCENEDIR", None),
 
+            # Optional hierarchy for the current Asset. This can be referenced
+            # as `{hierarchy}` is your file templates.
```
davidlatwe: Found a typo :P
`# as {hierarchy} is your file templates.`
Should be
`# as {hierarchy} i*n* your file templates.`

jasperges: Oh dear, so simple and still I need several commits... :flushed: :slightly_smiling_face:

PR #528, opened 2020-02-17
Style tweak, dimming highlight color

davidlatwe:
### What's changed
* Comment out `QHeaderView::section`'s color attribute so one could customize it
* Dimming selection highlight color so the icons could pop-up, having more contrast

#### Before
9986041b-fa52-44ba-bc62-fd64defeee5d

#### After
3c8e4214-919b-47ea-bfff-9fa2a790d7bf

#### Custom header section text color
image

### Motivation
Although I did think about overriding styles from the config, but then I thought it could become a bit harder to develop GUI in style that everyone could adopt since what other sees would be different from mine. So I decide to push the change I need and have a open discuss to see if everyone feels okay with it.

It's not much but I think it could improve the overall color contrast for seeing icons and other things that are a bit more important than selections. Bright icon friendly style. ☺️

### DIFF
--- a/avalon/style/style.qss
+++ b/avalon/style/style.qss
@@ -314,9 +314,9 @@ QWidget:disabled
 
 QAbstractItemView
 {
-    alternate-background-color: #3A3939;
+    alternate-background-color: #2D2C2C;
     color: silver;
-    border: 1px solid 3A3939;
+    border: 1px solid #2D2C2C;
     border-radius: 2px;
     padding: 1px;
 }
--- a/avalon/style/style.qss
+++ b/avalon/style/style.qss
@@ -1103,7 +1103,7 @@ QHeaderView
 
 QHeaderView::section  {
     background-color: #3A3939;
-    color: silver;
+    /*color: silver;*/
     padding: 4px;
     border: 1px solid #6c6c6c;
     border-radius: 0px;
--- a/avalon/style/style.qss
+++ b/avalon/style/style.qss
@@ -505,14 +505,6 @@ QPlainTextEdit
     border: 1px solid #3A3939;
 }
 
-QHeaderView::section
-{
-    background-color: #3A3939;
-    color: silver;
-    padding-left: 4px;
-    border: 1px solid #6c6c6c;
-}
-
 QSizeGrip {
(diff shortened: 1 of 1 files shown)


avalon/style/style.qss
```diff
@@ -4,6 +4,11 @@
     font-family: "Open Sans";
 }
 
+/*
+@ Selection highlight: #40576D
```
davidlatwe: This was like a quick access for testing color, since `qss` doesn't support declaring variables. Could be removed if you want it more cleaner.

avalon/style/style.qss
```diff
@@ -505,14 +510,6 @@ QPlainTextEdit
     border: 1px solid #3A3939;
 }
 
-QHeaderView::section
-{
-    background-color: #3A3939;
-    color: silver;
-    padding-left: 4px;
-    border: 1px solid #6c6c6c;
-}
```
davidlatwe: This section was duplicated, there are one exact definition with more attributes down below this `style.qss` file.

mottosso:
Nice one :)

BigRoy:
Looking good! At first it did give me the feeling that somehow it was "disabled" because of the dimmed color but likely it just needs some getting used to.

What does it look like when a row is disabled? Or the widget is disabled?

I always wondered whether we should make the Alternating Row color slightly subtler and easier to the eye? Any thoughts on that?

davidlatwe:
> What does it look like when a row is disabled? Or the widget is disabled?

Good question, here's my test, disabling row and widget 👇 

8612a1e2-4827-48bd-a130-5bdc90e46897

It seems disabling widgets doesn't affect the highlight color, and the same result before this PR.

> I always wondered whether we should make the Alternating Row color slightly subtler and easier to the eye?

I have thought about that, too. And I vote yes on that. 👍

davidlatwe:
A quick test on dimming alternative row color (`alternate-background-color`)

image

From `#3A3939` to `#2D2C2C`.

mkolar:
I like it. also with the dimming of the alternating colours. It's a bit softer which is always nice.

davidlatwe:
Can we merge this ? 😃

Issue #507, opened 2020-01-09
Use specific Creator Plugin when multiple Creators exist for single family


BigRoy:
### Issue

When writing two Creator plug-ins that are targeted to the same resulting family there is unexpected behavior when using it with the Creator tool.

Both entries will individually show in the Creator, however whenever you trigger either of them **both** will be created. This is because the [API triggers all Plugins based solely on the `family` matching, and it doesn't trigger a specific plugin](https://github.com/getavalon/core/blob/94aee92b1bbd8b0e54b47c484d0471cf411ce78a/avalon/pipeline.py#L882-L901)

I recall the API was designed for a reason but can't remember why. Now, with the Creator tool built as it is the the resulting behavior is super confusing. Is there *anyone* relying on this behavior of it to trigger multiple Creator plugins for a specific family?

```python
class CreateA(api.Creator):
    label = "A"
    family = "foobar"

    def process(self):
        print("A")

class CreateB(api.Creator):
    label = "B"
    family = "foobar"

    def process(self):
        print("B")
```

Now in the Creator tool create A or B, you will always see both plug-ins are triggered:
```
A
B
```

### Proposal

I'd propose to have the first argument for `api.Create` to be the `Creator` plugin to trigger as opposed to the family name.

For backwards compatibility I'd then make it so whenever it does *not* get an instance of `Creator` but a string that it will keep the old behavior. This way any custom code built against the API will remain to work as is.

Then we'll just update the Creator tool to not trigger by family name but explicitly pass it the Plugin.

tokejepsen:
Cant see why you'd need to have two plugins triggered. Voting for passing the `Creator` plugin with backwards compatibility.

mkolar:
I agree with Toke. And just as he, I can't for the life of me think of a situation where you want two creators. Do you have a use case?

ddesmond:
Had same behaviour when trying to do Imagesequence in Houdini, since i wanted RenderImageseqnce_MANTRA and RenderImageseqnce_ARNOLD. When created Houdini would spawn me both nodes. Case: Users want to be able to render localy some elements and not depend on deadline/farm.

davidlatwe:
Just adding notes.
* The `api.create` was originally from #291.
* Only one last `instance` is returned from `api.create` no matter how many plugin been processed.

BigRoy:
> The `api.create` was originally from #291.

[...]

PR #531, opened 2020-03-10
Make `api.create` able to run with one specific creator plugin

davidlatwe:
### Changes
This PR implements #507. Depends on the type of input argument `family` of `api.create`, try matching all plugins via `family` if it's a string, or dierctly run the given plugin if `family` is that plugin.

### DIFF
--- a/avalon/pipeline.py
+++ b/avalon/pipeline.py
@@ -878,12 +878,13 @@ def create(name, asset, family, options=None, data=None):
 
     host = registered_host()
 
-    plugins = list()
-    for Plugin in discover(Creator):
-        has_family = family == Plugin.family
+    if isinstance(family, str):
+        Plugins = [P for P in discover(Creator) if family == P.family]
+    else:
+        Plugins = [family]
 
-        if not has_family:
-            continue
+    plugins = list()
+    for Plugin in Plugins:
 
         Plugin.log.info(
             "Creating '%s' with '%s'" % (name, Plugin.__name__)
--- a/avalon/tools/creator/app.py
+++ b/avalon/tools/creator/app.py
@@ -436,13 +436,13 @@ class Window(QtWidgets.QDialog):
 
         subset_name = result.text()
         asset = asset.text()
-        family = item.data(FamilyRole)
+        Plugin = item.data(PluginRole)
         use_selection = self.data["Use Selection Checkbox"].isChecked()
 
         try:
             api.create(subset_name,
                        asset,
-                       family,
+                       Plugin,
                        options={"useSelection": use_selection})
 
         except NameError as e:


davidlatwe:
Here's my test code

```python
class CreateA(api.Creator):
    label = "A"
    family = "foobar"

    def process(self):
        print("A")

class CreateB(api.Creator):
    label = "B"
    family = "foobar"

    def process(self):
        print("B")
        

from avalon import api
from avalon.tools import creator
# Deregister all plugins for clarity
for path in api.registered_plugin_paths()[api.Creator]:
    api.deregister_plugin_path(api.Creator, path)

api.register_plugin(api.Creator, CreateA)
api.register_plugin(api.Creator, CreateB)

creator.show()
```

After this PR, only one creator obove will be triggered.

avalon/pipeline.py
```diff
@@ -878,12 +878,13 @@ def create(name, asset, family, options=None, data=None):
 
     host = registered_host()
 
-    plugins = list()
-    for Plugin in discover(Creator):
-        has_family = family == Plugin.family
+    if isinstance(family, str):
+        Plugins = [P for P in discover(Creator) if family == P.family]
+    else:
+        Plugins = [family]
```
BigRoy, line 884: This got me wondering for a second what is going on here. Should this maybe be in reverse, something like this:

```python
    if isinstance(family, Creator):
        # Allow passing in a specific Creator to run only that plug-in
        Plugins = [family]
    else:
        Plugins = [P for P in discover(Creator) if family == P.family]
```

Also makes the check more explicit to it being a Creator plug-in.

BigRoy (commented):
Solid work @davidlatwe - admittedly seeing `family` being used as plug-in directly without a clear commet regarding it got me confused for a second. Added a comment to maybe tweak that code layout a bit for readability. With that tweaked, let's merge this!

avalon/pipeline.py
davidlatwe: Good point !
Changed in commit ebcb78c
( note that the `family` has now changed to also accept a subclass of `api.Creator`, so it should be using `issubclass` ;) )

davidlatwe:
Merging this today :D

avalon/pipeline.py
BigRoy: Well spotted!

BigRoy:
@davidlatwe I'm wondering. Since you've a ready to go test for this? Any chance you can add a test too for the unittests?

One that tests the old by string family behavior, and the new behavior?

It would be in `core/avalon/tests` then.

davidlatwe:
Just added, also fixed an obvious bug that I missed in ebcb78c. :P

davidlatwe:
Weird, the test actually failed, but build passed. :/

The test passed in my Maya session, maybe I should write the test plugin into an actual script file.

davidlatwe:
Ah, now I see the problem, the `host` registed in `test_pipeline` has no attribute 'maintained_selection'
```
avalon.pipeline: WARNING: 'module' object has no attribute 'maintained_selection'
```

davidlatwe:
Test fixed ! See [log](https://travis-ci.org/github/getavalon/core/builds/731580482#L507)

BigRoy:
Awesome work! 

The only thing you are missing in the test is the test case where two plug-ins register to the same family where running it by family string name it should run both (as originally) and by solely Plugin run only the single one. That's the test case we want to preserve over time I suppose (or at least know about that it changes).

```python
# psuedocode

data= {"value": 0}

class CreateA(api.Creator):
    label = "A"
    family = "foobar"

    def process(self):
        data["value"] += 1

class CreateB(api.Creator):
    label = "B"
    family = "foobar"

    def process(self):
        data["value"] += 10
        

from avalon import api
from avalon.tools import creator
# Deregister all plugins for clarity
for path in api.registered_plugin_paths()[api.Creator]:
    api.deregister_plugin_path(api.Creator, path)

api.register_plugin(api.Creator, CreateA)
api.register_plugin(api.Creator, CreateB)

api.create("foo", "my_asset", family="foobar")
assert data["value"] == 11, "Must run both Creator Plugins"

api.create("foo", "my_asset", family=CreateA)
assert data["value"] == 12, "Must run onlyCreatorA Plugin"
```
Following counting-like logic from pyblish tests, [example](https://github.com/pyblish/pyblish-base/blob/3290fd5b51b496d431cc2ca608639e21e727ccbd/tests/test_logic.py#L24-L67).

davidlatwe:
That indeed much better ! Thanks. 🚀

PR #535, opened 2020-03-17
Make Silos optional

mkolar:
## Problem
The concept of silos becomes problematic as projects grow larger and longer. The problem is visual where it's uncomfortable to browse through them with tabs, but also it's a bit limiting in terms of project structure.

## Suggested solution

We suggest dropping them completely and instead consider any asset witch has `visualParent` set to `Null` as highest hierarchy member to be shown directly under the project in the GUIs. We've run this at all our sites for around 6 months now, so the key question is backwards compatibility, which I hope we achieved.

This is what the loader for example looks like without the silo tabs.
image

------
This is a revival of an old PR #372 that was closed after we discovered more caveats to this. General consensus seemed to be towards this change though so hopefully this will work for everyone too.

If someone could give this a shot it would help a lot. Unfortunately, our setups have quite some dependency on Pype and it's hard to be sure that the backwards compatibility doesn't catch on something somewhere on projects of others.

### DIFF
--- a/avalon/io.py
+++ b/avalon/io.py
@@ -199,7 +199,7 @@ def _from_environment():
         ) if os.getenv(item[0], item[1]) is not None
     }
 
-    session["schema"] = "avalon-core:session-1.0"
+    session["schema"] = "avalon-core:session-2.0"
     try:
         schema.validate(session)
     except schema.ValidationError as e:
--- a/avalon/tools/lib.py
+++ b/avalon/tools/lib.py
@@ -108,6 +108,75 @@ def iter_model_rows(model,
 
 
 @contextlib.contextmanager
+def preserve_states(
+    tree_view, column=0, role=None,
+    preserve_expanded=True, preserve_selection=True,
+    expanded_role=QtCore.Qt.DisplayRole, selection_role=QtCore.Qt.DisplayRole
+
+):
+    """Preserves row selection in QTreeView by column's data role.
+    This function is created to maintain the selection status of
+    the model items. When refresh is triggered the items which are expanded
+    will stay expanded and vise versa.
+        tree_view (QWidgets.QTreeView): the tree view nested in the application
+        column (int): the column to retrieve the data from
+        role (int): the role which dictates what will be returned
+    Returns:
+        None
+    """
+    # When `role` is set then override both expanded and selection roles
+    if role:
+        expanded_role = role
+        selection_role = role
+
+    model = tree_view.model()
+    selection_model = tree_view.selectionModel()
+    flags = selection_model.Select | selection_model.Rows
+
+    expanded = set()
+
+    if preserve_expanded:
+        for index in iter_model_rows(
+            model, column=column, include_root=False
+        ):
+            if tree_view.isExpanded(index):
+                value = index.data(expanded_role)
+                expanded.add(value)
+
+    selected = None
+
+    if preserve_selection:
+        selected_rows = selection_model.selectedRows()
+        if selected_rows:
+            selected = set(row.data(selection_role) for row in selected_rows)
+
+    try:
+        yield
+    finally:
+        if expanded:
+            for index in iter_model_rows(
+                model, column=0, include_root=False
+            ):
+                value = index.data(expanded_role)
+                is_expanded = value in expanded
+                # skip if new index was created meanwhile
+                if is_expanded is None:
+                    continue
+                tree_view.setExpanded(index, is_expanded)
+
+        if selected:
+            # Go through all indices, select the ones with similar data
+            for index in iter_model_rows(
+                model, column=column, include_root=False
+            ):
+                value = index.data(selection_role)
+                state = value in selected
+                if state:
+                    tree_view.scrollTo(index)  # Ensure item is visible
+                    selection_model.select(index, flags)
+
+
+@contextlib.contextmanager
 def preserve_expanded_rows(tree_view,
                            column=0,
                            role=QtCore.Qt.DisplayRole):
--- a/avalon/tools/lib.py
+++ b/avalon/tools/lib.py
@@ -452,3 +452,19 @@ def get_active_group_config(asset_id, include_predefined=False):
         })
 
     return ordered
+
+
+def project_use_silo(project_doc):
+    """Check if templates of project document contain `{silo}`.
+
+    Args:
+        project_doc (dict): Project document queried from database.
+
+    Returns:
+        bool: True if any project's template contain "{silo}".
+    """
+    templates = project_doc["config"].get("template") or {}
+    for template in templates.values():
+        if "{silo}" in template:
+            return True
+    return False
--- a/avalon/tools/lib.py
+++ b/avalon/tools/lib.py
@@ -108,12 +108,13 @@ def iter_model_rows(model,
 
 
 @contextlib.contextmanager
-def preserve_states(
-    tree_view, column=0, role=None,
-    preserve_expanded=True, preserve_selection=True,
-    expanded_role=QtCore.Qt.DisplayRole, selection_role=QtCore.Qt.DisplayRole
-
-):
+def preserve_states(tree_view,
+                    column=0,
+                    role=None,
+                    preserve_expanded=True,
+                    preserve_selection=True,
+                    expanded_role=QtCore.Qt.DisplayRole,
+                    selection_role=QtCore.Qt.DisplayRole):
     """Preserves row selection in QTreeView by column's data role.
     This function is created to maintain the selection status of
     the model items. When refresh is triggered the items which are expanded
--- a/avalon/tools/loader/app.py
+++ b/avalon/tools/loader/app.py
@@ -39,9 +39,9 @@ class Window(QtWidgets.QDialog):
 
         container = QtWidgets.QWidget()
 
-        assets = AssetWidget(parent=self)
+        assets = AssetWidget()
         families = FamilyListWidget()
-        subsets = SubsetWidget(parent=self)
+        subsets = SubsetWidget()
         version = VersionWidget()
 
         # Create splitter to show / hide family filters
@@ -425,8 +425,8 @@ def show(debug=False, parent=None, use_context=False):
         lib.refresh_group_config_cache()
 
         window = Window(parent)
-        window.setStyleSheet(style.load_stylesheet())
         window.show()
+        window.setStyleSheet(style.load_stylesheet())
 
         if use_context:
             context = {"asset": api.Session["AVALON_ASSET"]}
--- a/avalon/tools/loader/app.py
+++ b/avalon/tools/loader/app.py
@@ -408,7 +408,7 @@ def show(debug=False, parent=None, use_context=False):
             module.window.refresh()
             return
         except RuntimeError as exc:
-            if not exc.message.rstrip().endswith("already deleted."):
+            if not str(exc).rstrip().endswith("already deleted."):
                 raise
 
             # Garbage collected
--- a/avalon/tools/loader/app.py
+++ b/avalon/tools/loader/app.py
@@ -168,19 +168,29 @@ class Window(QtWidgets.QDialog):
 
         t1 = time.time()
 
-        asset_item = assets_model.get_active_index()
-        if asset_item is None or not asset_item.isValid():
-            return
+        asset_item = assets_model.get_active_asset()
+        if asset_item is None:
+            document_id = None
+            document_name = None
+            document_silo = None
+        elif asset_item["type"] == "silo":
+            document_id = None
+            document_name = None
+            document_silo = asset_item["name"]
+        else:
+            document = asset_item["_document"]
+            document_id = document["_id"]
+            document_name = document["name"]
+            document_silo = document.get("silo")
 
-        document = asset_item.data(DocumentRole)
-        subsets_model.set_asset(document["_id"])
+        subsets_model.set_asset(document_id)
 
         # Clear the version information on asset change
         self.data["model"]["version"].set_version(None)
 
-        self.data["state"]["context"]["asset"] = document["name"]
-        self.data["state"]["context"]["assetId"] = document["_id"]
-        self.data["state"]["context"]["silo"] = document.get("silo")
+        self.data["state"]["context"]["asset"] = document_name
+        self.data["state"]["context"]["assetId"] = document_id
+        self.data["state"]["context"]["silo"] = document_silo
         self.echo("Duration: %.3fs" % (time.time() - t1))
 
     def _versionschanged(self):
--- a/avalon/tools/loader/app.py
+++ b/avalon/tools/loader/app.py
@@ -24,8 +24,10 @@ class Window(QtWidgets.QDialog):
         super(Window, self).__init__(parent)
         self.setWindowTitle(
             "Asset Loader 2.1 - %s/%s" % (
-                api.registered_root(),
-                api.Session.get("AVALON_PROJECT")))
+                api.registered_root().replace("\\", "/"),
+                api.Session.get("AVALON_PROJECT")
+            )
+        )
 
         # Enable minimize and maximize for app
         self.setWindowFlags(QtCore.Qt.Window)
@@ -37,9 +39,9 @@ class Window(QtWidgets.QDialog):
 
         container = QtWidgets.QWidget()
 
-        assets = AssetWidget(silo_creatable=False)
+        assets = AssetWidget(parent=self)
         families = FamilyListWidget()
-        subsets = SubsetWidget()
+        subsets = SubsetWidget(parent=self)
         version = VersionWidget()
 
         # Create splitter to show / hide family filters
@@ -178,7 +180,7 @@ class Window(QtWidgets.QDialog):
 
         self.data["state"]["context"]["asset"] = document["name"]
         self.data["state"]["context"]["assetId"] = document["_id"]
-        self.data["state"]["context"]["silo"] = document["silo"]
+        self.data["state"]["context"]["silo"] = document.get("silo")
         self.echo("Duration: %.3fs" % (time.time() - t1))
 
     def _versionschanged(self):
@@ -217,10 +219,6 @@ class Window(QtWidgets.QDialog):
 
         """
 
-        silo = context.get("silo", None)
-        if silo is None:
-            return
-
         asset = context.get("asset", None)
         if asset is None:
             return
@@ -235,8 +233,7 @@ class Window(QtWidgets.QDialog):
             self._refresh()
 
         asset_widget = self.data["model"]["assets"]
-        asset_widget.set_silo(silo)
-        asset_widget.select_assets([asset], expand=True)
+        asset_widget.select_assets(asset)
 
     def echo(self, message):
         widget = self.data["label"]["message"]
@@ -400,8 +397,8 @@ def show(debug=False, parent=None, use_context=False):
             module.window.activateWindow()     # for Windows
             module.window.refresh()
             return
-        except RuntimeError as e:
-            if not str(e).rstrip().endswith("already deleted."):
+        except RuntimeError as exc:
+            if not exc.message.rstrip().endswith("already deleted."):
                 raise
 
             # Garbage collected
@@ -418,12 +415,11 @@ def show(debug=False, parent=None, use_context=False):
         lib.refresh_group_config_cache()
 
         window = Window(parent)
-        window.show()
         window.setStyleSheet(style.load_stylesheet())
+        window.show()
 
         if use_context:
-            context = {"asset": api.Session["AVALON_ASSET"],
-                       "silo": api.Session["AVALON_SILO"]}
+            context = {"asset": api.Session["AVALON_ASSET"]}
             window.set_context(context, refresh=True)
         else:
             window.refresh()
--- a/avalon/tools/loader/model.py
+++ b/avalon/tools/loader/model.py
@@ -17,15 +17,18 @@ def is_filtering_recursible():
 
 
 class SubsetsModel(TreeModel):
-    Columns = ["subset",
-               "family",
-               "version",
-               "time",
-               "author",
-               "frames",
-               "duration",
-               "handles",
-               "step"]
+    Columns = [
+        "subset",
+        "asset",
+        "family",
+        "version",
+        "time",
+        "author",
+        "frames",
+        "duration",
+        "handles",
+        "step"
+    ]
 
     SortAscendingRole = QtCore.Qt.UserRole + 2
     SortDescendingRole = QtCore.Qt.UserRole + 3
@@ -35,8 +38,9 @@ class SubsetsModel(TreeModel):
         self._asset_id = None
         self._sorter = None
         self._grouping = grouping
-        self._icons = {"subset": qtawesome.icon("fa.file-o",
-                                                color=style.colors.default)}
+        self._icons = {
+            "subset": qtawesome.icon("fa.file-o", color=style.colors.default)
+        }
 
     def set_asset(self, asset_id):
         self._asset_id = asset_id
@@ -50,7 +54,7 @@ class SubsetsModel(TreeModel):
 
         # Trigger additional edit when `version` column changed
         # because it also updates the information in other columns
-        if index.column() == 2:
+        if index.column() == self.Columns.index("version"):
             item = index.internalPointer()
             parent = item["_id"]
             version = io.find_one({"name": value,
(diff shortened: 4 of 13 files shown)


davidlatwe:
I like this change !

> The concept of silos becomes problematic as projects grow larger and longer.

This also emerged in a couple of big projects here.

> so the key question is backwards compatibility, which I hope we achieved.

At first glance, I think you guys nailed it, I quickly checkout this branch and looks great !
The silo icon is different than yours, but I think what I see here is the correct one.

image

I have done some model publishes in both new and old assets, seems working just fine. :)

One problem though, Loader's subset tree gets reseted when the refresh button is pressed, the only way to get that tree updated is switching to other asset and back. Any idea ?

davidlatwe:
Oh and, I haven't review all the code, need more time to do that. ☺️

iLLiCiTiT:
> The silo icon is different than yours, but I think what I see here is the correct one.

The "database" icon is for "silo" which we don't use at all. We use only `visualParent` hierarchy.

> One problem though, Loader's subset tree gets reseted when the refresh button is pressed, the only way to get that tree updated is switching to other asset and back. Any idea ?

You mean that there is missing refresh button for subsets tree, because when you hit refresh button in assets tree, you'll get reseted also assets?
- If yes then we agreed that resetting asset tree is just one call to database, which is not so heavy and has minimal performance affect, so why to cruble gui with more buttons with ephemeral advantage. But can be changed if you don't agree :)

avalon/tools/loader/app.py
```diff
@@ -400,8 +407,8 @@ def show(debug=False, parent=None, use_context=False):
             module.window.activateWindow()     # for Windows
             module.window.refresh()
             return
-        except RuntimeError as e:
-            if not str(e).rstrip().endswith("already deleted."):
+        except RuntimeError as exc:
+            if not exc.message.rstrip().endswith("already deleted."):
```
BigRoy: `Exception.message` does not seem to exist in Python 3.6

```python
>>> try:
...     raise RuntimeError("TEST")
... except RuntimeError as exc:
...     print(exc.message)
...
  File "<stdin>", line 2, in <module>
RuntimeError: TEST

During handling of the above exception, another exception occurred:

  File "<stdin>", line 4, in <module>
AttributeError: 'RuntimeError' object has no attribute 'message'
```

This doesn't seem Python 3 compatible. See [PEP-0352](https://www.python.org/dev/peps/pep-0352/).

> The 'message' attribute is also deprecated.

avalon/tools/loader/app.py
```diff
@@ -418,12 +425,11 @@ def show(debug=False, parent=None, use_context=False):
         lib.refresh_group_config_cache()
 
         window = Window(parent)
-        window.show()
         window.setStyleSheet(style.load_stylesheet())
+        window.show()
```
BigRoy: Is there a reason for these lines to be swapped? Double check that. Because some of these lines have swapped before I believe to fix stylesheet issues in specific hosts. Make sure they are the correct order, maybe check "blame/history" to see what's intended.

avalon/tools/loader/model.py
```diff
@@ -17,15 +17,18 @@ def is_filtering_recursible():
...
-               "version",
-               "time",
-               "author",
-               "frames",
-               "duration",
-               "handles",
-               "step"]
+    Columns = [
+        "subset",
+        "asset",
```
BigRoy: This seems new too. Any reason for having this in this particular PR?

avalon/tools/loader/model.py
```diff
@@ -206,23 +210,23 @@ def data(self, index, role):
             return
 
         if role == QtCore.Qt.DisplayRole:
-            if index.column() == 1:
+            if index.column() == self.Columns.index("family"):
```
BigRoy: Be aware that `data()` is called very very very often. `Column.index()` on a list is relatively slow. I like the increased readability but maybe it's worth it to have a lookup table that is initialized on `__init__` only once?

```python
self.columns_index = dict((key, i) for i in enumerate(self.Columns))
```

Then this line just becomes:
```python
self.columns_index["family"]
```

---

Of course this would be for all of the `self.Columns.index()` lines.

avalon/tools/models.py
```diff
@@ -213,19 +213,34 @@ def _get_task_icons(self):
                                       color=style.colors.default)
                 self._icons[task["name"]] = icon
 
-    def set_assets(self, asset_ids):
+    def set_assets(self, asset_ids=[], asset_entities=None):
         """Set assets to track by their database id
 
         Arguments:
             asset_ids (list): List of asset ids.
+            asset_entities (list): List of asset entities from MongoDB.
```
BigRoy: The extra argument seems a bit odd. It seems explicit but also like an odd choice.

Would it make more sense to do Python duck-typing. E.g. check if it's a `dict` when provided in `asset_ids` and then assume it's an `asset_entity`. Also, should the terminology be `asset document`? I believe MongoDB refers to Documents over Entities.

Not saying it's better. I'm just wondering what others think.

avalon/tools/models.py
```diff
@@ -336,6 +341,19 @@ def _add_hierarchy(self, assets, parent=None):
             None
 
         """
+        if silos:
+            # WARNING: Silo item "_id" is set to silo value
+            # mainly because GUI issue with perserve selection and expanded row
```
BigRoy: 1. `perserve` -> `preserve`

2. What exactly is the issue described here? What in particular caused that issue? Is that the case when the "silo" is *automagically* added based on an asset's "data.silo" value and as such it would have the same `_id` as the asset? Either way, good to see this clearly "warned" in a comment.

avalon/tools/projectmanager/app.py
```diff
@@ -232,9 +244,9 @@ def show(root=None, debug=False, parent=None):
         io.install()
 
     with tools_lib.application():
-        window = Window(parent)
-        window.show()
+        window = Window(parent=parent)
         window.setStyleSheet(style.load_stylesheet())
+        window.show()
```
BigRoy: Same comment on ordering of show and stylesheets as before.

avalon/tools/widgets.py
```diff
@@ -61,64 +47,50 @@ def __init__(self, silo_creatable=True, parent=None):
...
-        self.silo.set_silos(silos)
-
         self._refresh_model()
 
-    def get_current_silo(self):
-        """Returns the currently active silo."""
-        return self.silo.get_current_silo()
-
     def get_active_asset(self):
         """Return the asset id the current asset."""
```
BigRoy: Is this still returning the `id` of the document with the changed role? It seems like this docstring might be wrong currently.

avalon/tools/widgets.py
```diff
@@ -129,204 +101,68 @@ def get_active_index(self):
...
 
         Returns:
             None
 
+        Default `key` is "name" in that case `assets` should contain single
+        asset name or list of asset names. (It is good idea to use "_id" key
+        instead of name in that case `assets` must contain `ObjectId` object/s)
+        It is expected that each value in `assets` will be found only once.
+        If the filters according to the `key` and `assets` correspond to
+        the more asset, only the first found will be selected.
```
BigRoy: Seems to have some odd grammar, maybe this is better?
```
        Default `key` is "name" in that case `assets` should contain single
        asset name or list of asset names. It is recommended to use "_id" key
        instead of name. In that case `assets` must contain `ObjectId` objects.
        It is assumed that each value in `assets` is found only once.
        On multiple matches only the first found will be selected.
```

avalon/tools/widgets.py
```diff
@@ -129,204 +101,68 @@ def get_active_index(self):
...
         """
         # TODO: Instead of individual selection optimize for many assets
 
-        assert isinstance(assets,
-                          (tuple, list)), "Assets must be list or tuple"
+        if not isinstance(assets, (tuple, list)):
+            assets = [assets]
+
+        # convert to list - tuple cant be modified
+        assets = list(assets)
```
BigRoy: Looking at the usage of `assets` variable below, would this make more sense to be a `set` for quick lookup?
You might be looking over many many rows for many assets, so it could turn out a slow loop.

It seems `ObjectId` also works inside a `set` so it should be good.

avalon/tools/workfiles/app.py
```diff
@@ -295,11 +295,12 @@ def __init__(self):
...
 
-    def set_asset(self, asset_id):
+    def set_asset(self, asset):
 
-        if asset_id is None:
+        assets = []
+        if asset is not None:
             # Asset deselected
-            return
+            assets.append(asset)
```
BigRoy: What exactly changed that required this check to be the opposite and have no return on `None`?
Is it required to run through to reset something else?

Would this be more readable as the following:
```python
assets = [asset] if asset else []
```

Also the `# Asset deselected` comment seems to now be wrong with the change of the condition too.

avalon/tools/workfiles/app.py
```diff
@@ -748,7 +749,7 @@ def __init__(self, parent=None):
         widgets = {
             "pages": QtWidgets.QStackedWidget(),
             "body": QtWidgets.QWidget(),
-            "assets": AssetWidget(silo_creatable=False),
+            "assets": AssetWidget(parent=self),
```
BigRoy: As asked before - what's changed that now we need to pass the parent?

avalon/tools/workfiles/app.py
iLLiCiTiT: Comment will be removed, good spot, and the condition is from future PR, can change it as you wrote...

avalon/tools/loader/app.py
iLLiCiTiT: There is reason. PySide / PySide2 / PyQt4 / PyQt5, each has different approach to propagation of stylesheets. One have issues when stylesheets are applied after show and other have issues if they are applied before show. Issue is not with widget itself but to it's children widgets (If they don't have set parent). With setting stylesheets before show and adding `parent` to initialization of widgets was only working way cross all Python's Qt frameworks.

- You're right that each widget should have set parent to make it 100% working.
- @davidlatwe it was your commit which changed the order of these lines. Can you check this works now in Houdini? To be at least 99% sure this works.

avalon/tools/models.py
iLLiCiTiT: With `asset_document` I'm ok to change it and I agree it should be document.

The `asset_ids` was kept due to kind of backwards compatibility, I can imagine somebody has it's own tool using this widget. But if you don't agree, it's easy to remove whole part with `asset_ids`.

avalon/tools/loader/model.py
iLLiCiTiT: No, this is from future PR...

avalon/tools/models.py
iLLiCiTiT: 1.) I will never be a copywriter :(

2.) It is because silo can be used as parent in `assets` argument for assets with `visualParent` set to None.
And `preserve_states` may use `ObjectIdRole` which means it is better to have set `_id` to any other value then `None` (because each silo item would have _id value set to None). Makes that sence?

avalon/tools/workfiles/app.py
iLLiCiTiT: Same answer as before, but not sure about effect here, because I prepared this branch just in time workfiles + context manager PR was merged, so I didn't have time to verify if this solved stylesheet issue.
Now I can't verify, because home-office means no applications for me.

As I look into code now, I'm 90% sure it is not right... But it seems impossible to solve parenting here due to new widgets storing way.
```
pages = {
    "home": QtWidgets.QWidget()
}

widgets = {
    "pages": QtWidgets.QStackedWidget(), # parent=pages ["home"]
    "body": QtWidgets.QWidget(),         # parent=widgets["pages"]
    "assets": AssetWidget(parent=self),  # parent=widgets["body"]
    "tasks": TasksWidget(),              # parent=widgets["body"]
    "files": FilesWidget()               # parent=widgets["body"]
}
```

avalon/tools/widgets.py
```diff
@@ -153,7 +152,7 @@ def select_assets(self, assets, expand=True, key="name"):
                 continue
 
             # Remove processed asset
-            assets.pop(assets.index(value))
+            assets.discard(assets.index(value))
```
BigRoy: This should be: `assets.discard(value)`

avalon/tools/workfiles/app.py
mottosso: Are these widgets not part of a layout?

The only time you need to explicitly provide a parent is when it's a standalone widget. Whenever you call `layout.addWidget(widget)` then `widget` is assigned the layout as a parent, which itself is child to another widget, most likely the window.

If not, you can also call [`widget.setParent`](https://doc.qt.io/qt-5/qwidget.html#setParent) after instantiation.

iLLiCiTiT: Not really, mainly PySide have issues with parenting and propagation of styles when parent is not explicitly set and should be propagated internally. And other (I think it was PyQt4, not sure) had issues when stylesheet is set after `show` without understandable reason, stylesheets for specific `objectName` worked, but not for it's children.
e.g. `#MainWidget QLabel {backgroud: red;}` had no effect for Qlabels in `MainWidget`.

You got true, `setParent` should help here.

iLLiCiTiT:
We've realized that parenting and stylesheet issue should be resolved in new PR. So I've move stylesheet setters where they were and removed parenting.

mkolar:
I believe all the comments we resolved by this last commit now. Could do you guys think @BigRoy @davidlatwe ? Can we move forward with this?

All our other PRs are quite dependent on this one, so we'll wait with further work until this is approved and merged.

avalon/io.py
```diff
@@ -199,7 +199,7 @@ def _from_environment():
         ) if os.getenv(item[0], item[1]) is not None
     }
 
-    session["schema"] = "avalon-core:session-1.0"
+    session["schema"] = "avalon-core:session-2.0"
```
mottosso: Just to confirm, can someone confirm this is backwards compatible? We aren't asserting any particular version anywhere I think?

avalon/tools/lib.py
```diff
@@ -107,6 +107,75 @@ def iter_model_rows(model,
         yield index
 
 
+@contextlib.contextmanager
+def preserve_states(
```
mottosso: This formatting, I see it a lot lately. Is this pycharm or black or something? Let's stick with having the first argument on the same line as usual, and every other argument on a new line with the same indentation.

mottosso:
Style wise I think this looks good, and the PR is well focused and to the point, thanks for that. I'll leave it to David and Roy to judge the functionality.

BigRoy:
First off, great work on this PR. I pulled it and it instantly seemed to work on our existing projects with backwards compatibility. Good stuff.

I do have some additional notes and issues after testing it - some are related to the new style projects (and some are notes for others trying to adopt on how they might need to tweak their studio config):

- Whenever you work in a project that had silos in use then the Asset Create dialog will open up with the "silo" text field added which needs to be filled in manually. However, when trying to add a child asset underneath a parent this gets extra confusing since that should just inherit the "silo" from the parent (if it has any). Technically otherwise I could also add an asset underneath silo `asset` like `characters/hero` and set its hero silo to: `film`, a world of ambiguity and confusion will open up. What about disallowing the setting of "silo" if it has a parent AND if it has a parent with silo to inherit that silo. That way it should be fully backwards compatible on that front.
- Clicking refresh on the Asset Widget in the Loader does seem to reselect the original asset that I had selected but it seems to deactivate and clear the Subset/Publish Widget. Shouldn't that reselect/refresh also reshow the publishes for the current asset. It's a bit confusing that it does keep the asset selected but doesn't keep its publishes on the right hand side. (Not sure if this is a new bug) This same issue is apparent when the Loader was opened previously and you reopen it. It uses the same Loader application that was still in memory and forces a refresh, it reselects the asset on left hand side but show no published subsets on right hand set. Only clicking into another asset and back into the current asset refreshes it manually.

- On creating a new project by using an empty older project template I ran into this error on creating a new asset.:

afbeelding

Did I do something wrong or should I be setting something specific on my project to make this work? This seems to be an issue with the Project managers `avalon.tools.projectmanager.lib.create_asset()` which is trying to use the old 2.0 asset schema. Should this be updated to 3.0 when silo is NOT required? Also the docstring for that function seems outdated since it states it always requires silo. Anyway, changing that to asset-3.0 allowed me to create assets.

- Then the next problem is that the [Avalon Launcher](https://github.com/getavalon/launcher/pull/22) is unable to show these assets that have no silo. It will just [list an empty project due to it explicitly listing silos as opposed to top-level assets](https://github.com/getavalon/launcher/blob/1cec0a92e6af6e627072486d3f38bd53786981b3/launcher/control.py#L291-L298). Unfortunately a backwards compatible PR would then also need to be made there I suppose.

- With a little hack I then "went into that project" inside the avalon pipeline without the launcher. However, then I noticed the Work Files tool started messing up things on switching between assets:
```python
# Traceback (most recent call last):
#   File "D:\git_pipeline\avalon\development\git\core\avalon\tools\workfiles\app.py", line 631, in on_save_as_pressed
#     self._enter_session()   # Make sure we are in the right session
#   File "D:\git_pipeline\avalon\development\git\core\avalon\tools\workfiles\app.py", line 486, in _enter_session
#     api.update_current_task(asset=self._asset, task=self._task)
#   File "D:\git_pipeline\avalon\development\git\core\avalon\pipeline.py", line 1057, in update_current_task
#     os.environ.update(changes)
#   File "C:\Program Files\Autodesk\Maya2020\bin\python27.zip\os.py", line 461, in update
#   File "C:\Program Files\Autodesk\Maya2020\bin\python27.zip\os.py", line 422, in __setitem__
# TypeError: must be string, not None
```
I believe that might be due to [this](https://github.com/getavalon/core/blob/473dbdeea60f293bc0f7e292dcb4c09b960c5c2d/avalon/pipeline.py#L996) setting it to `None` and then that being passed to `os.environ.update()` later down the line. Potentially that should install fall back to an empty string `.get("silo", "")`? That seemed to do the trick for me. Or potentially even better set it to `asset_document.get("silo") or ""` so that if the database returned a value of "None" stored for `silo` that it wouldn't run into the same issue.
- Then I noticed that our project template included `{silo}` to decide where to save its assets. As opposed to giving an error on "requiring silo" in the project template it just created the asset without it.. this resulted in a path on disk like `{project}/{silo}/{asset}` turning into `test/None/foobar` on disk which is both unexpected and unwanted. I would stress that it should force the project to "require silo" of the project template uses it. Yes? --> **However, after a full restart of Maya** it instead resulted in a path like: `P:\Projects/test2//dev_test/work/model/maya` due to `{silo}` turning into an empty string. I'd then still argue that it should error on the template requiring silo to be a non-empty value.

- Then of course I noticed some of our publishing plug-ins in our config required the `silo` data and information, including our integrator. Updating that to `asset.get("silo") showed that `None` is not a valid value for silo anyway. In particular on the Representation schema in its context data.
```python
  File "D:\git_pipeline\avalon\development\git\pyblish-base\pyblish\plugin.py", line 517, in __explicit_process
    runner(*args)
  File "D:\git_pipeline\avalon\development\git\colorbleed-config\colorbleed\plugins\global\publish\integrate.py", line55, in process
  File "D:\git_pipeline\avalon\development\git\colorbleed-config\colorbleed\lib.py", line 433, in process
    self.register(instance)
  File "D:\git_pipeline\avalon\development\git\colorbleed-config\colorbleed\lib.py", line 483, in register
    version=version
  File "D:\git_pipeline\avalon\development\git\colorbleed-config\colorbleed\lib.py", line 739, in create_representations
    schema.validate(representation)
  File "D:\git_pipeline\avalon\development\git\core\avalon\schema.py", line 59, in validate
    resolver=resolver)
  File "D:\git_pipeline\avalon\development\git\core\avalon\vendor\jsonschema\validators.py", line 428, in validate
    cls(schema, *args, **kwargs).validate(instance)
  File "D:\git_pipeline\avalon\development\git\core\avalon\vendor\jsonschema\validators.py", line 117, in validate
    raise error
ValidationError: None is not of type u'string'

Failed validating u'type' in schema[u'properties'][u'context'][u'properties'][u'silo']:
    {u'type': u'string'}

On instance[u'context'][u'silo']:
    None
```
This last one was resolved by only inlcuding that data when silo was set:
```python
            silo = asset.get("silo")
            if silo:
                representation["context"]["silo"] = silo
```

---

Sorry for the extensive block of text. I just wanted to make sure I accurately wrote down the issues as how they occurred to me.

Unfortunately in its current state this PR is unusable in production for us due to the issues described above.

mkolar:
> Sorry for the extensive block of text. I just wanted to make sure I accurately wrote down the issues as how they occurred to me.

no worries, this is exactly the testing we need. All our projects either don't use silo, or are temporarily created without it, but it's really hard to catch there errors without testing on existing project with silo, so thanks for that!

We'll go through it and get back here, but it all seems very fixable.

davidlatwe:
It looks like most of the issues @BigRoy mentioned have been resolved ! 🎉 

But I think instead of querying database with `io.distinct.("silo")` to define the project is silo required or not, how about just looking into project's publish path template to see if keyword `{silo}` exists ?

iLLiCiTiT:
> how about just looking into project's publish path template to see if keyword {silo} exists ?

I like it. We can't use that because we don't set templates in project's document, but that's our issue we can easily solve. What others think about it?

davidlatwe:
Also, should we make Project Manager App to add `["data"]["parents"]` field into asset document while creating asset in project that has `silo` opt-out ? (to adopt `"AVALON_HIERARCHY"`)

One additional question, when using `AVALON_HIERARCY`, does that produce a folder structure that `work` and `publish` dirs may live with other child asset folder ? Like:
```
    characters
     └ boys
        ├ boyA
        │   ├ publish
        │   └ work  # Working on Modeling, Rig, LookDev
        ├ boyB
        │   ├ publish
        │   └ work  # Working on Modeling, Rig, LookDev
        ├ publish
        └ work  # Working on LookDevs or Crowd Assets

```

mkolar:
>One additional question, when using AVALON_HIERARCY, does that produce a folder structure that work and publish dirs may live with other child asset folder ? Like:

yes that is possible if you have task on `boys` you'll get work and publish folder there. We actually use it for editorials and similar types of work.

avalon/tools/lib.py
iLLiCiTiT: > Is this pycharm or black or something?

This is actually Me :) . I personally don't like that strange offset when first argument is on the same line. It's hard to read and edit for me and in this case it would cause PEP 8 E501. But I agree there should be one argument per line.

iLLiCiTiT:
> how about just looking into project's publish path template to see if keyword {silo} exists

Looking "only" into `publish` template, or all templates set in project's config? Still is possible that workfile has "{silo}" inside. (small chances but possible)

avalon/tools/models.py
```diff
@@ -6,6 +6,7 @@
 from ..vendor import qtawesome
 from .. import io
 from .. import style
+from .. import lib
```
davidlatwe, line 10: Not tested yet, but based on the changes down below, I think this relative import should change to `from . import lib` ☺️

avalon/tools/models.py
```diff
@@ -393,11 +394,11 @@ def refresh(self):
...
         # Get all assets sorted by name
         db_assets = io.find({"type": "asset"}).sort("name", 1)
-        silos = db_assets.distinct("silo") or None
+        project_doc = io.find({"type": "project"})
 
-        # if any silo is set to None then it's expected it should not be used
-        if silos and None in silos:
-            silos = None
+        silos = None
+        if lib.roject_use_silo(project_doc):
```
davidlatwe, line 400: Typo here: `lib.roject_use_silo` should be `lib.project_use_silo`

mkolar: @iLLiCiTiT boooo

avalon/tools/models.py
iLLiCiTiT: That's because I though I put `project_use_silo` into `avalon.lib`, but i guess `avalon.tools.lib` is better position since is used only in tools :)

mkolar:
Anyone had time to give this another shot? We have other stuff waiting for being put into PRs but all of it's is dependent on this one.

mottosso:
Sorry this is taking so long! How about we merge this as-is, and deal with issues as they arise? It's a little too large to be entirely confident ahead of time that nothing would break, but it looks like a step in the right direction.

davidlatwe:
Sounds good, better to move forward !
Should we make a release before merging this one ?

mottosso:
Sounds like a good idea, could you take care of it?

davidlatwe:
Done. 👍  https://github.com/getavalon/core/releases/tag/5.7.15

mottosso:
Bam!

PR #537, opened 2020-03-19
Change inventory action select back to select all columns in row

### LINKED ISSUE #536: Scene Inventory action selection cannot pass into other actions
### What happened ?
When running inventory action that selects items from process result, although the selection is visually correct (rows get highlighted), but other actions like "Remove items" can only get one item from selection (the item which has mouse clicked) to process.

### What did you expect ?
Items that selected (highlighted) by action should be able to pass into other actions.

### To Reproduce
1. Register following inventory action

    ```python
    import avalon.api
    import avalon.maya
    from maya import cmds


    class SelectFromScene(avalon.api.InventoryAction):

        label = "Select From Scene"
        icon = "hand-o-up"

        @staticmethod
        def is_compatible(container):
            return True

        def process(self, containers):
            containers = avalon.maya.ls()
            container_names = set(c["objectName"] for c in containers)

            selected = cmds.ls(sl=True)
            selected_items = set()

            for node in selected:
                if node in container_names:
                    selected_items.add(node)
                    continue

                objsets = cmds.listSets(object=node) or []
                for objset in objsets:
                    if objset in container_names:
                        selected_items.add(objset)
                        break

            return selected_items
    ```
1. Load multiple subsets into Maya scene
1. Select random containerized nodes that accross multiple subsets
1. Open up Scene Inventory and use the above action to select items
1. Remove action highlighted items and only one item gets detected

### Cause
The action select back feature was shipped in commit 2851fd0 from #353, and the implementation only select the first column in the row. 👇 

https://github.com/getavalon/core/blob/f8186bf1804aa1bc0ba5471a13f4c86c12428cd1/avalon/tools/sceneinventory/app.py#L253

But scene inventory pass items into action only if the row if fully selected. 👇 

https://github.com/getavalon/core/blob/f8186bf1804aa1bc0ba5471a13f4c86c12428cd1/avalon/tools/sceneinventory/app.py#L296

So the bug (unexpected behavior) appears..

---

I have a fix in my hand now, a PR will be submitted in no time.

davidlatwe:
### What's changed ?
The scene inventory will select entire rows by action process result, instead of selecting only first column of rows. And will fixes #536.

### DIFF
--- a/avalon/tools/sceneinventory/app.py
+++ b/avalon/tools/sceneinventory/app.py
@@ -250,7 +250,12 @@ class View(QtWidgets.QTreeView):
             name = node.get("objectName")
             if name in object_names:
                 self.scrollTo(item)  # Ensure item is visible
-                selection_model.select(item, select_mode)
+                # Select entire row
+                for i in range(model.columnCount()):
+                    column = item.sibling(item.row(), i)
+                    if column.isValid():
+                        selection_model.select(column, select_mode)
+
                 object_names.remove(name)
 
             if len(object_names) == 0:
--- a/avalon/tools/sceneinventory/app.py
+++ b/avalon/tools/sceneinventory/app.py
@@ -250,11 +250,8 @@ class View(QtWidgets.QTreeView):
             name = node.get("objectName")
             if name in object_names:
                 self.scrollTo(item)  # Ensure item is visible
-                # Select entire row
-                for i in range(model.columnCount()):
-                    column = item.sibling(item.row(), i)
-                    if column.isValid():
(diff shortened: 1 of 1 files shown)


BigRoy:
As opposed to the for loop isn't it better to use the [Rows flag](https://doc.qt.io/qt-5/qitemselectionmodel.html#SelectionFlag-enum)?

E.g.
```python
flags = select_mode | selection_model.Rows
selection_model.select(column, flags)
```

davidlatwe:
Yikes, I didn't find that !
Thanks !

davidlatwe:
Merging this 🚀

PR #539, opened 2020-03-30
Resilience against missing paths

iLLiCiTiT:
New PR based on https://github.com/getavalon/core/pull/424/files.

- each of 3 methods in `get_representation_path` return path even if formatted path does not exist
- returned path is not normalized if path does not exist

### DIFF
--- a/avalon/pipeline.py
+++ b/avalon/pipeline.py
@@ -1240,8 +1240,10 @@ def get_representation_path(representation):
             # Template references unavailable data
             return None
 
-        if os.path.exists(path):
-            return os.path.normpath(path)
+        normalized_path = os.path.normpath(path)
+        if os.path.exists(normalized_path):
+            return normalized_path
+        return path
 
     def path_from_config():
         try:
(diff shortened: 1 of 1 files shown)


mkolar:
The result is identical from the previous PR, but it returns full path instead of just folder, and it's a bit safer by checking the existence of the path first.

avalon/pipeline.py
```diff
@@ -1282,16 +1284,20 @@ def path_from_config():
...
         path = representation["data"]["path"]
-        if os.path.exists(path):
-            return os.path.normpath(path)
+        normalized_path = os.path.normpath(path)
+        if os.path.exists(normalized_path):
+            return normalized_path
+        return path
 
     return (
         path_from_represenation() or
```
davidlatwe, line 1303: Typo: `path_from_represenation` -> `path_from_representation` (missed a `t`)

davidlatwe (commented):
Haven't test it, but I think the logic is okay.
If the conditions sufficed for the function, then the path should be returned even not exists.

Some improvement notes/propose for future:

1. add kwarg `parenthood=None`
    One could provide pre-queried parenthood so wouldn't need to query again.

2. add kwarg `by="try_all"`
    This kwarg should have 4 valid input values: `try_all`, `context`, `config`, `baked`.
    And the return statement may change into like this:
    ```python
    if by == "try_all":
        return (
            path_from_representation() or
            path_from_config() or
            path_from_data()
        )
    else:
        return {
            "context": path_from_representation,
            "config": path_from_config,
            "baked": path_from_data,
        }[by]()
    ```
    This propose should provide an interface to get path from a specific method.

The above proposal maybe a bit overkill, but thought it might be worth think about.

Issue #503, opened 2019-12-31
CGWire / Kitsu Integration


BigRoy:
This is about a "connection" integration with [CGWire](https://www.cg-wire.com/) / [Kitsu](https://www.cg-wire.com/en/kitsu.html). The nice thing about CG-Wire and Kitsu is that, like Avalon, it is open source and you can run it completely free.

### Implementation references

Some prototype work was done on connecting Kitsu with Avalon in the past:

- [Example API for linking Avalon with CGWire / Kitsu](https://gist.github.com/BigRoy/c26fec6b1b789589d3cb371f07dc90f4)
- [Issue on `getavalon/docker` on CG-Wire syncing](https://github.com/getavalon/docker/issues/23) which resulted in [this syncing code](https://github.com/getavalon/docker/blob/master/volume/sync.py) and some [details discussed in its PR](https://github.com/getavalon/docker/pull/15)
- For Qt interfaces for CG-Wire/Kitsu I set up the [`qtazu`](https://github.com/Colorbleed/qtazu) repository some time ago. Having Qt interfaces ready for Kitsu would be helpful for exposing the connection with Kitsu to artists right in the host applications where Avalon is integrated.
- There is a [cgwire-sync](https://github.com/getavalon/cgwire-sync) repository however this currently seems _completely unused_.

### This is NOT a host integration

**This is _not_ a host integration like Maya, Houdini, Fusion** Instead this would be a "connection" to CG-Wire. Similar to #280 is for Ftrack. The code doesn't necessarily need to end up in Avalon core. But I'm creating this issue to track the status of any ongoing work on integrations and eventually allow others to work together on getting Avalon + CG-Wire working well togeher.

### References

- There's also a dedicated `#avalon` channel in the [CGWire Slack](https://slack.cg-wire.com/) but of course the [Avalon Gitter lobby](https://gitter.im/getavalon/Lobby) is perfectly suited to chat about it too.
- [Avalon Issues #280: Ftrack Integration](https://github.com/getavalon/core/issues/280)

frankrousseau:
I coded sync functions to synchronize two Kitsu instances. The code written for this may be helpful if you want to develop synchronization for Avalon.

https://github.com/cgwire/zou/blob/master/zou/app/services/sync_service.py

tokejepsen:
What might be good is to get an outline of goals for this "connection".

For example when working with Ftrack/Pype, they have integration plugins so any publish is immediate and there is no waiting for a "sync".

[...]

PR #540, opened 2020-04-05
CGWire Integration

tokejepsen:
**Goal**

Full synchronization with CGWire; #503 

**Implementation**

Initially this PR is mainly tests so we can experiment with the code. Currently only have tests for Avalon to CGWire.

- [x] CGWire basic integration tests.
- [x] CGWire full sync.
- [x] Non-hardcoded determination of CGWire asset/shot.
- [x] Non-hardcoded CGWire host and login.
- [x] Store Avalon Events.
- [ ] Store synced event id to pick up syncing after shutdown.
- [ ] On demand syncing Avalon to CGWire (register sync module).
- [ ] Avalon basic integration tests.
- [ ] Avalon full sync.
- [ ] Store synced event id to pick up syncing after shutdown.
- [ ] On demand syncing CGWire to Avalon (event listener).
- [ ] Travis reports false positive. Load by project name is failing.
- [ ] Fix CI (Azure pipelines)

**NOTE: This is a work-in-progress PR.**


tokejepsen:
**Current State**

- Tests for syncing project, asset, episode, sequence, shot.
- Testing for:
    - name of Avalon entity matches with CGWire entity.
    - existence of `avalon_id` in CGWire entity's `data` field.
    - tasks in project `config` matches CGWire task types.
    - tasks on assets and shots match.
- **Clears the databases between tests.**
- Full syncing only atm.
- Syncing determines asset/shot with silos names.
- Syncing determines episode/sequence/shot with visual children:
    - No visual children == skip. CGWire requires a minimum of a sequence, so 1 child minimum.
    - One visual child == sequence hierarchy.
    - Two visual children == episode hierarchy.
- Asset and shot are currently hardcoded to be in silos; `film` and `assets`.
- Finds existing entities by `avalon_id` first, then `label`, then `name`.

tokejepsen:
Currently look at storing the io events in Avalon database. Thinking that it might be better to store them on the project, so dropping an entire project will also remove the event log.
Storing all the events might also reduce performance when looking up assets.

The other option is to have a separate document with the events. Something like:
```python
client = pymongo.MongoClient(os.environ["AVALON_MONGO"])
db = client["avalon_events"]
```
This document would mirror the project names, but only store events.

tokejepsen:
**Current State**

- Entity types in CGWire are determined from two environment variables; `AVALON_CGWIRE_SHOT_PARENTS` and `AVALON_CGWIRE_ASSET_PARENTS`. These can multiple names of parents since we can have multiple visual root parent (folders) in Avalon.
- CGWire login details are determined with three environment variables; `CGWIRE_HOST`, `CGWIRE_USERNAME` and `CGWIRE_PASSWORD`.
- `io` events are stored in the database in the same collection as the project with following schema;
```
    event = {
        "schema": "avalon-core:event-1.0",
        "type": "event",
        "datetime": datetime.now(pytz.utc).isoformat(),
        "method": "insert_one",
        "args": json_util.dumps([item]),
        "kwargs": {}
    }
```
Events are validated and tested.

PR #543, opened 2020-04-22
Fix #534 by implementing name validating widget

### LINKED ISSUE #534: Project Manager allows to add assets with new line characters
**What happened?**

Somehow it's possible to add/create a new Avalon asset using the project manager that includes a new line character `\n`. It somehow manages to pass validation.

**What did you expect?**

This should never get through.

**To Reproduce**

E.g. copy past the text below including the new line:

```
invalid

```

Note how you can create an asset using this.

**Additional context**

It should not pass the [schema.validate](https://github.com/getavalon/core/blob/f8186bf1804aa1bc0ba5471a13f4c86c12428cd1/avalon/tools/projectmanager/lib.py#L58) check which it magically does. Trying to create a name with a space in it does get captured correctly, however the [schema](https://github.com/getavalon/core/blob/f8186bf1804aa1bc0ba5471a13f4c86c12428cd1/avalon/schema/asset-2.0.json#L36-L40) does not invalidate a new line. 

It only passes if the new line is at the end of the string. Even adding a space after it makes it fail.

*Below examples are written with the escape characters explicitly written, when entering these in the asset name field copy pasting won't reproduce the same behavior. `\n` must be a new line, `\t` must be a tab (not spaces).*

This incorrrectly passes validation:
`invalid\n`

This is correctly disallowed:
`invalid\n ` (note space after `\n`)
`invalid\t`
`invalid\naa`
`\ninvalid`
`\tinvalid`

I'm not sure why it allows the `\n` at the end of the name.

davidlatwe:
**What's changed?**

This fixes #534, not only invalid character will be blocked in GUI, invalid input from pasting will be cleaned for better UX.

But since Creator App also shares the same widget, it has changed, too.

69d82bae-f008-4036-804f-a1c3a407f39e

### DIFF
--- a/avalon/tools/creator/app.py
+++ b/avalon/tools/creator/app.py
@@ -1,13 +1,12 @@
 import sys
 import inspect
-import re
 
 from ...vendor.Qt import QtWidgets, QtCore, QtGui
 from ...vendor import qtawesome
 from ...vendor import six
 from ... import api, io, style
 
-from .. import lib
+from .. import lib, widgets
 
 module = sys.modules[__name__]
 module.window = None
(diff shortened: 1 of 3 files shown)


BigRoy:
@davidlatwe That's great work! :) Lovely. 🚀 

I wonder however if it should include the reported message (the invalid characters) somewhere on the Widget too so the user is aware of that aside of the "red fade" on the QLineEdit. What do you think? Or would it be overkill?

davidlatwe:
Thanks @BigRoy  :D

Yeah, I agree that it would be better if we put a message printer in the bottom of that dialog.
Furthermore, should we also check the asset name exists or not ? Like the Creator App did.

Issue #544, opened 2020-04-29
Workfiles close on Open


tokejepsen:
**Goal**

When hitting "Open" in the new Workfiles app, the dialog should close.

**Motivation**

This is behaviour from previous version of the Workfiles. This just reduces the clicks by one, which I think it significant when talking about opening work files.

jrsndl:
+ one to that

PR #545, opened 2020-04-29
Close Workfiles window on Open

tokejepsen:
Fix for #544

### DIFF
--- a/avalon/tools/workfiles/app.py
+++ b/avalon/tools/workfiles/app.py
@@ -595,7 +595,8 @@ class FilesWidget(QtWidgets.QWidget):
             print("No file selected to open..")
             return
 
-        return self.open_file(path)
+        self.open_file(path)
+        self.window().close()
 
     def on_browse_pressed(self):
 
--- a/avalon/tools/workfiles/app.py
+++ b/avalon/tools/workfiles/app.py
@@ -516,7 +516,8 @@ class FilesWidget(QtWidgets.QWidget):
                 pass
 
         self._enter_session()
-        return host.open_file(filepath)
+        host.open_file(filepath)
+        self.window().close()
 
     def save_changes_prompt(self):
         self._messagebox = QtWidgets.QMessageBox()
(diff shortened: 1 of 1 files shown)


tokejepsen:
Seems to be some issues with coveralls and MacOS builds.

This PR is ready to merged otherwise.

mkolar:
I'm happy with this

BigRoy:
I did intentionally remove the behavior when I implemented the new tool as I used it much more to quickly browse and go between different scenes where having it open appeared faster to me. Nevertheless, I'm fine with closing on open - especially because it seems the preference for most people. 👍

BigRoy:
> Just a note. This only changes the behavior of on_open_pressed, not on_browse_pressed.

That's actually a very good point. Maybe this is better placed in the `open_file` method of the Widget.

tokejepsen:
> That's actually a very good point. Maybe this is better placed in the open_file method of the Widget.

Sure, we can do that!

iLLiCiTiT:
+ I would add "action" to workfiles view which will allow to open workfile and keep window opened if @BigRoy need that?

BigRoy:
> I would add "action" to workfiles view which will allow to open workfile and keep window opened if @BigRoy need that?

Not sure what you mean with Action? Do you mean a toggle somewhere, or a right mouse context menu?

Anyway, having the option to keep it open is no hard requirement for me. So I'd be happy to merge as is and potentially open an issue if it becomes a problem.

iLLiCiTiT:
> Not sure what you mean with Action? Do you mean a toggle somewhere, or a right mouse context menu?

Sorry for ugliness.
image

BigRoy:
Ah yes, I wouldn't bother. It wouldn't be very clear that that specific one would then **not** close the Window, right? 

For me this is ready to merge as is.

BigRoy:
For what it's worth - just heard back from some Artists who also stated they "tend to keep it open" to quickly browse between different shots for some minor things. I'll create a separate issue to discuss a potential "toggle" to keep the tool open.

davidlatwe:
If no other objections, merging this in one hour !

PR #547, opened 2020-05-05
Close on save

tokejepsen:
In addition to #545 I propose the window to close on saving as well.


tokejepsen:
Dont think this has any traction.

PR #548, opened 2020-05-08
Prevent double save confirmation

tokejepsen:
Previous behaviour resulted in double confirmation about saving the Nuke script. To be precise with "saveScriptAs" Nuke asks whether to overwrite the file, but this has already been confirmed by the user in the Workfiles app.


BigRoy:
How is the changed version applying the save path? :/ Code-wise this doesn't make sense.

tokejepsen:
Actually I dont this is worth the effort. Closing...

PR #550, opened 2020-05-26
Asset multiselection in loader 

### LINKED ISSUE #453: Loader - allow asset multiselection and bulk load
**Problem**
This is a repeating request from multiple clients to speed up set dressing. Usually when set dressing artists need to import a lot of assets at the same time especially when making cities forests and similar location which need a lot of individual assets with smaller variation in them. Currently this takes ages, becuase to import 10 trees (unless they are subsets) I need to click on asset, right click subset, load...repeat.

**Proposed Solution**
The solution would be to allow artist to select any number of assets on the left, which would show the union of all their available subsets on the right. The artist could import say 10 buildings at the same time. The problem of course is knowing what subsets belong to what asset and so on. We've tried solviing this with some grouping and color coding to make it clear what is going to be affected. I'm attaching a preview without much more explanation in hope that the UIX is clear enough on it's own, If it isn't we might want to think a bit more. 

We've deployed this to testing now and the first feedback from artists is very positive. Have a look

multiselect

**Additional context**
Combined with #346 this might become mighty set dressing tool.

To make PR we'll first have to get in line with current state of avalon-core, which is almost done, so we have some time to discuss possible changes.

iLLiCiTiT:
# Feature: Multiselection of assets in loader
With this feature it is possible to load multiple subsets from multiple assets at once.
65226426-1eb47200-dac7-11e9-80f8-ad4f5f4f5a35

This feature is used only in loader, rest of tools should work the same (visually) as before and if only one asset is selected then nothing should change too.

How it works (basics):
- When multiple assets are selected, subsets with same name are grouped in subsets view with colored circle, other are just kept in list.
- At the same time "Asset" column is added to know which item is related to which asset.
- Subset group has number representing number of assets having the subset.
- Group color is helper to identify the subset. There are 10 predefined(hardcoded) colors atm. They are reused if more than 10 subsets has same name.
    - If colored group is selected you can see in assets view that selected assets has bar with colors. The color helps to identify which subsets are available for the asset -> `missing color == missing subset`.
- It is not possible to see subset groups when multiple assets are selected.

RESOLVES: https://github.com/getavalon/core/issues/453

### DIFF
--- a/avalon/style/style.qss
+++ b/avalon/style/style.qss
@@ -1221,3 +1221,7 @@ QSplitter::handle:horizontal:hover {
 QSplitter {
     border: 0px;
 }
+
+#AssetsView[multiselection=true] {
+    selection-background-color: rgba(0, 0, 0, 0);
+}
--- a/avalon/tools/delegates.py
+++ b/avalon/tools/delegates.py
@@ -3,10 +3,16 @@ from datetime import datetime
 import logging
 import numbers
 
-from ..vendor.Qt import QtWidgets, QtCore
-from .. import io
+from .models import TreeModel, AssetModel
 
-from .models import TreeModel
+from .. import io
+from ..vendor.Qt import QtWidgets, QtGui, QtCore
+
+from ..vendor import Qt
+if Qt.__binding__ == "PySide":
+    from PySide.QtGui import QStyleOptionViewItemV4
+elif Qt.__binding__ == "PyQt4":
+    from PyQt4.QtGui import QStyleOptionViewItemV4
 
 log = logging.getLogger(__name__)
 
@@ -28,7 +34,7 @@ class VersionDelegate(QtWidgets.QStyledItemDelegate):
 
     def createEditor(self, parent, option, index):
         item = index.data(TreeModel.ItemRole)
-        if item.get("isGroup"):
+        if item.get("isGroup") or item.get("isMerged"):
             return
 
         editor = QtWidgets.QComboBox(parent)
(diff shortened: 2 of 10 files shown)


davidlatwe:
Had a try on this feature, it looks like the original subset groups gets *un-grouped* if multiple assets selected.

c5ee4a7c-d7d0-4c16-9cbb-d82c0f827267

> It is not possible to see subset groups when multiple assets are selected.

Just saw this line at the end. XD

This might be a problem for my studio, the artists here sometimes produce subsets that were just for testing or marking them as deprecated in subset groups which named like "Trash Bin" or "Tests".

If those subsets that should have been hidden will re-appear when enabled multi-assets selecting, then it's kind of like backward incompatible in UX.

Could you elaborate a bit more about why it is not possible ? The first problem that I can think of is how to visualize two different subset grouping motive in one view, and yes, it is really tricky.

davidlatwe:
One potential solution would be to change the subset grouping's visual implementation from group item in subset tree view to 
leaf item in asset tree view.

The goal of subset grouping was like creating visual partitions inside each asset, so it should also be reasonable if the grouping gets implemented in asset tree view, like the leaf item at the end of any asset branch.

*Let's change the term from `subset group` to `asset partition` for now.*

And we could even treat that *asset partition* like a normal asset, and grouping same named subset in subset tree view when using multi-asset selecting.

Seems like a good way to go !

iLLiCiTiT:
> Let's change the term from subset group to asset partition for now.

So, you're saying, that subset groups won't be shown in subset view but will kind of "children" visualization in asset view?

davidlatwe:
> So, you're saying, that subset groups won't be shown in subset view but will be kind of "children" visualization in asset view?

Yes, that's right !

davidlatwe:
> Let's change the term from subset group to asset partition for now.

I think we have to resolve the *visual subset grouping conflict* befroe merging this, so I'll try submit an issue with a proof of concept for *asset partition*.

iLLiCiTiT:
> I think we have to resolve the visual subset grouping conflict befroe merging

It looks like _asset partition_ is similar idea to _visualParent_ concept.

davidlatwe:
Yeah, but it's at the end of the asset's hierarchy.

mkolar:
Truth be told I'd prefer thinking of a different solution, rather than moving the groups under assets. It is not very consistent, especially when you only have some subsets in group and others are not. We use the grouping for render AOVs for example.  Having all of the subsets under assets, but having to click one level lower for renders feel very counter-intuitive. 

We actually have a way of dealing with "deprecated" in a different way. Rather than grouping them and just relying on the artist to name the `trash` group consistently, we change the type in mongo from `subset` or `asset` to `archived_subset`, `archived_asset`. That way they automatically drop out of the GUI, but stay in the system. 

The question is whether it is not more systematic to allow for doing this from the loader and giving the option to "un-archive" from there when needed as well. That would solve the first issue with your `bin group` workflow.

The second of course is the fact that it would be great to preserve both behaviours at the same time, but it could be though out properly how to do it visually. Should the group show above the dynamic multiselection grouping, or bellow for example?

That being said let's think of how to actually do this with full compatibility of course. Just wanted to throw the `archive` approach out there

mkolar:
We had a good old school paper brainstorm in the office and thin we have an elegant solution, we'll try to implement.

If you have existing groups, we'll treat them similarly as subsets and when you select multiple assets, we will visually keep any groups in the subset view and only group subsets themselves one level lower. That way, if multiple assets share the same groups and subsets, we should be able to fully keep that hierarchy. Of course if asset 1 has `group1/modelMain` while asset 2 has `modelMain` mong subsets. These two won't be considered as equivalent in multiselection if the `enable grouping` tickbox is on. Once it's turned off, they will be consider as equivalent in the same way as current implementation. 

This way we should keep best of both worlds and on top of that let the artist choose what the prefer by switching the grouping tickbox on and off.

davidlatwe:
Sorry for late reply, was on vaca pass few days. ☺️ 

> If you have existing groups, we'll treat them similarly as subsets and when you select multiple assets, we will visually keep any groups in the subset view and only group subsets themselves one level lower.

I did think of that path, but I was worried that pushing existing subset groups one level lower would complecate codes, and *maybe or maybe not*, require a few more clicks for user to expand groups and see what's in it.

The reason why I proposing *asset partitioning* was mostly because it's easier to implement and maintain in my imaging, and the GUI interaction result is also easier to match user's expectaion when involving multi-assets selection.

davidlatwe:
> I did think of that path, but I was worried that pushing existing subset groups one level lower would complecate codes

Having that said, maybe I was thinking wrong. I'll see if I could implement the *asset partition* and trying both ways for the best outcome.

BigRoy:
Would love to know where we're at with this feature.

iLLiCiTiT:
> Would love to know where we're at with this feature.

We though about solution and there is one, but we have to find spare time for implementation.

iLLiCiTiT:
Updated code but don't have ready testing environment to realy check if I did it right. Can someone test, please?

davidlatwe:
I checked out this branch directly and looks like the data was loaded correctly, but the view can not show anything ?
But there are colored underlines on selected assets when I selecting subsets blindly.

image

Not sure what problem would it be, updating PyQt5 version doesn't resolve this.

davidlatwe:
I comment out these lines and the subset widget is showing stuff now.

https://github.com/getavalon/core/pull/550/files?file-filters%5B%5D=.py#diff-657d23f15ec9e63b3877669b575e43a9e4532a6ef82b2a6633e77e1c3b67c756R453-R460

Looks like the problem was the loading screen didn't go off. :D

iLLiCiTiT:
> but the view can not show anything

My fault, forgot to add few lines.

> Looks like the problem was the loading screen didn't go off. :D

The issue was that loading state was never set, so `is_empty` was always True.

davidlatwe:
Thanks !
Just had another quick test, can confirm that works now.
But sometimes (really rare, but happens) the loading screen still there.

Anyway, I'll try reviewing this one ASAP.

PR #551, opened 2020-05-27
Asset switcher efficiency enhancement

iLLiCiTiT:
# Problem
Current asset switcher dialog is querying all documents in very inefficient way. The same document may be queried many times in separate database calls which makes asset switcher GUI really slow.

# Suggested solution
Query documents of one type only once per each refresh step. So all representations are queried in one call, their versions, subsets and assets are queried in separate calls which means, to get parenthood of all representations needs a maximum of 4 calls to DB. We also validate each step of refresh, to prevent querying data which will not be used. This is not perfect change, it can be done even faster (and more complicated), but the performance improvement over current implementation is significant.

We tested on a feature film with around 400 assets. The speed difference to open the switcher GUI is: 10s with current code, 1-2s with the new code.


avalon/tools/sceneinventory/app.py
```diff
@@ -454,12 +454,13 @@ def get_valid_value(self):
...
             return None
 
-        return text
+        return text or None
 
 
 class SwitchAssetDialog(QtWidgets.QDialog):
     """Widget to support asset switching"""
 
+    _fill_check = False
```
BigRoy, line 463: Shouldn't this be defined in `__init__`?

iLLiCiTiT: It was there for a reason, but the reason is now irrelevant. So yes, should.

BigRoy:
I believe @jasperges recently tested this feature and it "worked fine" for his needs. I believe @iLLiCiTiT at pype has also updated a variant of this code change.

Should we get this merged?

iLLiCiTiT:
> I believe @jasperges recently tested this feature and it "worked fine" for his needs.

@tokejepsen found a bug. https://github.com/pypeclub/pype/issues/423

PR #552, opened 2020-05-29
Toggle "Save As" button in Workfiles.

tokejepsen:
**Changes**
This PR enables developers to disable the "Save As" button when showing the Workfiles app.

```
import avalon.tools.workfiles
avalon.tools.workfiles.show(save=False)
```

**Motivation**
The specific use-case for this is the Harmony host. Unfortunately you cannot open a scene in an active application instance of Harmony, so you need to relaunch Harmony as a new process. For the Workfiles flow this means the user has to open an instance of Harmony then open it again to launch the scene they want.
With this PR we can show the Workfiles tool before the application is launched, and the user can open the scene file they want, reducing the wait by one application launch.

This "launch before application" behaviour can be ported to other hosts, but its less necessary if you can prevent launching a new process.

### DIFF
--- a/avalon/tools/workfiles/app.py
+++ b/avalon/tools/workfiles/app.py
@@ -854,7 +854,7 @@ class Window(QtWidgets.QMainWindow):
         files.refresh()
 
 
-def show(root=None, debug=False, parent=None, use_context=True):
+def show(root=None, debug=False, parent=None, use_context=True, save=True):
     """Show Work Files GUI"""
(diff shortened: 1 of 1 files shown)


jezscha:
Hi. How about to name the proposed attribute `save` to something like `saving` or `save_as`. This way it would be more explicit.

tokejepsen:
`save` maps directly to the widget.

tokejepsen:
This PR is redundant now.

PR #553, opened 2020-06-09
Non blocking loader

davidlatwe:
### Problem

Fetching subsets from database is blazing fast, but finding the last version for every subset could take a long time depends on how big the project is (or how well the asset are categorized).

### Solution

Refactor Loader's `SubsetsModel.refresh`, push version document lookup into work thread.

### What's Changed

* Before

before

Some of our asset may take up to 8 sec for GUI to respond. The asset which has longest waiting time, has about 200 subsets and some of them have 5-10 versions.

* After

after

With this PR, switching view between assets is much much faster (only fetching subset in main thread).

Versions that has not been fetched has a spinning circle animation, and will be updated once the version is fetched in **worker thread**.

User can double click one the spinning version and the **main thread** will fetch that version immediately and present the version editor.

User can still load the subsets without waiting versions being presented, the context menu action will take the latest version by default.

### Additional Changes

* Displaying child count on subset group item.
* Disabled QTreeView header's `ResizeToContents` resize mode to improve the performance.

   **EDIT:** We actually might need this change because when version data updated from worker thread, emitting `dataChanged` signal is not enough to trigger the column resize, the widget update call from the spinning version delegator could not trigger that as well.

  And one may find the version data like `author`, `date` gets cropped but could not resize the column from header, only drag the splitter or open/close the group item to trigger the layout change could work. :(

### Notes

While I was implementing this, I find that emiting `dataChanged` signal in worker thread is dangerous, may bring down the entire application (include the main app like Maya) if user switching and clicking stuff in a really fast pace. 😟 

It looks like it's because the signals get queued if sender and reciver are in different threads (see [doc.qt.io](https://doc.qt.io/qt-5/threads-qobject.html#signals-and-slots-across-threads)). And the `index` that shipped with `dataChanged` signal may be already outdated when it's being processed.

But after I added spin animation in the version delegator (powered by `QtSvg.QSvgRenderer`), not only we got a beautiful loading animation, but also a GUI updater ! (Since the animation needs to emit signal to trigger `paint` in order to render out *next frame*.) So now we won't need to inform main thread that data has changed from work thread, the spinner already took care of that in main thread. 🍻


avalon/tools/loader/app.py
```diff
@@ -262,13 +264,8 @@ def closeEvent(self, event):
             print("Force quitted..")
             self.setAttribute(QtCore.Qt.WA_DeleteOnClose)
 
-        # Kill on holding SHIFT
-        modifiers = QtWidgets.QApplication.queryKeyboardModifiers()
-        shift_pressed = QtCore.Qt.ShiftModifier & modifiers
-
-        if shift_pressed:
-            print("Force quitted..")
-            self.setAttribute(QtCore.Qt.WA_DeleteOnClose)
```
BigRoy, line 271: Is there a particular reason this code seems to be removed? Not crucial to me but I am just wondering whether was need to remove it since it seems unrelated to this PR.

avalon/tools/loader/widgets.py
```diff
@@ -98,15 +98,21 @@ def __init__(self, enable_grouping=True, parent=None):
...
-            header, QtWidgets.QHeaderView.ResizeToContents)
+        view.setColumnWidth(0, 240)  # subset
+        view.setColumnWidth(1, 120)  # family
+        view.setColumnWidth(2, 80)   # version
+        view.setColumnWidth(3, 120)  # time
+        view.setColumnWidth(4, 100)  # author
+        view.setColumnWidth(5, 80)   # frames
+        view.setColumnWidth(6, 60)   # duration
+        view.setColumnWidth(7, 60)   # handles
+        view.setColumnWidth(8, 60)   # step
```
BigRoy, line 109: > Disabled QTreeView header's ResizeToContents resize mode to improve the performance.

What's the performance difference for this? This doesn't seem like the thing on first sight that would be super but I can imagine it would need to go through all data so I could understand why it is. Just wouldn't have thought this would make a noticeable difference.

avalon/tools/loader/app.py
davidlatwe: Ah, this part of code was actually duplicated, there are exactly same piece of code avobe this section. It looks like it was duplicated from the commit 3fea272.
And yeah, it's unrelated, just thought that it could be removed all toghter while I was adding some stuff there.

avalon/tools/loader/widgets.py
davidlatwe: The improvement might only be noticeable when you have lots of subsets, the GUI may appear a bit laggy when you drag the splitter which will change the layout and make the view to update the header to fit the contents.

 So I thought maybe we could disable it and use default size for each column ?

Actually, I forgot to mention another reason, when version data updated from worker thread, emitting `dataChanged` signal is not enough to trigger the column resize, and the widget update call from the *spinning version delegator* could not trigger that as well.

And one may find the version data like `author`, `date` gets cropped but could not resize the column from header, only drag the splitter or open/close the group item to trigger the layout change could work. :(

davidlatwe:
> EDIT: I think this point is actually void because we only query for the last version, right?

Yes ! :D

avalon/tools/loader/widgets.py
davidlatwe: I have patched the reason that I forgot to mention into the top comment. :)

mottosso:
This is good some work! Now we're moving beyond what results I've seen in some of the best in-house tools. Nobody cares about this stuff, except all of the artists.

My first thought is whether this could overwhelm MongoDB? Before, an artists was only able to make 1 of these heavy requests at the same time, which when you've got 10-100 artists would max out at 100 concurrent requests. But if 10+ requests can come from each individual computer, it'd max out beyond that. 

So, it would be interesting to see how strained Mongo actually is in practice, at e.g. your company @davidlatwe. Do you have anything in place to monitor how busy the machine is during these queries? Odds are each request is a drop in the ocean and scales up to 1000x times more. But it's also possible each query freezes Mongo entirely, such that you would get synchronous performance, regardless of how many you've got going in parallel?

The second thing is whether there's a better way to actually query the data in the first place. I think the bottleneck lies in the recursive calls being made, to traverse a hierarchy of assets? Hierarchies are always tricky.. the most common method that I've seen involves caching. Either locally in the app or on the machine, like a temp-json, or server-side in another app like `memcached`. The idea generally being that you query the database less frequently, and update the cache infrequently, or on-demand. Because typically, most of those requests query the same unchanged data and return the same results again.

davidlatwe:
> Nobody cares about this stuff, except all of the artists.

I think that maybe just because that I am sitting right next to our artists without any walls or any kind of buffer. XD

> My first thought is whether this could overwhelm MongoDB?

Hmmm, good point. From the implementation I made, the difference was basically move the version query from the subset loop inside `SubsetModel.refresh` into one worker thread. 

So the total query count for listing subsets and versions should remain 2, I think...

But the period between queries indeed become a lot shorter. :/

> Do you have anything in place to monitor how busy the machine is during these queries?

I have to check that with our IT ! Currently we have about 40-50 machines could make queries to Avalon in the same time, shouldn't be a problem for our database setup, I guess..(single instance running inside a virtual machine) :/

> The second thing is whether there's a better way to actually query the data in the first place.

I did think of this, if we update the latest version document `_id` into subset document each time we publish, then it would be much faster since no document sorting (for finding latest version) would needed. (Not tested but I imaging it could bring imporvements)

*Or*, simply showing "latest" on the GUI, and only if user click on it will show the version list. But it would make the UX become *"not so good"* in my opinion.

> Because typically, most of those requests query the same unchanged data and return the same results again.

Most of the case I think it's true, but if the pipeline requires to change document from time to time, like the `subsetGroup` in subset document, or the file list in repesentation, the cache would failed, right ?

BigRoy:
> Most of the case I think it's true, but if the pipeline requires to change document from time to time, like the subsetGroup in subset document (which is for grouping subsets in Loader GUI), or the file list in repesentation, the cache would failed, right ?

Well. It would display the older data until the cache gets invalidated. So yes, technically using caching older data could be shown in some cases when the cache is not updated/cleared before the query.

> Hmmm, good point. From the implementation I made, the difference was basically move the version query from the subset loop inside SubsetModel.refresh into one worker thread.
> 
> So the total query count for listing subsets and versions should remain 2, I think...

Indeed little has changed with this implementation I believe. Even before there quite some queries to get the specific versions for each subset, with at least one per subset. If the amount of queries are what's slow as opposed to the amount of data being returned than technically all subsets could get their versions queried at once and then later within Python just "grouped" in a dictionary to the specific Subsets. Yes, it does mean you have another loop in Python to filter those however subsets on a single asset are usually under one hundred, which means just iterating roughly 100 entries in a list. So it could be worthwhile to bring back the 100 queries to 1 and then filter like that... If we're optimizing things this could be worth looking into. Especially if that filtering happens in another thread anyway. (Plus you could "propagate" the update for the versions as you finished filtering each, e.g. query with sorting by parent id and then iterate with `itertools.groupby`?

mottosso:
> Or, simply showing "latest" on the GUI, and only if user click on it will show the version list. But it would make the UX become "not so good" in my opinion.

Ah! Yes come to think of it, the reason this was so familiar was because this is what I did for Allzpark. With Rez, queries are far far slower, they're being made towards disk, where each version is an individual Python file (!). Furthermore, there are far more versions in an ever-growing repository (as opposed to one that ends with the completion of a project).

But yes, having versions loaded on clicking could reap enough performance benefits to avoid needing much else. It would mean that you'd have to start waiting when you ask for a version, which is the _least_ suitable place to wait (not quick enough to not notice, and not long enough to have time to do something else). So that's a tradeoff for sure. On the upside, if you're only grabbing versions for a single subset, then that query would take a lot less time than if you were  forced to wait for all of them to finish before you can pick just one.

> Most of the case I think it's true, but if the pipeline requires to change document from time to time, like the subsetGroup in subset document (which is for grouping subsets in Loader GUI), or the file list in repesentation, the cache would failed, right ?

Exactly. Easiest thing around that is a manual "refresh" for artists, in case they are expecting a newer version to appear or just aren't sure. Then there's lots of more options, like deeming any cache older than 1h to be too old, and recache it. Or pushing an update from your publishing pipeline to say that "hey, all caches made from the date of the last version of this subset are now old".

All of that said, first port to call is to see whether there is any need for any of that. Because these are just optimisations to complicate things.

davidlatwe:
> technically all subsets could get their versions queried at once and then later within Python just "grouped" in a dictionary to the specific Subsets.

Wow.

I did thought of using aggregation to get all the subsets with all latest versions in *one command* and it just won't work. 

But I revisit this with just getting all subsets first, and then do the aggregate, bam ! 
**8 sec down to less than 1 and all versions fetched !!**

Here's the aggregation pipeline I just composed:

```python
pipeline = [
    {"$match": {"type": "version", "parent": {"$in": subset_ids}}},
    {"$sort": {"name": 1}},  # Sorting versions all together
    {"$group": {"_id": "$parent",  # Take only one document from same "parent"
                "name": {"$last": "$name"},  # Take the *last* one
                "version_id": {"$last": "$_id"},  # Move version _id here
                "data": {"$last": "$data"}}},  # Keep data
]

versions = {doc["_id"]: doc for doc in io.aggregate(pipeline)}

# Matching versions with subsets that queried previously
...
```

Haha, how come I insist to do it in one aggregation and give up to use threads, but not trying with one query + one aggregation. 😭

BigRoy:
> 8 sec down to less than 1 and all versions fetched !!

Well, that combined with being in a thread. Win-win?

davidlatwe:
Hmmm, I think we could even drop the threading, the speed of the aggregation is blazing fast, and some other new features like *"load latest version if version not fetched"* could definitely be dropped as well. No human fingers could faster than mongodb's aggregation. (unless the asset has over 1000 subsets and 100 versions each of subsets)

If you guys don't mind, I'll create another branch to compare and submit a new PR if required.

Thanks for all the inputs !! 🚀

mottosso:
Very good news!

mkolar:
Give us till Monday please before merging, we'd like to double-check changes between this and #550 so we can get on the same page for smooth merges ;)

iLLiCiTiT:
> the speed of the aggregation is blazing fast

I like it's fast but aggregations are (for me) hard to read and understand on first view. Main disadvantage is that we can't use this aggregation in our fork because we have few changes here and there.

Is there any way that this approach can be accepted?
```
def latest_versions_by_subset_id(asset_id):
    subset_ids = [
        subset["_id"]
        for subset in io.find({"type": "subset", "parent": asset_id})
    ]
    version_docs = io.find({
        "type": "version",
        "parent": {"$in": subset_ids}
    }, sort=[("name", -1)])
    last_versions = {}
    for version_doc in version_docs:
        subset_id = version_doc["parent"]
        if subset_id not in last_versions:
            last_versions[subset_id] = version_doc
    return last_versions
```

BigRoy:
> Is there any way that this approach can be accepted?

That version query could easily start returning hundreds of versions right? It's not limited to querying only the latest version for each subset but the query will return all versions for all subsets for the particular asset. I might be micro-optimizing this but it doesn't seem to sensible.

Especially since versions usually contain quite some additional data that could result in a bulky result.

> I like it's fast but aggregations are (for me) hard to read and understand on first view. Main disadvantage is that we can't use this aggregation in our fork because we have few changes here and there.

What in particular from the aggregation won't work in your fork? Isn't the resulting data in the end the same data? Is it "master versions" that is complicating things? If so, could that also be in the aggregation pipeline?

davidlatwe:
I am currently trying another approach that has much less code changes in the subset-version fetching loop in `SubsetsModel.refresh` but still have the speed improvement (with aggregation 😉).

Which I think would also be easier for merging #550. Stay tuned !

One thing to be noted, this PR lists out all subsets then fetching versions later, so if there is subset that has no verisons at all, it will remian in view. Although this is a rare case in production but better to avoid that. The another approach I am currently trying will not have this issue.

iLLiCiTiT:
> Isn't the resulting data in the end the same data?

It was actually my bad. Was testing in hurry in the evening before day off, and made a mistake. So no problem anymore and I apologize for the hasty opinion.

> Is it "master versions" that is complicating things?

Not really, master versions can be queried in different DB call since there is one or none document per subset.

> Which I think would also be easier for merging #550. Stay tuned !

I looked there now and there are some nonsense queries I'll have to modify first.

BigRoy:
Has this been superseded by #555 - if so, should this be closed?

PR #554, opened 2020-06-09
Add current index preserving in qtreeview context

davidlatwe:
### Problem

Loader's subset view gets flushed after clicking asset widget's refresh button.

Subset widget needs asset document id to get subsets. And the asset document id is retrived from view's `currentIndex`.

https://github.com/getavalon/core/blob/3761095b151c47f5f6a905e1628900def51f7c47/avalon/tools/widgets.py#L90-L93

And it looks like the `tools.lib.preserve_states` did not preserve the asset's `view.currentIndex` while refresh the asset model.

https://github.com/getavalon/core/blob/3761095b151c47f5f6a905e1628900def51f7c47/avalon/tools/loader/app.py#L171-L173

https://github.com/getavalon/core/blob/3761095b151c47f5f6a905e1628900def51f7c47/avalon/tools/widgets.py#L79-L83

### Solution

Add `currentIndex` preserve function into `tools.lib.preserve_states`.

### DIFF
--- a/avalon/tools/lib.py
+++ b/avalon/tools/lib.py
@@ -113,6 +113,7 @@ def preserve_states(tree_view,
                     role=None,
                     preserve_expanded=True,
                     preserve_selection=True,
+                    current_index=True,
                     expanded_role=QtCore.Qt.DisplayRole,
                     selection_role=QtCore.Qt.DisplayRole):
     """Preserves row selection in QTreeView by column's data role.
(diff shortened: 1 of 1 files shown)


BigRoy:
Code looks good. Admittedly to match the other argument signatures I'd expect it to be `preserve_current_index` as opposed to `current_index` but I must admit the naming convention was long (and a tad redundant) on the others so I can see why just going for `current_index` is sensible.

Feel free to decide what to do yourself whether keeping it or changing it makes more sense.

Good stuff!

mkolar:
Looks good to me. It was actually recently reported by one of the clients.

Issue #533, opened 2020-03-12
Work Files tool save modified changes on scene open prompt confusing


BigRoy:
**What happened?**

There's been a note on the Work Files tool by one of our artists regarding it asking to save or don't save on opening another scene. Somehow this one artist tries to go to quick and ends up clicking the wrong button.

This is the current prompt:

![afbeelding](https://user-images.githubusercontent.com/2439881/76529480-a2e22380-6472-11ea-85ad-b371112bdb4c.png)

**What did the user expect?**

He asked instead to change the buttons "Yes" and "No" to "Save" and "Don't Save" which makes it more similar to e.g. what Maya does by default.

Maya's default prompt:

![afbeelding](https://user-images.githubusercontent.com/2439881/76529712-f6ed0800-6472-11ea-97cb-ce591a12ce52.png)

Houdini's default prompt:

![afbeelding](https://user-images.githubusercontent.com/2439881/76529755-08ceab00-6473-11ea-845d-f12fb2d6f4fe.png)

And the best UX reference, Paint:

![afbeelding](https://user-images.githubusercontent.com/2439881/76529832-1f750200-6473-11ea-91ad-7588f6c243ea.png)

It does make the specific buttons more explicit and seems more in line with the hosts. I am wondering whether others have heard this too or feel the same (or indifferent about it).

Any thoughts on this?

tokejepsen:
Sure, sounds good to me.

mkolar:
Never heard the complaint, but don't see a reason not to change it as you're suggesting.

PR #555, opened 2020-06-21
Loader with aggregation and visual loading state

davidlatwe:
This PR supersede and will close #553.
The problem in #533 was, quoting from https://github.com/getavalon/core/pull/553#issuecomment-643961476:

> One thing to be noted, this PR lists out all subsets then fetching versions later, **so if there is subset that has no versions at all, it will remain in view.** Although this is a rare case in production but better to avoid that. The another approach I am currently trying will not have this issue.

And this is the PR what was promised in that comment.

### What's changed
* Subset, version fetching speed improved by involving MongoDB's aggregation.
* The subset, version fetching is running in worker thread and is breakable.
* Added loading spinner on subset tree view, provide visual feed while waiting results from database, although you may not even sees it in most cases. (aggregation is fast !)
* Displaying "No data" in subset tree view if no versioned subset found.
* Previously, toggling "Enable Grouping" check box will refresh the view and query the database again, now it will only reset the model. This was benefit from caching documents in worker fetching thread implementation.
* Disabled QTreeView header's `ResizeToContents` resize mode to improve the performance of layout changing. (Inherited from #553)

And here's the visual result. Note that I am not in the office at the moment, so I mocked the loading time in order to see the loading spinner in view.

a2465c4d-5e79-4992-af0d-cd52b11e3522

And sorry for Pype guys @mkolar, @iLLiCiTiT , need another review from you and test if this PR will cause any trouble to merge #550.

Please let me know what you think !


davidlatwe:
Hmmm, just had a stab on merging #550, seems not an easy task.

avalon/tools/loader/widgets.py
```diff
@@ -309,6 +328,44 @@ def echo(self, message):
...
+                    rect.y() + rect.height() / 2 - size / 2)
+        rect.setSize(QtCore.QSizeF(size, size))
+        painter = QtGui.QPainter(self.viewport())
+        self.spinner.render(painter, rect)
+
+    def paint_empty(self, event):
+        painter = QtGui.QPainter(self.viewport())
+        rect = event.rect()
+        rect = QtCore.QRectF(rect.topLeft(), rect.bottomRight())
+        qtext_opt = QtGui.QTextOption(QtCore.Qt.AlignHCenter |
```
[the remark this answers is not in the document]
davidlatwe: Either before or after line break, Hound barks. 😠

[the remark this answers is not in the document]
iLLiCiTiT: We use line break before so operators are on new line (both won't work with hound if hound is not set ;) ).

I would personally in this case use:
```
qtext_opt = QtGui.QTextOption(
    QtCore.Qt.AlignHCenter | QtCore.Qt.AlignVCenter
)
```

avalon/tools/loader/model.py
```diff
@@ -141,17 +150,72 @@ def set_version(self, index, version):
...
+            _pipeline = [
+                # Find all versions of those subsets
+                {"$match": {"type": "version", "parent": {"$in": _ids}}},
+                # Sorting versions all together
+                {"$sort": {"name": 1}},
+                # Group them by "parent", but only take the last
+                {"$group": {"_id": "$parent",
+                            "_version_id": {"$last": "$_id"},
+                            "name": {"$last": "$name"},
+                            "data": {"$last": "$data"},
```
iLLiCiTiT, line 180: I would add "type" key here, but don't have explanation why :)

avalon/tools/loader/model.py
```diff
@@ -141,17 +150,72 @@ def set_version(self, index, version):
...
+            _subsets = list()
+            _ids = list()
+            for subset in io.find({"type": "subset",
+                                   "parent": self._asset_id}):
+                if self._doc_fetching_stop:
+                    return
+                _subsets.append(subset)
+                _ids.append(subset["_id"])
+
+            _pipeline = [
```
iLLiCiTiT, line 171: Would be great if there is (e.g. in avalon.lib) function for this pipeline to be usable in multiple places.
```
def last_version_pipeline(subset_ids):
    return [
        # Find all versions of those subsets
        {"$match": {"type": "version", "parent": {"$in": subset_ids}}},
        ...
    ]
```

But it is only nice to have, because it is not used in other places yet.

avalon/tools/loader/widgets.py
```diff
@@ -309,6 +328,44 @@ def echo(self, message):
         print(message)
 
 
+class SubsetTreeView(QtWidgets.QTreeView):
+
+    def __init__(self, parent=None):
+        super(SubsetTreeView, self).__init__(parent=parent)
+        loading_gif = os.path.dirname(style.__file__) + "/svg/spinner-200.svg"
+        spinner = QtSvg.QSvgRenderer(loading_gif)
```
iLLiCiTiT, line 336: My experience with svg in Qt is that it is slow in comparison to png or gif. Recommendation is to implement rotating QLabel with png, on the other side I guess that one svg won't affect Qt much.

avalon/tools/loader/model.py
```diff
@@ -141,17 +150,72 @@ def set_version(self, index, version):
...
-    def refresh(self):
+    def fetch_subset_and_version(self):
+        """Query all subsets and latest versions from aggregation
+
+        (NOTE) The returned version documents are NOT the real version
+            document, it's generated from the MongoDB's aggregation so
+            some of the first level field may not be presented.
+
+        """
+        def _fetch():
```
iLLiCiTiT, line 161: `_fetch` should be class method, there is not much reason to have it inside.

iLLiCiTiT:
Logic seems to be ok for me. I'll have to modify it to be able to test it in our fork, but at this moment I don't see any possibly breaking part. Notes I have are nice to have changes, nothing will happen if they're not accepted.

avalon/tools/loader/model.py
```diff
@@ -141,17 +150,72 @@ def set_version(self, index, version):
...
+                    return
+                doc["parent"] = doc["_id"]
+                doc["_id"] = doc.pop("_version_id")
+                versions[doc["parent"]] = doc
+
+            self._doc_payload[:] = [(subset, versions.get(subset["_id"]))
+                                    for subset in _subsets]
+            self.doc_fetched.emit()
+
+        self._doc_payload[:] = []
```
iLLiCiTiT, line 196: I am curious why this is used `self._doc_payload[:] = []` and not just `self._doc_payload = []`?

iLLiCiTiT:
I can confirm it works with current multiselection implementation.

avalon/tools/loader/model.py
BigRoy: This is basically an in-place assignment. It would be the same as doing:
```python
my_list = ["a", "b", "c"]
my_list.clear()
```

I believe the reason to do explicit in-place assignments is due to the "mutable list" potentially being shared to another thread and otherwise that one working on an older copy of a list that isn't actually the same list as `self._doc_payload`. However, because it doesn't seem to be passed around as an object I don't think that's actually the case. @davidlatwe is it redundant for this case?

There should be some informational topics online on replacing lists in-place with regards to its mutability.

avalon/tools/loader/model.py
BigRoy: I guess it's there because it acts more like a local lambda written as a function. 

Also, if it were put on the class it shouldn't be a classmethod right? Because it uses variables/functions of the instance of the class inside of it (not how it refers to self). So if put on the class it would be a regular method.

iLLiCiTiT: > Also, if it were put on the class it shouldn't be a classmethod right?

I meant as object method `def _fetch(self):`, not as class method, my fault.

> it acts more like a local lambda written as a function

I see advantage of lamda but there's no advantage of that here, all `_fetch` need is `self`. Method `fetch_subset_and_version` has 4 lines in fact. All what these lines do, is preparation and launching another method in thread, but because the definition of launched method is inside it's harder to read (for me) what's really happening inside. 

This is "formatting" comment of my preferences. I have no ambition to change mindset because I know it's hard to change mine :)

mottosso: Looks right to me. Nesting a function tells me that it is only useful to the parent function, which seems to be the case here. If you put it in the class or instance, I would expect it to be useful standalone and by others, and require its own tests. The smaller the interface the better. Also, if Python did support multiline lambdas, this would be a good place for it.

avalon/tools/loader/model.py
davidlatwe: Maybe it could be in the `avalon.io` module. 🤔

avalon/tools/loader/model.py
davidlatwe: Sure :D

avalon/tools/loader/widgets.py
davidlatwe: Yep, there's only one, and it's animation playback will be stopped once the loading is completed. So yeah, I think we are safe :P

avalon/tools/loader/model.py
davidlatwe: > However, because it doesn't seem to be passed around as an object I don't think that's actually the case.

Yes, it's true that there's no need to use in-place assignment, just re-using the same object for no special reason. Will change it for good.

davidlatwe:
> I can confirm it works with current multiselection implementation.

Nice ! 🎉

BigRoy:
Should we get this PR ready for merging? I imagine @davidlatwe you've been using this in production since.
And likely @iLLiCiTiT adopted it in the meantime too?

iLLiCiTiT:
> And likely @iLLiCiTiT adopted it in the meantime too?

Yes, we use it.

mottosso:
Hope you're not waiting for me, please go ahead and merge if it's ready!

BigRoy:
Tested this and worked flawlessly here. @davidlatwe feel free to merge - or I'll merge in a few days.

davidlatwe:
Merging this !!

PR #556, opened 2020-07-24
Asset task options implementation

davidlatwe:
### Intooduce Asset Task Option

This feature aims to allow each asset to have different config in different task at different period of production time, which could be useful in publish process or other part of pipeline.

### Setup

Register task options by adding "options" into each "config.tasks" field in Avalon database's `project` document.
The peroject document would be like this:
```json
{
    "type": "project",
    "config": {
        "tasks": [
            {
                "name": "rigging",
                "options": {
                    "autoModelUpdate": {
                        "label": "Auto update model",
                        "default": false,
                        "help": "Auto update model to latest version after model publish."
                    }
                }
            },
            {
                "name": "otherTaskThatHasNoOption"
            }
        ]
    }
}
```
And in `.config.toml` file:
```toml
[[tasks]]
name = "rigging"

[tasks.options.autoModelUpdate]
label = "Auto update model"
default = false
help = "Auto update model to latest version after model publish."

[[tasks]]
name = "otherTaskThatHasNoOption"

```

And the asset document will be added one extra field `data.taskOptions` when it's been applied:
```json
{
    "type": "asset",
    "name": "Foo",
    "data": {
        "label": "Foo",
        "taskOptions": {
            "rigging": {
                "autoModelUpdate": {
                    "value": false
                }
            }
        },
        "tasks": [
            "rigging",
            "otherTaskThatHasNoOption"
        ]
    }
}
```

### Usage

To apply/change task options to each asset, we implement a new widget into Avalon's project manager tool.

image

With this, you can:
* View and update task option on each asset.
* Able to update multiple tasks' options on multiple assets in one action when enabled "Batch Edit".

### Notes

* This widget is hidden by default, you need to drag it out from the right side of the window.
* Option setup change will be written to database only when you pressed the "Save" button.
* Current edit will lost if you changed the selection of asset or task.
* Task widget's `view` has been enabled multi-selection for editing multiple task option on each asset. (This only applys to Project Manager, does not affect other tools)

### DIFF
--- a/avalon/inventory.py
+++ b/avalon/inventory.py
@@ -332,6 +332,7 @@ def _save_config_1_0(project_name, data):
     config["apps"] = data.get("apps", [])
     config["tasks"] = data.get("tasks", [])
     config["template"].update(data.get("template", {}))
+    config["assetOptions"] = data.get("assetOptions", [])
     config["families"] = data.get("families", [])
     config["groups"] = data.get("groups", [])
 
--- a/avalon/io.py
+++ b/avalon/io.py
@@ -353,22 +353,26 @@ def insert_many(items, ordered=True):
 
 
 @auto_reconnect
-def find(filter, projection=None, sort=None):
+def find(filter, projection=None, sort=None, *args, **kwargs):
     return self._database[Session["AVALON_PROJECT"]].find(
         filter=filter,
         projection=projection,
-        sort=sort
+        sort=sort,
+        *args,
+        **kwargs
     )
 
 
 @auto_reconnect
-def find_one(filter, projection=None, sort=None):
+def find_one(filter, projection=None, sort=None, *args, **kwargs):
     assert isinstance(filter, dict), "filter must be <dict>"
 
     return self._database[Session["AVALON_PROJECT"]].find_one(
         filter=filter,
         projection=projection,
-        sort=sort
+        sort=sort,
+        *args,
+        **kwargs
     )
 
 
--- a/avalon/pipeline.py
+++ b/avalon/pipeline.py
@@ -1153,19 +1153,12 @@ def load(Loader, representation, namespace=None, name=None, options=None,
 def _get_container_loader(container):
     """Return the Loader corresponding to the container"""
 
-    # loader = container["loader"]
-    # for Plugin in discover(Loader):
-    #
-    #     # TODO: Ensure the loader is valid
-    #     if Plugin.__name__ == loader:
-    #         return Plugin
+    loader = container["loader"]
+    for Plugin in discover(Loader):
 
-    from avalon import api
-    loaders = api.loaders_from_representation(api.discover(api.Loader),
-                                              container['representation'])
-    Loader = next((i for i in loaders if i.__name__ == container["loader"]), None)
-    print('Get loader from container: {}\n'.format(Loader))
-    return Loader
+        # TODO: Ensure the loader is valid
+        if Plugin.__name__ == loader:
+            return Plugin
 
 
 def remove(container):
--- a/avalon/pipeline.py
+++ b/avalon/pipeline.py
@@ -1153,12 +1153,19 @@ def load(Loader, representation, namespace=None, name=None, options=None,
 def _get_container_loader(container):
     """Return the Loader corresponding to the container"""
 
-    loader = container["loader"]
-    for Plugin in discover(Loader):
+    # loader = container["loader"]
+    # for Plugin in discover(Loader):
+    #
+    #     # TODO: Ensure the loader is valid
+    #     if Plugin.__name__ == loader:
+    #         return Plugin
 
-        # TODO: Ensure the loader is valid
-        if Plugin.__name__ == loader:
-            return Plugin
+    from avalon import api
+    loaders = api.loaders_from_representation(api.discover(api.Loader),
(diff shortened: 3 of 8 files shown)


tokejepsen:
Not sure I entirely get the end goal of this?

Is it like a per task metadata, so the config can disable/enable plugins?

davidlatwe:
> Is it like a per task metadata, so the config can disable/enable plugins?

Yes, per task metadata that can be applied to each asset, and for example, like in publish plugins, can change the behavior based on those metadata in each asset.

davidlatwe:
Our use case is, a rigger can set for one or more assets' rigs that is currently working on, which can be safely auto updated when model is being published.

tokejepsen:
Side question about "auto update". How are you tracking this and when does it get triggered?

BigRoy:
I like the fact of having specific rules/settings for an Asset but I do have some doubts about how this works. To get a clear picture I have a question first:

**Could this also just be an "Asset" specific setting? As in, does it really need to be on the Task?**

I imagine this could also just be on the asset's `data` 
```
rigFaceJointLimit: 10
rigBodyJointLimit: 20
rigAutoUpdate: True
```

Seems like that was how it was originally intended (to be specific data to that asset).
In that sense it's more of User Interface front-end for asset data, which I think is a great step forward. E.g. also allowing to set an FPS override or resolution setting for a Shot in the Project Manager.

I guess the specification could allow for assetOptions that apply only to specific tasks (`allowedTasks: ["rigging"]`)?

davidlatwe:
> Side question about "auto update". How are you tracking this and when does it get triggered?

We have a simple dependency tracking mechanism, but only in Maya. When publishing, the instance that is being published will lookup the history and see if any node that is part of loaded subset. If so, the version document's `_id` of that loaded subset will written into the instance's version document for future lookup.

> Could this also just be an "Asset" specific setting? As in, does it really need to be on the Task?

*(Hmm, maybe I shouldn't use the phrase "to each asset", since it does sounds like applying same settings to every asset ?)*

Yes, it's asset specific settings when you apply options from those has been registered in project's tasks.

Before those option being applied to any assets, we need a way to define and present those options, and binding them with tasks seems a good fit.

> E.g. also allowing to set an FPS override or resolution setting for a Shot in the Project Manager.
> I guess the specification could allow for assetOptions that apply only to specific tasks (allowedTasks: ["rigging"])?

Ah, need to think about this, organizing those option into project document's "config.assetOptions" with `allowedTasks` entry seems to be more flexible !

davidlatwe:
If we move the option's defenition it into a higher scope like "assetOptions", then the project document would be like:
```json
{
    "type": "project",
    "config": {
        "assetOptions": [
            {
                "name": "rigAutoModelUpdate",
                "label": "Auto update model",
                "default": false,
                "help": "Auto update model to latest version after model publish.",
                "onTasks": [
                    "rigging"
                ]
            },
            {
                "name": "shotFps",
                "label": "Frame rate",
                "default": 24,
                "help": "Override frame rate for shot.",
                "onTasks": [
                    "layout",
                    "animating",
                    "lighting"
                ]
            }
        ]
    }
}
```

`.config.toml`:
```toml
[[assetOptions]]
name = "rigAutoModelUpdate"
label = "Auto update model"
default = false
help = "Auto update model to latest version after model publish."
onTasks = [
    "rigging"
]
```
*Which is a bit better to write* 👍 

And the asset document that has been applied would be like:

```json
{
    "type": "asset",
    "name": "Foo",
    "data": {
        "label": "Foo",
        "assetOptions": {
            "rigAutoModelUpdate": {
                "value": false
            }
        },
        "tasks": [
            "rigging"
        ]
    }
}
```

davidlatwe:
A note to self:
Need to pay attention on existing entries in assets like "startFrame" or "endFrame" (or "frameStart", "frameEnd"), since those may have been written into the asset document in existing production, so the landing opint of those *assetOptions* may not be able to like the example asset document above. 🤔

BigRoy:
> new assetOptions may have to live directly under "data" field in asset document

Exactly what I was expecting yes. I imagined these new "rig specific settings" similar to an Asset/Shot specific setting like Resolution/FPS (if overridden from a project's default) or things like that.

Would love to know whether that would work for your use cases?

davidlatwe:
Yeah, that would work in our case even better 🚀

davidlatwe:
I have refactored to adopt `"assetOptions"`, but noted that the `config-1.0` schema also need to be updated as well since "assetOptions" is a new property in project config.

Or should we make a `config-1.1` schema ? Or  `config-2.0` ?

mkolar:
Coming late to the party, but good to see what you guys agreed on at the end. It is a lot more flexible and easily expandable to many other types of parameters than the original concept.

davidlatwe:
> Or should we make a config-1.1 schema ? Or config-2.0 ?

I will add `config-1.1`, since it's a minor change.

BigRoy:
What's the status on this? Anyone using this in production?

davidlatwe:
> > Or should we make a config-1.1 schema ? Or config-2.0 ?
> 
> I will add `config-1.1`, since it's a minor change.

I think the last comment I wrote was the last piece but some how I never finish it. 😮 
But I did put this into our production branch.

PR #494, opened 2019-12-16
Fix #492: Resolve CI coverage bug

davidlatwe:
### What's changed ?

This PR resolve #492 and unpinned `sphinx==1.7.9` (Since #338)

PR #559, opened 2020-07-27
Fix ci coverage & azure

### LINKED ISSUE #558: CI test failing on sending coverage report
**What happened?**
Avalon's CI test keeps failing when it need to send the coverage report via `mayapy` which is now an outdated Python..

See https://travis-ci.org/github/getavalon/core/builds/712074966#L548-L572
```log
Ran 32 tests in 11.158s
OK
/usr/autodesk/maya/lib/python2.7/site-packages/urllib3/util/ssl_.py:394: SNIMissingWarning: An HTTPS request has been made, but the SNI (Server Name Indication) extension to TLS is not available on this platform. This may cause the server to present an incorrect TLS certificate, which can cause validation failures. You can upgrade to a newer version of Python to solve this. For more information, see https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
  SNIMissingWarning,
/usr/autodesk/maya/lib/python2.7/site-packages/urllib3/util/ssl_.py:143: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. You can upgrade to a newer version of Python to solve this. For more information, see https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
  InsecurePlatformWarning,
  File "<string>", line 1, in <module>
  File "/usr/autodesk/maya/lib/python2.7/site-packages/coveralls/__init__.py", line 95, in wear
    skip_ssl_verify=args.skip_ssl_verify,
  File "/usr/autodesk/maya/lib/python2.7/site-packages/coveralls/api.py", line 8, in pos
    return requests.post(url, files={'json_file': json_file}, verify=(not skip_ssl_verify))
  File "/usr/autodesk/maya/lib/python2.7/site-packages/requests/api.py", line 119, in post
    return request('post', url, data=data, json=json, **kwargs)
  File "/usr/autodesk/maya/lib/python2.7/site-packages/requests/api.py", line 61, in request
    return session.request(method=method, url=url, **kwargs)
  File "/usr/autodesk/maya/lib/python2.7/site-packages/requests/sessions.py", line 530, in request
    resp = self.send(prep, **send_kwargs)
  File "/usr/autodesk/maya/lib/python2.7/site-packages/requests/sessions.py", line 643, in send
    r = adapter.send(request, **kwargs)
  File "/usr/autodesk/maya/lib/python2.7/site-packages/requests/adapters.py", line 514, in send
    raise SSLError(e, request=request)
requests.exceptions.SSLError: HTTPSConnectionPool(host='coveralls.io', port=443): Max retries exceeded with url: /api/v1/jobs (Caused by SSLError(SSLError(1, '_ssl.c:507: error:14077410:SSL routines:SSL23_GET_SERVER_HELLO:sslv3 alert handshake failure'),))
The command ". test_docker.sh" exited with 1.
```

### Proposal

Currently the test is invoked in Maya docker container, also generates coverage report and sending it there.
The test should be ran by `mayapy`, but we should define a path for it to save coverage file in somewhere else and pick up by another python process to send the report.

davidlatwe:
### What's changed

This PR fix #558.

The code coverage submission was decoupled from maya test script, and provide a new scheme for future testing on multiple DCC tools or versions.

### New code coverage scheme

1. The test script generates a `.coverage` data file after complete.
2. Rename the `.coverage` into `.coverage.{subject}`, where the *subject* could be DCC tool name and version.
3. After all test scripts completed, run `coverage combine` to combine all `.coverage` prefixed data file into one.
4. Submit to coveralls.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,10 @@ language: python
 python:
   - 3.7
 
+env:
+  global:
+    - COVERALLS_SERVICE_NAME=travis-ci
+
 services:
   - docker
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ script:
 
 after_success:
   - coverage combine
-  - coveralls
+  - coveralls debug
 
 before_deploy:
   - sudo chmod -R 777 docs/build/
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,15 +3,11 @@ language: python
 python:
   - 3.7
 
-env:
-  global:
-    - COVERALLS_SERVICE_NAME=travis-ci
-
 services:
   - docker
 
 install:
-  - pip install coverage coveralls
+  - pip install coverage python-coveralls
   - docker run --name avalon-mongo -d mongo
   - . build_docker.sh
 
@@ -21,7 +17,7 @@ script:
 
 after_success:
   - coverage combine
-  - coveralls debug
+  - coveralls
 
 before_deploy:
   - sudo chmod -R 777 docs/build/
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,18 +1,23 @@
 language: python
 
 python:
-  - 2.7
+  - 3.7
 
 services:
   - docker
 
 install:
+  - pip install coverage coveralls
   - docker run --name avalon-mongo -d mongo
-  - docker build -t avalon/core -f Dockerfile-maya2016 .
+  - docker build -t avalon/core:maya -f Dockerfile-maya2016 .
 
 script:
   - . test_docker.sh
   - . build_docs.sh
+
+after_success:
+  - coverage combine
+  - coveralls
 
 before_deploy:
   - sudo chmod -R 777 docs/build/
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,8 @@ services:
   - docker
 
 install:
-  - pip install coverage python-coveralls
+  - pip install coverage==4.5.4
+  - pip install python-coveralls
   - docker run --name avalon-mongo -d mongo
   - . build_docker.sh
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ services:
 install:
   - pip install coverage coveralls
   - docker run --name avalon-mongo -d mongo
-  - docker build -t avalon/core:maya -f Dockerfile-maya2016 .
+  - . build_docker.sh
 
 script:
   - . test_docker.sh
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -54,7 +54,7 @@ jobs:
 
   - job: MacOS
     pool:
-      vmImage: "macOS-10.13"  # High-Sierra
+      vmImage: "macOS-10.15"  # Catalina
     strategy:
       matrix:
         Python27:
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,6 +1,12 @@
 [run]
 source = avalon
 
+[paths]
+# Remap source root path in docker
+source =
+    avalon
+    /workspace/avalon
+
 [report]
 include = *avalon*
 omit = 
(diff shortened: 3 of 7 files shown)


.travis.yml
```diff
@@ -1,19 +1,24 @@
...
-  - docker build -t avalon/core -f Dockerfile-maya2016 .
+  - . build_docker.sh
 
 script:
   - . test_docker.sh
   - . build_docs.sh
 
+after_success:
+  - coverage combine
+  - coveralls
```
tokejepsen, line 20: Seems to be issues with this command; https://travis-ci.org/github/getavalon/core/builds/712117798#L608

azure-pipelines.yml
```diff
@@ -54,7 +54,7 @@ jobs:
 
   - job: MacOS
     pool:
-      vmImage: "macOS-10.13"  # High-Sierra
+      vmImage: "macOS-10.15"  # Catalina
```
tokejepsen, line 57: Awesome stuff!

davidlatwe:
About the error message which @tokejepsen mentioned in [review](https://github.com/getavalon/core/pull/559#discussion_r460746217):
```sh
$ coveralls
Submitting coverage to coveralls.io...
Could not submit coverage: 422 Client Error: Unprocessable Entity for url: https://coveralls.io/api/v1/jobs
```
This reminds me what happened in getavalon/core#494, looks like it was the same cause.

So the reason to that error was because the `COVERALLS_REPO_TOKEN` did not get passed, which I am sure it does exists in travis-ci setup.

Turns out the problem was in the Coveralls's python integration module.

I have changed to use [coveralls-clients/coveralls-python](https://github.com/coveralls-clients/coveralls-python) (`pip install coveralls`) in commit b17b129, since the original integration module [z4r/python-coveralls](https://github.com/z4r/python-coveralls) (`pip install python-coveralls`) **has stopped updating for almost a year..**

And with [coveralls-clients/coveralls-python](https://github.com/coveralls-clients/coveralls-python) integration, it just won't get the coveralls repo token. Because I tried `coveralls debug` command in 1c83ce1 and the token (should be something like `"repo_token": "[secret]"`) did not exists in the JOSN dump string in submission debug message.

Anyway, I changed back to the [z4r/python-coveralls](https://github.com/z4r/python-coveralls) integration, which requires to pin `coverage==4.5.4`, but here's a new error message:
```
INFO:coveralls:422
INFO:coveralls:{"message":"Couldn't find a repository matching this job.","error":true}
```
Maybe it will be gone in the merge commit ? Not sure.

Or, maybe it's time to change to use [codecov](https://codecov.io/) ? @mottosso

.travis.yml
davidlatwe: Please see comment https://github.com/getavalon/core/pull/559#issuecomment-664370166

mottosso:
> Maybe it will be gone in the merge commit ? Not sure.

I saw the error, and had a look at coveralls.io, and indeed it does not exist. :O I think GitHub updated the permission features a while back, and it looks like Coveralls got left out. I've added it now and updated the repo token, and expect it will work. For both PRs and merges. Let's give it a try, and if not switch to something else.

mottosso:
Well, it does work, but not really.

https://coveralls.io/builds/32424978

davidlatwe:
Looks like there was a problem when combining coverage data
```
$ coverage combine
Coverage.py warning: Couldn't read data from '/home/travis/build/getavalon/core/.coverage.maya': UnicodeDecodeError: 'utf-8' codec can't decode byte 0xf6 in position 106: invalid start byte
No usable data files
```
https://travis-ci.org/github/getavalon/core/jobs/713255428#L683

I *guess*.., it was because we are foced to use outdated `coverage==4.5.4`  because of we can only use `python-coveralls` to send coverage data to `coveralls.io`.

So many pitfall after Python 2.7 being rejected :(

Maybe `codecov` is the only way out for code coverage. 😭

davidlatwe:
> I guess.., it was because we are foced to use outdated `coverage==4.5.4`

Ah, sorry, maybe it's because I forgot to pin `coverage==4.5.4` back in `Dockerfile-maya2016`. :S

Issue #409, opened 2019-08-05
Allzparkalon


mottosso:
### Goal

Implement final touches for compatibility with Allzpark.

- https://allzpark.com/avalon/

### Tasks

Overall it's quite compatible, with a few minor adjustments to the overall workflow.

- [ ] Make environment variables `AVALON_TASK`, `AVALON_ASSET` and `AVALON_WORKDIR` **optional** for launch of an app, like Maya
- [ ] Make the "Set Context" tool capable of setting those, and creating a working directory
- [ ] Make a package for the Project Manager, such that you can actually manage assets completely via the GUIs; I suspect that should be a plain port of the Project Manager "action" that Avalon Launcher is currently using

See aforementioned website for details and background.

mottosso:
Because Allzpark changes the fundamental nature of how a host is launched and assets specified with Avalon, we'll need to adjust how users interact with it.

Here's some ideas for how to accomplish that.

![allzparkalon_mockup](https://user-images.githubusercontent.com/2152766/63464689-af137e80-c457-11e9-86f7-3d935f43a86b.gif)

### Landing Screen

Upon entering a host, the Avalon Landing Screen appears.

<img width=400 src=https://user-images.githubusercontent.com/2152766/63464334-f9e0c680-c456-11e9-8095-c5b959a912b2.png>

It'll display three things.

1. Your project, along with related graphics like icon, label and custom backdrop
2. Your recent assets/tasks, as thumbnails
3. Button for showing Context Manager, post-implementation of #412 

Clicking on a recent item takes you immediately into the latest file for that task. The goal of this dialog is to get out of the way; it isn't the first time an artist launches this app, the only thing on their mind is to carry on with what they were doing the day before.

### Context Manager+

Behind the Landing Screen is the new Context Manager.

<img width=400 src=https://user-images.githubusercontent.com/2152766/63464507-517f3200-c457-11e9-9b9b-3d3483541597.png>

This'll account for more complex choice, or first-choice. When a new task or project is started. It'll resemble Allzpark, but rather than changing project, it'll change asset, task and scene file.

It'll provide:

1. Thumbnail per Project, Asset and Task, the latter two being editable by clicking
1. Thumbnail from saved file
2. Comment for saved file
3. Metadata in sidepanel, such as:
    - Last modified
    - Who modified
    - Loaded representations, along with their state (imported/referenced)
    - Whether representations are of the latest versions (e.g. old rig)

[...]

PR #560, opened 2020-09-03
Supporting Rez/Allzpark

davidlatwe:
Based on #409, this PR adds the support of Rez/Allzpark styled environment setup and application launching workflow.

## Preface

Let's compare the current and up-coming launcher first.

Current Launcher:
1. User selects which project/asset/task to work on, click application
2. Launcher creates an instance of `avalon.pipeline.Application` and loads `{app}.toml`
3. And compose environment
4. Initialize working directory, auxiliary file copying may occur
5. Finally, launch application

In Allzpark:
1. User selects which project to work on, click application
2. Rez resolves environment
3. Launch application

The difference between these two workflows are, other than Asset/Task selecting is not in Allzpark, working directory's initialization is not there as well.

Although selecting Asset/Task after application startup can be easily accepted with the help of Avalon's *Workfile app*, but the bottom line of doing any task is, there must be a working directory.

**So with or without asset/task, we must *define* a working directory path in Allzpark.**
**But do we need to *initialize* it in Allzpark as well ?**

By saying "in Allzpark", it means "in Rez package", which mainly used to build up environment so it's not a good fit for file system I/O related operations (it could, but not trivial).

Hence, **the application that is being launched should be able to handle working directory's initialization itself.**

## TL;DR

So here's the goals for supporting new application launching process :

* `AVALON_ASSET`, `AVALON_TASK` can be undefined
* Avalon should respect and use predefined `AVALON_WORKDIR`
* Working directory can be initialized by the application (e.g. Maya) itself

## What's changed

* `pipeline.install` will initialize working directory if flag `_AVALON_APP_INITIALIZED` is not set.
* `AVALON_WORKDIR` will be respected if it's presented in `Session` during install.
* Avalon menu's context label will become `--, --` if asset/task is undefined.
* New session schema 3.0 is added, for opt-out `AVALON_ASSET`.
* ~New config schema 2.0 is added, for opt-out `apps`.~ (reverted in b6429f5)
* New environment var `AVALON_SESSION_SCHEMA` is added, for adopting new workflow.
* Explicitly defined environment var `AVALON_APP_NAME` in session. This isn't new, just making it clearer.

**This PR does/should NOT change current workflow, just supporting new one.**

## Reference

Here's my current Avalon Rez package setup on adopting Allzpark :

https://github.com/davidlatwe/rez-park/blob/master/_pln/avalon/package.py

Noted that I am still using `{app}.toml` files for working directory initialization, but just the part of file copying and creating default dirs.

https://github.com/davidlatwe/rez-park/tree/master/_pln/merlin/apps

Here's what it *looks* like :

image


mottosso:
Looks great! What's missing is a getting-started step-by-step.

E.g.

```py
$ pip install allzpark
$ allzpark
...
```

Along with how to..

1. Create your project/package, maybe reference the https://allzpark.com docs for in-depth details
2. Link a few dependencies to your project
3. How to use that project with Allzpark, to launch e.g. Maya
4. Once in Maya, how to set your shot and task, along with creating the workdirs
5. Profit

PR #561, opened 2020-09-18
Speed up getting assets for asset switcher

jasperges:
**What's changed?**

When querying the assets to populate the asset switcher a database call was made for every asset. If you have lots of assets this can really add up and take quite some time (for me roughly 30 seconds for 2000 assets).

This change does only 2 database calls and then does the filtering afterwards. In my case this reduces the time taken to around 3 seconds for 2000 assets.

### DIFF
--- a/avalon/tools/sceneinventory/app.py
+++ b/avalon/tools/sceneinventory/app.py
@@ -677,14 +677,12 @@ class SwitchAssetDialog(QtWidgets.QDialog):
 
     def _get_assets(self):
         filtered_assets = []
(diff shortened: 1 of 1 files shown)


tokejepsen:
@iLLiCiTiT has worked on the same thing.

jasperges:
Whoops, forgot to check other PR's first... will have a look at https://github.com/getavalon/core/pull/551.

jasperges:
Okay, sorry for the noise and nevermind. https://github.com/getavalon/core/pull/551 works fine for me. Closing this one.

Issue #565, opened 2020-10-27
Pype & Avalon - let's sync the cores


mkolar:
First of all, we would like to say that we truly appreciate the community efforts and a huge part of our current state is owed to the first years of hard work by Marcus and the community he had built around the Avalon and Pyblish projects. 

Based on multiple conversations on gitter and a few other places over the past months, we want to propose a possible solution to getting Pype fork of avalon-core and upstream repository in sync once again. I’ll try to explain the current situation, the functional differences and our proposal.

## Where we are

Pype’s fork of avalon-core is currently just over 1000 commits ahead of the upstream and a few commits behind (33 at the time of typing this) and we’ve gone through over 200 internal PRs in the past 2 years.

Pype fork of avalon-core is deployed in production at 10 studios with more waiting for availability, but our team is spread quite thin as, apart from supporting our clients, we are also busy working on pype 3.0, which is a major refactor and architectural change to our current code base.

There are plenty of reasons why not that many of our changes made it to the core along the way, but we feel that these are the main ones
Our development is a lot more dynamic than that of the upstream. We are directly funded by our users and very often have to react extremely quickly to production requirements.
On the other hand, thanks to the “fail often, fail early” approach, we believe that we arrive at desired production solutions through iteration and improvements quicker, rather than trying to come up with the “perfect” solution from the start. 
Resulting from that, we saw a fairly slow review process of our PRs, which naturally resulted in less push from our end. 
Complaints about lower code standard on pype side compared to avalon-core.

There is, of course, a variety of changes in all of that but, really it boils down to this list. 

## Feature Differences

**Master Version**s - Option to publish versionless on top of standard incremental versions. We make a “master” version that essentially just points to the latest published explicit version. Scene inventory then allows the artist to switch between explicit and master versions of the subset.

**Thumbnails in loader** - Mechanism that allows the studio to publish thumbnails and have them loaded back into the loader. Basic disk resolver for the thumbnails is provided, but configs can extend this with more resolvers if there is a preference to store thumbs differently. For instance, serving them from url or database.

**Library Loader** - essentially the same as a standard loader, but allow an artist to import content from a different project that has been tagged as `library`

**More robust tasks** - we tweaked the schema for tasks on the project and assets to be able to store `task type` information and potentially any other metadata, like task short codes. The key reason for this addition was the ability to filter on various places based on the task type, which is more consistent than just the task name. Type and short codes could of course also be used in path templates.

[...]

PR #567, opened 2020-12-10
Avalon 6.0 proposal

mkolar:
This PR relates to https://github.com/getavalon/core/issues/565 and should be considered as a starting point of Avalon 6.0 conversation. It is not ready to be merged at the time of creation.

We have removed as much pype dependencies as possible at this point, with only two remaining:
Pype Anatomy - our own path templating system that we use as replacement for the avalon standard templates, This should of course be removed completely before considering merge, but right now we at least try to bypass it and used avalon default if PYPE is not present
Creator tries to pre-select correct family for a corresponding task based on our settings system. This is also bypassed if pype is not use, but will be removed from core ultimately as it depends on our system

### **Enhancements**

**Master Versions** - Option to publish versionless on top of standard incremental versions. We make a “master” version that essentially just points to the latest published explicit version. Scene inventory then allows the artist to switch between explicit and master versions of the subset.
**Thumbnails in loader** - Mechanism that allows the studio to publish thumbnails and have them loaded back into the loader. Basic disk resolver for the thumbnails is provided, but configs can extend this with more resolvers if there is a preference to store thumbs differently. For instance, serving them from url or database.
**Library Loader** - essentially the same as a standard loader, but allow an artist to import content from a different project that has been tagged as library
**More robust tasks** - we tweaked the schema for tasks on the project and assets to be able to store task type information and potentially any other metadata, like task short codes. The key reason for this addition was the ability to filter on various places based on the task type, which is more consistent than just the task name. Type and short codes could of course also be used in path templates.
**Faster asset switcher** - running Avalon on large CG scenes it was becoming impossible to switch any asset representations due to the way they were queried from DB, we had to change this to be able to run some live productions. That, however, introduces behaviour that doesn’t match 100% the original, so this piece needs to be solved still.
**Multiselection loader** - see #550
**IO tweaks** - we can use functions from IO module in different contexts when needed. Very important for example for the library loader.

### **New Hosts**

**Photoshop** - Production tested, uses Websocket server that would have to be ported to avalon from Pype.
**Harmony** - Production tested, under continuous develoment
**Blender** - Limited production testing, Used on pre-production of a large project, abandoned for reasons outside of pipeline scope. The base implementation was stabilised and is ready for configs to extend it with plugins, loaders and such.
**Unreal Engine** - Beta stage. Limited testing was done, and more development has been done by Clothcat Animation on the config side (pending approval and merge to pype)
**TV Paint** - Production Beta - under testing
**After Effects** - Production Beta - under testing, uses same server as photoshop

I will update the description and add requirements to run this soon, just wanted to get it here, so people can maybe start looking at it in the meantime.


mottosso:
This PR could use some love, this has flown completely under my radar. @BigRoy and @davidlatwe have you seen this? Thanks @mkolar for making this, has there been any changes on your side since last year that could be incorporated into here?

What I'd love to see happen is:

1. Maintain backwards compatibility
2. Add all the new things
3. Add log message/warning to replaced/deprecated functions

From which point we can start removing old things piece-by-pieace in their own PRs with their own discussions.

PR #568, opened 2021-06-16
Better error reporting for filepath resolution

icyvapor:
**What's changed?**
- Add assertion to ensure we resolve to non-empty path result, instead of silently returning None. 
- Add more descriptive error messaging to the `get_representation_path` function
- Misc. code fixes.

### DIFF
--- a/avalon/pipeline.py
+++ b/avalon/pipeline.py
@@ -1296,7 +1296,9 @@ def get_representation_path(representation):
         if os.path.exists(path):
             return os.path.normpath(path)
         else:
-            log.warning("path from representation doesn't exist: {}".format(path))
+            log.warning(
+                "path from representation doesn't exist: {}".format(path)
+            )
             return None
 
     def path_from_config():
@@ -1304,7 +1306,9 @@ def get_representation_path(representation):
             version_, subset, asset, project = io.parenthood(representation)
         except ValueError:
             log.warning(
-                "Representation {} missing in database!".format(representation["name"])
+                "Representation {} not found in database!".format(
+                    representation["name"]
+                )
             )
             return None
 
@@ -1312,7 +1316,9 @@ def get_representation_path(representation):
             template = project["config"]["template"]["publish"]
         except KeyError:
             log.warning(
-                "No publish path template in config of {}!".format(project["name"])
+                "No publish path template in config of {}!".format(
+                    project["name"]
+                )
             )
             return None
 
--- a/avalon/pipeline.py
+++ b/avalon/pipeline.py
@@ -1276,41 +1276,43 @@ def get_representation_path(representation):
 
     """
 
-    def path_from_represenation():
+    def path_from_representation():
         try:
             template = representation["data"]["template"]
-
         except KeyError:
+            log.warning("Cannot find template in representation document")
             return None
 
         try:
             context = representation["context"]
             context["root"] = registered_root()
             path = template.format(**context)
-
-        except KeyError:
-            # Template references unavailable data
+        except KeyError as e:
+            log.warning(
+                "Missing value for filepath template:{}".format(e)
+            )
             return None
 
         if os.path.exists(path):
             return os.path.normpath(path)
+        else:
+            log.warning("path from representation doesn't exist: {}".format(path))
+            return None
 
     def path_from_config():
         try:
             version_, subset, asset, project = io.parenthood(representation)
         except ValueError:
-            log.debug(
-                "Representation %s wasn't found in database, "
-                "like a bug" % representation["name"]
+            log.warning(
+                "Representation {} missing in database!".format(representation["name"])
             )
             return None
 
         try:
             template = project["config"]["template"]["publish"]
         except KeyError:
-            log.debug(
-                "No template in project %s, "
-                "likely a bug" % project["name"]
+            log.warning(
+                "No publish path template in config of {}!".format(project["name"])
             )
             return None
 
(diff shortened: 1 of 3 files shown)


avalon/pipeline.py
```diff
@@ -1276,41 +1276,49 @@ def get_representation_path(representation):
 
     """
 
-    def path_from_represenation():
+    def path_from_representation():
```
mottosso, line 1279: This is great, but let's also add backwards compatibility by just putting a `path_from_represenation = path_from_representation` below this function to create an alias. If you really want to go all-in, the alias could be an actual function that calls the real function, but raises and catches a deprecation warning so it can print which line it was called on in the user's code. :)

avalon/pipeline.py
```diff
@@ -1325,31 +1333,36 @@ def path_from_config():
...
-            "task": Session.get("AVALON_TASK", "")
+            "task": Session.get("AVALON_TASK", ""),
+            "hierarchy": asset["data"].get("hierarchy", "")
         }
 
         try:
             path = template.format(**data)
         except KeyError as e:
-            log.debug("Template references unavailable data: %s" % e)
+            log.warning("Missing value for filepath template: {}".format(e))
```
mottosso, line 1343: Could be made *even* more clear I think. How about this?

```
"One of the variables used in this filepath is unavailable: {}, these are the variables available: {}".format(
  path, ", ".join(data.keys()))
```

avalon/pipeline.py
BigRoy: This would be redundant right since the function declaration is inside another function? Out of that scope nobody would be calling it. Right?

mottosso: Ah yes that's right, good catch. It's safe.

avalon/pipeline.py
BigRoy: Changing all these logs to warnings I think will get confusing quick for most artists. Say your pipeline relies not on the first behavior but the second or third. In that case every time you are trying to get a representation the artist will end up seeing the first two warnings. E.g. when loading assets.

Having better messaging I think is good. I am just wondering whether making these warnings is the most sensible direction.

davidlatwe: Yeah, better to keep the log level at `debug` I think.

avalon/pipeline.py
```diff
@@ -1325,31 +1333,36 @@ def path_from_config():
...
-    return (
-        path_from_represenation() or
-        path_from_config() or
-        path_from_data()
-    )
+    resolved_path = (path_from_representation()
+                     or path_from_config()
+                     or path_from_data())
+
+    assert resolved_path, "Cannot resolve path!"
```
davidlatwe, line 1364: Not sure if adding this assertion now is a good idea. One may already checking path resolved by return value is `None` or not. Adding assertion would mean forcing everyone to handle `AssertionError` on top of current usage after this. No ?

One alternative would be:

```python
def get_representation_path(representation, error_out=False):
    ...
    if error_out:
        assert resolved_path, "Cannot resolve path!"
    return resolved_path

```

mottosso: Yes, definitely don't introduce new exceptions. Debug log it instead, that's for developers whereas new exceptions are raised all the way to the artist.

avalon/pipeline.py
mottosso: Agreed.

icyvapor: Lowered.

avalon/pipeline.py
icyvapor: I can agree with this being a core API function thus not throwing additional error/exceptions.

I've moved this over to the loader initialization, where a valid path is required for Loader to behave properly. If that happens, it means something went horribly wrong with the asset being loaded. So I think it's reasonable to error out with assert failure.

mottosso: An assert isn't the right approach here. This will still propagate an error to the user, and the user won't be the one to deal with it. At this point, the onus is on Avalon to handle and format any error that occurs in a way that makes sense to the user. If the Loader is to have an assert, than whomever instantiates it needs to handle that assert.

In this case, if a filename really is invalid, the problem is further up the pipeline. Either an asset wasn't published correctly, or a physical path that used to exist exists no more. That's what we can tell the end user.

In practice, it looks like this fname isn't actually used?

https://github.com/getavalon/core/blob/91f5ae87120f61d8beb57af099d321e28178399f/avalon/maya/pipeline.py#L554-L561

So we'd be throwing an assert error for no reason. Let me know what you think.

icyvapor: Right, an assert failure is probably too blunt for the end user, though I don't quite see how the current instance of Loader can  gracefully function when there's no valid filename. ( Hmm I guess it's very unlikely to happen, but hopefully this discussion is interesting in theory..) 

A bit more details about what led me here: 
I was hacking and experimenting with the AVA config, where I included {hierarchy} in the filepath template. The very first error I saw was on Line 559 in the code snippet above, saying something like `'None' type has no .replace()`. So I put the line into a ` if self.fname: ` block, to move on when there's no `fname`. That's when I got the Maya error I initially posted in the Gitter chat. This `self.fname` will be passed to cmds as an argument, having no value would give that Maya error popup.

So, consider we create an Loader instance, and the `context`  provided is broken somehow, and causing `self.fname` to be None. Then this loader instance can't fulfill its mission, since it can't figure out which file to load. 

I was also sorting out thoughts while I'm typing this, I now think maybe the initialization code of Loader class should handle possible exception, since the 'context' it provides may result in a broken Loader. 

Sorry for the long-winded reply. If the changes required are too complicated, I don't mind going with whatever you suggest at all.  =)

davidlatwe: Agree that `'None' type has no .replace()` is confusing for everyone. I think just add an if-statement to check `self.fname is None` is enough.

The `fname` attribute in `Loader.__init__()` is just a start point for developer to implement the rest of the function of loader plugin, e.g.
https://github.com/getavalon/core/blob/91f5ae87120f61d8beb57af099d321e28178399f/avalon/pipeline.py#L178-L190

So, yeah, if you prefer raising AssertionError or any other action, that should be in your own implemented loader plugin method e.g. `Loader.load()`

mottosso: > though I don't quite see how the current instance of Loader can gracefully function when there's no valid filename

I think the proper approach would be to catch this *before* it gets to the loader. Who's passing an invalid filename to it? And why? That's what we've got to catch and handle, at the source. So that's something I think would be welcomed and and encouraged!

The debug messages are still great though, because it allows you the developer to enable debug messages and gain insight into what's going on. It's just important to keep these two audiences in mind, and provide relevant information to the relevant party.

BigRoy: > Right, an assert failure is probably too blunt for the end user, though I don't quite see how the current instance of Loader can gracefully function when there's no valid filename. ( Hmm I guess it's very unlikely to happen, but hopefully this discussion is interesting in theory..)

It could very well be that the Loader itself does some custom logic based on the context data that it in the end does get an existing path and that it could disregard `self.fname` completely. So technically, it could already be used.

I agree however that the case you hit should come with more graceful handling. I'd say in your case it sounds like the particular loader could handle it more gracefully with a message to the user maybe?

icyvapor: Thank you guys for the input. Now the Maya Loader simply log an warning when the fname is None.

I was thinking of adding an assertion in the `avalon.maya.compat.load()`, which was on my original stacktrace. But just realized this function is deprecated. So I'll let it be.

mottosso:
Brilliant, excellent work everyone!

